PR #755, opened 2026-02-11
🛡️ Sentinel: Fix Safe Evaluation Bypass via Package Separator

EffortlessSteven:
🛡️ Sentinel: [CRITICAL/HIGH] Fix Safe Evaluation Bypass

🚨 Severity: HIGH
💡 Vulnerability: Safe Evaluation Bypass via Package Separator
The "Safe Evaluation" mode in the debug adapter could be bypassed using Perl's archaic package separator syntax (e.g., `CORE'system`). The parser incorrectly identified the package separator `'` as the start of a string literal, causing the security filter to ignore the subsequent dangerous function call.

🎯 Impact: Arbitrary Code Execution
An attacker (or a user unknowingly evaluating a malicious expression) could execute arbitrary commands (e.g., `system`, `exec`) even when "Safe Evaluation" is strictly enforced, defeating the purpose of the security feature.

🔧 Fix: Improved String Detection
The `is_in_single_quotes` helper function was updated to recognize that a single quote preceded by a word character is likely a package separator, not a string delimiter. This ensures that expressions like `CORE'system` are correctly parsed as code and blocked.

✅ Verification:
- Added a regression test `crates/perl-dap/tests/security_repro_test.rs` that attempts to evaluate `CORE'system("ls")`.
- Verified that the test fails (bypasses security) without the fix and passes (blocks execution) with the fix.
- Verified that valid single-quoted strings (e.g., `print 'hello'`) are still handled correctly (though `print` itself is blocked, the string parsing logic is sound).

---
*PR created automatically by Jules for task [7605457544879714927](https://jules.google.com/task/7605457544879714927) started by @EffortlessSteven*

### DIFF
--- a/crates/perl-dap/src/debug_adapter.rs
+++ b/crates/perl-dap/src/debug_adapter.rs
@@ -1823,6 +1823,7 @@ impl DebugAdapter {
 fn is_in_single_quotes(s: &str, idx: usize) -> bool {
     let mut in_sq = false;
     let mut escaped = false;
+    let mut prev_char: Option<char> = None;
 
     for (i, ch) in s.char_indices() {
         if i >= idx {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as stale/superseded during repository cleanup after merged security/perf/UX batches; reopen if this specific variant should continue.

PR #1305, opened 2026-03-12
Fix `my` declarations in function call parentheses

EffortlessSteven:
### Motivation
- The parser treated `my`/`our`/`local`/`state` as bare identifiers in expression contexts, which broke cases like `foo(my $x)` where the inner token sequence is a variable declaration.
- The intent is to correctly recognize declaration-shaped tokens when these keywords appear inside function-call parentheses and produce a proper variable declaration node in the AST.

### Description
- Update `crates/perl-parser-core/src/engine/parser/expressions/primary.rs` to detect when `my`/`our`/`local`/`state` is followed by a declaration-shaped token (via `peek_second`) and call `parse_variable_declaration()` in that case. 
- Preserve the previous behavior of treating those keywords as bare identifiers when the following tokens do not match a declaration shape by falling back to the original identifier path.
- Add a regression test `test_my_declaration_inside_function_call_parens` in `crates/perl-parser-core/src/engine/parser/tests.rs` that parses `foo(my $x);` and asserts the AST contains both the call and the inner `my` declaration.

### Testing
- Ran `cargo fmt --all` which completed successfully.
- Ran `cargo test -p perl-parser-core test_my_declaration_inside_function_call_parens` which passed (the new regression test succeeded).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b2175e567c833396eff80a82dc0b68)

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/primary.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/primary.rs
@@ -587,9 +587,35 @@ impl<'a> Parser<'a> {
             | TokenKind::Class
             | TokenKind::Method
             | TokenKind::Format => {
-                // In expression context, some keywords can be used as barewords/identifiers
+                // `my/our/local/state` can also appear inside expression contexts,
+                // e.g. function call parens: `foo(my $x)`.
+                // Parse these as declarations when followed by a variable/list start.
+                if matches!(
+                    token_kind,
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as duplicate of #1289 which was already merged. That PR ("fix(parser): handle my/our/local/state in call args") addresses the same my-declaration-in-call-args issue.

PR #1309, opened 2026-03-12
Fix map/grep/sort block+list parsing in postfix expressions

EffortlessSteven:
### Motivation
- Postfix parsing for builtin list-operators (`map`, `grep`, `sort`) treated a `{ ... }` block as a single-expression block in expression context, which failed for real `BLOCK` statement-lists followed by a `LIST` (e.g. `map { my $x = $_; $x * 2 } @list`).

### Description
- In the postfix argument parser, replace the ad-hoc single-expression `{ ... }` handling with a call to the dedicated `parse_builtin_block()` to correctly parse statement-list blocks for builtins.
- After consuming the builtin block, parse trailing list arguments with `parse_assignment()` (instead of `parse_ternary()`) to align behavior with the statement-level path that already handled `BLOCK LIST` correctly.
- Add regression coverage in `crates/perl-parser-core/src/engine/parser/hash_vs_block_tests.rs` with test cases exercising `map`/`grep`/`sort` in both statement and expression contexts (including blocks that contain local variable statements).
- Files changed: `crates/perl-parser-core/src/engine/parser/expressions/postfix.rs` and `crates/perl-parser-core/src/engine/parser/hash_vs_block_tests.rs`.

### Testing
- Ran `cargo test -p perl-parser-core test_map_grep_sort_blocks -- --nocapture` and the test passed successfully.
- Ran `cargo fmt --all` to format changes and re-ran the same test to ensure no regressions, and it passed.
- Tests run were automated unit tests added/updated in `perl-parser-core` and they succeeded.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b2175e6dd083338c0e9aef96556e65)

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/postfix.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/postfix.rs
@@ -397,26 +397,10 @@ impl<'a> Parser<'a> {
                                 if matches!(name.as_str(), "sort" | "map" | "grep")
                                     && self.peek_kind() == Some(TokenKind::LeftBrace)
                                 {
-                                    // Parse block expression as first argument
-                                    let block_start = self.current_position();
-                                    self.expect(TokenKind::LeftBrace)?;
-
-                                    // Parse the expression inside the block (if any)
-                                    let mut statements = Vec::new();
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #1321, opened 2026-03-12
Fix keyword autoquoting before fat arrow

EffortlessSteven:
### Motivation
- Perl treats many keywords as barewords (autoquoted) when they appear as the left side of a fat-arrow (`=>`), and the parser currently failed to handle several such cases causing syntax errors for constructs like `if => 1` used as hash keys.

### Description
- Add a shared helper `is_autoquotable_keyword` to centralize which `TokenKind` values may be treated as bareword keys in `key => value` contexts (new in `helpers.rs`).
- Update primary-expression parsing so autoquotable keywords are parsed as `Identifier` nodes when immediately followed by `FatArrow` while preserving normal keyword behavior otherwise (changes in `expressions/primary.rs`).
- Special-case `sub` to still recognize anonymous subs, and keep dedicated handling for `return`, `eval`, and `do` unless they appear in the fat-arrow autoquote position (adjusted logic between `precedence.rs` and `primary.rs`).
- Add regression tests that validate keyword autoquoting both in parenthesized list contexts and brace/hash contexts (tests added to `comprehensive_unit_tests.rs`).

### Testing
- Ran `cargo fmt --all` to format changes successfully.
- Ran `cargo test -p perl-parser-core wave2b_keyword_autoquote_before_fat_arrow -- --nocapture` and the test passed.
- Ran `cargo test -p perl-parser-core wave2b_ -- --nocapture` which exercised related fat-arrow regression tests and all affected tests passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b21835acc48333966b1cee656d2dbc)

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/precedence.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/precedence.rs
@@ -153,12 +153,6 @@ impl<'a> Parser<'a> {
             return self.parse_word_not_expr();
         }
 
-        // Handle 'return' as an expression in expression context
-        // This allows patterns like: open $fh, $file or return;
-        if self.peek_kind() == Some(TokenKind::Return) {
-            return self.parse_return();
-        }
-
         let mut expr = self.parse_ternary()?;
 
         // Check for assignment operators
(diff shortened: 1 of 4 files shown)


EffortlessSteven:

PR #1322, opened 2026-03-12
Fix keyword autoquoting before fat arrow in call args

EffortlessSteven:
### Motivation
- The parser auto-quoted bare identifiers on the left-hand side of `=>` in call argument lists but treated reserved keywords (e.g. `if`, `while`, `default`) as non-autoquotable, causing incorrect ASTs for `foo(if => 1)`-style calls.

### Description
- Add `is_autoquotable_keyword` helper and a targeted pre-check in `parse_args` to convert keyword tokens immediately followed by `FatArrow` into `NodeKind::String` nodes before normal argument parsing.
- Preserve the existing identifier auto-quoting path so `key => 1` behavior remains unchanged.
- Modify `crates/perl-parser-core/src/engine/parser/expressions/calls.rs` and add regression tests in `crates/perl-parser-core/src/engine/parser/tests.rs` for keyword and identifier auto-quoting cases.

### Testing
- Ran formatting with `cargo fmt --all` which completed successfully.
- Ran the parser unit tests with `cargo test -p perl-parser-core autoquoting_before_fat_arrow` and the new tests `test_keyword_autoquoting_before_fat_arrow_in_call_args` and `test_identifier_autoquoting_before_fat_arrow_still_works` passed (all tests in that run passed).
- Verified the new tests exercise both keyword-to-string conversion and that existing identifier autoquoting remains working.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b21835acc48333966b1cee656d2dbc)

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -1,4 +1,48 @@
 impl<'a> Parser<'a> {
+    fn is_autoquotable_keyword(kind: TokenKind) -> bool {
+        matches!(
+            kind,
+            TokenKind::My
+                | TokenKind::Our
+                | TokenKind::Local
+                | TokenKind::State
+                | TokenKind::Sub
+                | TokenKind::If
+                | TokenKind::Elsif
+                | TokenKind::Else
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #1488, opened 2026-03-15
fix(parser): handle fat arrow pairs in use/import args

EffortlessSteven:
## Summary
- Fix parsing of `use overload '""' => \&stringify` and similar patterns where string-literal keys are followed by fat arrows in import argument lists
- Extract reusable `consume_use_import_value` helper that handles numbers, strings, identifiers, anonymous subs, and complex expressions (like `\&name`) after `=>`
- Adds dedicated `use_overload_tests` module covering arithmetic, string, comparison, and dereferencing overloads

## Test plan
- [x] `cargo test -p perl-parser-core --lib` passes (222 tests)
- [ ] Verify no regressions in `cargo test --workspace --lib`


EffortlessSteven:
Closing as duplicate of #1492 — same changes, #1492 has a more thorough description with root cause and test evidence.

PR #1491, opened 2026-03-15
fix(parser): parse return/next/last/redo in expression context

EffortlessSteven:
## Summary
- Parse `return`, `next`, `last`, `redo` as real control-flow nodes when they appear in expression context (ternary branches, short-circuit operators, etc.) rather than treating them as bareword identifiers
- Add `parse_return_expr` variant that respects expression boundaries (`:`, `)`, `]`, `,`) to avoid greedy token consumption
- Keywords followed by `=>` are still autoquoted as hash keys (e.g. `return => 1`)

## Test plan
- [x] `cargo test -p perl-parser-core --lib` passes (225 tests)
- [ ] Verify no regressions in `cargo test --workspace --lib`


EffortlessSteven:
Closing as duplicate of #1483 — same changes, #1483 has more detailed description and includes a docs update.

PR #2013, opened 2026-03-19
fix(parser): handle ternary expressions in use constant declarations

### LINKED ISSUE #1895: parser: fix ternary in use constant declarations (~50 CPAN files)
`use constant FOO => EXPR ? ... : ...` fails because consume_use_import_value stops after first atom. Root cause in declarations.rs ~line 952.

EffortlessSteven:
## Summary

Fixes #1895.

- `consume_use_import_value` in `declarations.rs` had a semicolon check that wasn't depth-aware — `eval { require Foo; 1 } ? 1 : 0` would break at the inner `;` and leave orphaned tokens
- Gate the semicolon-terminates check on `at_top_level` so semicolons inside nested braces/parens/brackets are consumed as part of the value expression

## Test plan

- [x] 5 new tests added: variable condition, ENV hash, Perl version, eval block, defined() in use constant ternary values
- [x] All 12 ternary tests pass (7 existing + 5 new)
- [x] All perl-parser-core tests pass (only pre-existing `parser_ac3_prevent_recursive_recovery` failure)
- [x] `cargo fmt --all` clean
- [x] `cargo clippy -p perl-parser-core --lib` clean

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/declarations.rs
+++ b/crates/perl-parser-core/src/engine/parser/declarations.rs
@@ -956,7 +956,10 @@ impl<'a> Parser<'a> {
             let kind = self.peek_kind();
             let at_top_level = paren_depth == 0 && bracket_depth == 0 && brace_depth == 0;
 
-            if Self::is_statement_terminator(kind) {
+            // Only treat semicolons as terminators at the top level — inside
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing: has merge conflicts and parser has changed significantly. The ternary-in-use-constant fix should be re-implemented against current master if still needed.

Issue #2360, opened 2026-03-20
feat(perl): BEGIN/END/CHECK/INIT/UNITCHECK block phasing support


EffortlessSteven:
## Feature
IDE support for Perl's code execution phases:
- Distinguish BEGIN (compile-time), END (cleanup), INIT (post-compile)
- Hover explains execution order and use cases
- Diagnostic: warn if code in BEGIN that should be in INIT
- Track global side effects in BEGIN blocks

## Current State
Parser recognizes block keywords but LSP treats them as normal blocks.

## Implementation Plan
1. Add phase-aware AST nodes for block types
2. Enhance hover with phase execution semantics
3. Implement ordering checker (BEGIN must come before END)
4. Warn on non-idiomatic phase usage

## Test Plan
- Hover on BEGIN shows 'executed at compile time'
- Diagnostic warns if file I/O in BEGIN
- Phase ordering is validated

## Perl Ecosystem Context
Perl's multiple execution phases are powerful but confusing. Teaching IDE support helps even experienced developers avoid subtle phase-order bugs.

PR #2516, opened 2026-03-20
feat(semantic-analyzer): add phase-aware hover docs for BEGIN/END/CHECK/INIT/UNITCHECK blocks

EffortlessSteven:
## Summary

Perl's five execution phase blocks (BEGIN, END, INIT, CHECK, UNITCHECK) previously produced semantic highlight tokens but no hover information. Hovering over a phase block keyword in an IDE returned nothing.

This change adds `HoverInfo` (signature + documentation) for each phase in the `PhaseBlock` arm of `SemanticAnalyzer::analyze_node`. Each phase gets a concise description of when it runs, its ordering behavior (FIFO vs LIFO for multiples), and its idiomatic use.

The `phase_block_hover` helper is a pure `fn` with no side effects — it matches on the phase name string and returns `(signature, documentation)`. An `other` fallback arm handles any unexpected phase name gracefully without panicking.

## Interface & compatibility

- `perl-parser` API: unchanged
- `perl-semantic-analyzer` API: unchanged (adds hover entries to existing `hover_info` map)
- LSP surface: additive — hover responses now populated for phase blocks that previously returned nothing
- DAP surface: unchanged
- CLI: unchanged

## What changed

`crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs`:
- `NodeKind::PhaseBlock` arm now binds `phase` (was `phase: _`) and inserts a `HoverInfo` entry keyed by `node.location`
- New free function `phase_block_hover(phase: &str) -> (String, String)` returns signature and documentation for each of the five phases

`crates/perl-semantic-analyzer/tests/phase_hover_tests.rs` (new):
- 6 tests covering all five phases plus a parametric documentation-presence check
- TDD: written before the fix, all 6 failed (confirmed), then all 6 pass after

## How to review

Start at `node_analysis.rs` line ~501 — the `PhaseBlock` arm is a 6-line addition followed by the new `phase_block_hover` function at the bottom of the file. The test file exercises each phase independently.

## Evidence

```
running 6 tests
test test_end_phase_block_hover_contains_cleanup ... ok
test test_phase_block_hover_has_documentation ... ok
test test_check_phase_block_hover_present ... ok
test test_begin_phase_block_hover_contains_compile_time ... ok
test test_init_phase_block_hover_contains_post_compile ... ok
test test_unitcheck_phase_block_hover_present ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo fmt -p perl-semantic-analyzer -- --check  → PASS
cargo clippy -p perl-semantic-analyzer --lib -- -D warnings  → PASS (clean)
cargo clippy -p perl-semantic-analyzer --test phase_hover_tests -- -D warnings  → PASS (clean)
```

Pre-existing failures NOT caused by this PR:
- `attribute_introspection_tests.rs`: imports `get_attribute_documentation` which has never existed in source (pre-existing broken test)
- `perl-lsp-perltidy`: trait method arity mismatch in lib.rs (pre-existing)

## Risk & rollback

Blast radius is minimal — this is an additive change to the hover info map. If hover is not queried for a phase block, nothing changes at runtime. The `HoverInfo` struct has no side effects. Rollback: revert the 6-line insertion in the `PhaseBlock` arm and remove the helper function.

## Follow-ups

- The diagnostic for "file I/O in BEGIN block" mentioned in the issue is out of scope for this PR (would require a separate analysis pass)
- Phase ordering validation (BEGIN must come before END) is similarly out of scope

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs
@@ -498,13 +498,19 @@ impl SemanticAnalyzer {
                 }
             }
 
-            NodeKind::PhaseBlock { phase: _, phase_span: _, block } => {
+            NodeKind::PhaseBlock { phase, phase_span: _, block } => {
                 // Handle BEGIN/END/INIT/CHECK/UNITCHECK blocks
                 self.semantic_tokens.push(SemanticToken {
                     location: node.location,
                     token_type: SemanticTokenType::Keyword,
                     modifiers: vec![],
                 });
+
+                let (signature, documentation) = phase_block_hover(phase);
+                self.hover_info.insert(
+                    node.location,
+                    HoverInfo { signature, documentation: Some(documentation), details: vec![] },
+                );
 
                 self.analyze_node(block, scope_id);
             }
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
## Builder Retrospective

### What was done

Added phase-aware hover documentation for Perl's five execution phase blocks (BEGIN/END/INIT/CHECK/UNITCHECK) in `perl-semantic-analyzer`. The `PhaseBlock` arm in `analyze_node` previously pushed a semantic token but inserted nothing into `hover_info`. The fix adds a `phase_block_hover` helper that returns a signature and documentation string for each phase, then inserts a `HoverInfo` entry at the node's location. Six TDD tests were written first (all failing), then the implementation was added (all pass).

### What was harder or easier than expected

Easier: The insertion point was obvious — a single 6-line addition in one arm of a large match. The `HoverInfo` struct API was consistent with all other hover insertions in the same file, so no new patterns were needed.

Harder: The `cargo test --workspace` gate fails due to two pre-existing issues (`attribute_introspection_tests.rs` references `get_attribute_documentation` which doesn't exist, and `perl-lsp-perltidy` has a trait method arity mismatch). These cause confusion when evaluating whether the verify gate is green — the builder had to distinguish pre-existing noise from new failures.

Surprising: The `hover_info` field is `pub(super)`, so external integration tests cannot use `analyzer.hover_info.iter()` directly as some internal tests do. The test had to use `hover_at(location)` with a location found by traversing the AST — which is actually the correct LSP-realistic test pattern anyway.

1. `attribute_introspection_tests.rs` references `get_attribute_documentation` which never existed — this is a separate broken test that should be tracked as its own issue. It blocks `cargo test -p perl-semantic-analyzer` from giving a clean green.
2. The diagnostic features from issue #2360 (warn on file I/O in BEGIN, phase ordering validation) were explicitly deferred — they require a separate analysis pass and are not part of this PR.
3. The `phase_block_hover` function lives at the bottom of `node_analysis.rs` as a free function. If a reviewer wants it as an `impl` method or in a separate module, that's a cosmetic refactor.
4. The hover info is keyed by `node.location` (the full block span including braces), not by `phase_span` (just the keyword). LSP clients typically query by cursor position, and the cursor could be anywhere in the block — this is consistent with how other nodes are keyed in this file.

Issue #2708, opened 2026-03-21
feat(diagnostics): wire heredoc_antipatterns into DiagnosticsProvider::get_diagnostics


EffortlessSteven:
## Problem

The heredoc antipattern detector (\`perl-lsp-diagnostics::heredoc_antipatterns::detect_heredoc_antipatterns\`) is called directly in \`perl-lsp\`'s \`diagnostics.rs:109\` but is **not** called from \`DiagnosticsProvider::get_diagnostics()\` in the diagnostics crate itself.

This means that:
1. **Push diagnostics** (from \`publish_diagnostics\`) include heredoc warnings (\`diagnostics.rs:109\`)
2. **Pull diagnostics** (from \`handle_document_diagnostic\`) also call \`get_diagnostics\` but heredoc checks are duplicated at the LSP layer (\`diagnostics.rs:458\`)
3. Any other consumer of \`DiagnosticsProvider\` (tests, future tools) would miss heredoc antipatterns

The same pattern applies to unused_imports and security lints — issues #2694 and #2693 track those.

## Proposed Fix

Move the \`detect_heredoc_antipatterns\` call into \`DiagnosticsProvider::get_diagnostics()\` alongside the other lint calls:

```rust
// In get_diagnostics(), after check_printf_format:
let heredoc_diags = crate::heredoc_antipatterns::detect_heredoc_antipatterns(source);
diagnostics.extend(heredoc_diags);
```

Then remove the duplicate calls in the LSP layer (\`perl-lsp/src/runtime/diagnostics.rs\`).

## Quick Win Assessment

- **Effort**: Small (~3 lines added, ~3 lines removed per call site)
- **Impact**: Medium — consistent diagnostics across push and pull paths
- **Risk**: Low — same diagnostics, just centralized

Issue #2694, opened 2026-03-21
fix(diagnostics): wire check_unused_imports lint into DiagnosticsProvider::get_diagnostics()


EffortlessSteven:
## Background

The `check_unused_imports` function in `crates/perl-lsp-diagnostics/src/lints/unused_imports.rs` is fully implemented (219 LOC) with **14 tests** in a dedicated test file `crates/perl-lsp-diagnostics/tests/unused_imports_tests.rs`, but it is **never called** from `DiagnosticsProvider::get_diagnostics()` in `crates/perl-lsp-diagnostics/src/diagnostics.rs`.

Same pattern as #2538.

## What exists

- **File**: `crates/perl-lsp-diagnostics/src/lints/unused_imports.rs`
- **Entry point**: `pub fn check_unused_imports(node: &Node, source: &str, diagnostics: &mut Vec<Diagnostic>)`
- **Detects**: `use Module` statements where the module name never appears elsewhere in the source. Pragmas and implicit-export modules are excluded via `PRAGMA_SKIP_LIST` to avoid false positives.
- **Diagnostic code**: `unused-import` (Hint severity)
- **Tests**: 14 dedicated tests in `crates/perl-lsp-diagnostics/tests/unused_imports_tests.rs`
- **Module exported**: Already in `crates/perl-lsp-diagnostics/src/lints/mod.rs` as `pub mod unused_imports`

## Wiring path (~2 lines)

In `crates/perl-lsp-diagnostics/src/diagnostics.rs`:

1. Add import: `use crate::lints::unused_imports::check_unused_imports;`
2. Add call after the existing lint checks (line ~117): `check_unused_imports(ast, source, &mut diagnostics);`

## Acceptance criteria

- [ ] `check_unused_imports` is called from `get_diagnostics()`
- [ ] All 14 existing unused import tests pass
- [ ] `cargo clippy -p perl-lsp-diagnostics` clean
- [ ] Integration test: parse `use Some::Module;\nprint "hello";\n` and verify diagnostic with code `unused-import` is produced

## Related

- #2538 (wired the other 3 lints)
- #2667 (parent scan issue)

Issue #2693, opened 2026-03-21
fix(diagnostics): wire check_security lint into DiagnosticsProvider::get_diagnostics()


EffortlessSteven:
## Background

The `check_security` function in `crates/perl-lsp-diagnostics/src/lints/security.rs` is fully implemented (131 LOC) with **11 tests** in `crates/perl-lsp-diagnostics/tests/unit_tests.rs`, but it is **never called** from `DiagnosticsProvider::get_diagnostics()` in `crates/perl-lsp-diagnostics/src/diagnostics.rs`.

This is the same pattern as the previously-wired lints (`check_deprecated_syntax`, `check_strict_warnings`, `check_common_mistakes`) fixed in #2538/#2544.

## What exists

- **File**: `crates/perl-lsp-diagnostics/src/lints/security.rs`
- **Entry point**: `pub fn check_security(node: &Node, diagnostics: &mut Vec<Diagnostic>)`
- **Detects**: two-arg `open` (PL security-two-arg-open), string `eval` (PL security-string-eval), backtick exec (PL security-backtick-exec)
- **Tests**: 11 tests covering two-arg open, three-arg open (no warning), string eval, eval with variable, eval block (no warning), backtick detection, normal string (no warning), non-interpolated (no warning), other functions (no warning), eval with concatenation, related info quality
- **Module exported**: Already in `crates/perl-lsp-diagnostics/src/lints/mod.rs` as `pub mod security`

## Wiring path (~2 lines)

In `crates/perl-lsp-diagnostics/src/diagnostics.rs`:

1. Add import: `use crate::lints::security::check_security;`
2. Add call after the existing lint checks (line ~117): `check_security(ast, &mut diagnostics);`

## Acceptance criteria

- [ ] `check_security` is called from `get_diagnostics()`
- [ ] All 11 existing security lint tests pass
- [ ] `cargo clippy -p perl-lsp-diagnostics` clean
- [ ] Integration test: parse `open(FH, ">file")` and verify diagnostic with code `security-two-arg-open` is produced

## Related

- #2538 (wired the other 3 lints)
- #2667 (parent scan issue)

PR #2729, opened 2026-03-22
test(diagnostics): verify heredoc_antipatterns wiring through get_diagnostics() (#2708)

EffortlessSteven:
## Summary

Adds three pipeline-level integration tests that verify
`detect_heredoc_antipatterns()` is properly wired through
`DiagnosticsProvider::get_diagnostics()`, closing the test-coverage gap
identified in issue #2708.

The implementation itself was already present (commit 1d52d3332, PR #2438
wired the call at `diagnostics.rs:112`). What was missing was a test that
exercises the full path — real Perl source → parser → `DiagnosticsProvider`
— and asserts that PL800-PL806 heredoc antipattern codes come out the
other end. Without this test, a future refactor could silently remove the
call and no test would fail.

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: unchanged
- DAP surface: unchanged
- CLI: unchanged

## What changed

`crates/perl-lsp-diagnostics/tests/lint_pipeline_integration_tests.rs`
gains three new tests under section heading "15. Heredoc antipatterns":

- `lint_pipeline_format_heredoc_emits_pl800_via_get_diagnostics` —
  asserts that `format STDOUT =` source produces at least one PL8xx
  diagnostic via `get_diagnostics()`, confirming the wiring is live.
- `lint_pipeline_source_filter_range_valid_via_get_diagnostics` —
  validates that if PL803 fires for `use Filter::Util::Call`, its range
  is within source bounds (detector is heuristic; fires or not is OK,
  but if it fires the range must be valid).
- `lint_pipeline_clean_heredoc_emits_no_pl8xx_via_get_diagnostics` —
  asserts a plain single-quoted heredoc produces zero PL8xx diagnostics.

No production code was changed.

## How to review

Single file, append-only. The three new test functions start at line 330.
Compare with the analogous `format_heredoc_detected` test in
`heredoc_antipattern_tests.rs` — the key difference is that this test
calls `diagnostics_for()` (which uses `DiagnosticsProvider`) rather than
`detect_heredoc_antipatterns()` directly.

## Evidence

```
running 16 tests
test lint_pipeline_format_heredoc_emits_pl800_via_get_diagnostics ... ok
test lint_pipeline_source_filter_range_valid_via_get_diagnostics ... ok
test lint_pipeline_clean_heredoc_emits_no_pl8xx_via_get_diagnostics ... ok
... (13 other existing tests) ...
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured
cargo fmt: clean
cargo clippy -p perl-lsp-diagnostics --tests: clean (no warnings)
```

## Risk & rollback

Test-only change. Zero production code modified. No risk of runtime
regression. Rollback = revert the single file addition.

## Follow-ups

Issues #2693 (security lints) and #2694 (unused_imports) track the same
"built but not wired" pattern for other lint categories. Those are
separate issues and out of scope here.

### DIFF
--- a/crates/perl-lsp-diagnostics/tests/lint_pipeline_integration_tests.rs
+++ b/crates/perl-lsp-diagnostics/tests/lint_pipeline_integration_tests.rs
@@ -414,3 +414,81 @@ fn lint_pipeline_clean_heredoc_emits_no_pl8xx_via_get_diagnostics() {
         heredoc_diags.iter().map(|d| d.code.as_deref()).collect::<Vec<_>>()
     );
 }
+
+// =========================================================================
+// 16. Heredoc antipattern detection survives alongside parse errors
+//     Edge case: if the source is malformed and produces parse errors,
+//     detect_heredoc_antipatterns() must still fire — the two paths are
+//     independent in get_diagnostics().
+// =========================================================================
+
+#[test]
+fn lint_pipeline_heredoc_antipattern_survives_parse_errors() {
+    // Source has a syntax error (missing semicolon) AND a source filter that
+    // triggers PL803. Both kinds of diagnostic should appear together.
+    // This guards that the heredoc detection block (lines 111-113 of diagnostics.rs)
+    // is not short-circuited by an earlier error path.
+    let source = "use Filter::Util::Call\nuse strict;\n";
+    let diags = diagnostics_for(source);
+
+    let heredoc_diags: Vec<_> = diags
+        .iter()
+        .filter(|d| d.code.as_deref().map(|c| c.starts_with("PL8")).unwrap_or(false))
+        .collect();
+
+    assert!(
+        !heredoc_diags.is_empty(),
+        "Expected PL8xx diagnostic even when source has parse errors, \
+         got {} total diags with codes: {:?}",
+        diags.len(),
+        diags.iter().map(|d| d.code.as_deref().unwrap_or("none")).collect::<Vec<_>>()
+    );
+}
+
+// =========================================================================
+// 17. Heredoc antipattern range is always within source bounds
+//     Edge case: the offset returned by extract_offset() must never exceed
+//     source.len(). This validates the (offset, (offset+1).min(source.len()))
+//     calculation in heredoc_antipatterns.rs for any PL8xx diagnostic.
+// =========================================================================
+
+#[test]
+fn lint_pipeline_heredoc_antipattern_range_in_bounds() {
+    // Multiple antipatterns in one source — verify all emitted ranges are valid.
+    let source = "use Filter::Util::Call;\nformat REPORT =\n@<<<< @>>>>\n$name\n.\n";
+    let diags = diagnostics_for(source);
+
+    let heredoc_diags: Vec<_> = diags
+        .iter()
+        .filter(|d| d.code.as_deref().map(|c| c.starts_with("PL8")).unwrap_or(false))
+        .collect();
+
+    assert!(
+        !heredoc_diags.is_empty(),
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
## Review Wrapup

**Summary:** Deep reviewer improved PR #2729, strengthening 1 vacuous assertion and adding 2 edge case tests. 18 tests passing (up from 13 before PR).

---

### Ending State

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/2729
- Changes made: 2 commits pushed to branch `worktree-agent-a4b62f5d`
  1. `144c0e5d4` — strengthen PL803 assertion to be non-vacuous
  2. `4ebf5d611` — add edge case tests for heredoc wiring robustness (2 new tests)
- Test artifacts: `crates/perl-lsp-diagnostics/tests/lint_pipeline_integration_tests.rs`
- Status: approved (`reviewed-deep` label set, receipt written)
- Next step: ready to merge

---

### What I verified

1. **Wiring exists and is correct.** `detect_heredoc_antipatterns()` is called at `diagnostics.rs:112` inside `get_diagnostics()`. Commit `1d52d3332` added it. No LSP-layer duplicate calls exist.

2. **Tests are non-vacuous.** Removing the call at line 112 would cause all three original tests to fail. PL8xx codes are emitted only from that path — confirmed by grepping the entire `perl-lsp-diagnostics/src/` directory.

3. **Test 1 relies on a pre-existing false positive.** `format STDOUT =` with `@<<<<` format column operators fires PL800 because `@<<<<` contains `<<`. The existing lower-layer `format_heredoc_detected` test in `heredoc_antipattern_tests.rs` uses the identical input, so this is intentional and consistent.

4. **Test 2 was vacuous on empty collection.** The source filter test looped over results but would pass trivially if `source_filter_diags` were empty (e.g. if wiring were removed). Fixed by adding upfront assertion.

---

### Retrospective

**Harder than expected:** Verifying that the test binary in the PR worktree was actually compiled with the new tests — the file timestamp was 22:27 but the binary was from 23:20, and it still only had 13 tests. Had to build fresh with a separate `CARGO_TARGET_DIR` to get the correct binary.

**What surprised me:** The issue description (#2708) claimed the wiring was missing, but it had already been added by commit `1d52d3332` before the issue was filed. The PR is really a "coverage gap" fix, not an implementation fix. This is fine — the tests are genuinely useful — but the issue title says `feat` (move the call) when the actual PR is `test` (verify the call already exists). This mismatch didn't affect correctness but cost time re-reading the issue to understand why the fix wasn't in the diff.

**What I'd do differently:** Check `git log` for the relevant function FIRST before reading the issue description. The commit history told the real story faster than the issue narrative.

---

- Logical next: mark `merge-ready` and include in next merge batch (with #2693, #2694 which are tracked as follow-ups in the PR body)
- Related issues: #2693 (security lints wiring), #2694 (unused_imports wiring) — same "built but not wired" pattern, same test approach applies
- Gotchas: `FormatHeredocDetector` in `crates/perl-ts-heredoc-analysis/src/anti_pattern_detector.rs` detects `@<<<<` format column operators as false-positive `<<` heredoc markers. Any test using `format NAME =` with `@<<<` fields will fire PL800 even with no actual heredoc.
- Gotchas: `TiedHandleDetector` compiles `Regex::new()` per tied handle found — pre-existing perf issue for files with many `tie` calls.
- Traps: The worktree binary at `.claude/worktrees/agent-a4b62f5d/target/debug/deps/lint_pipeline_integration_tests-001deedaf2245553` was stale (13 tests, not 16). Always build fresh with `CARGO_TARGET_DIR=/tmp/...` when verifying PR tests.
- Confidence for next: high

Issue #2108, opened 2026-03-19
UX: Detect and warn on Windows line endings (CRLF) misalignment


EffortlessSteven:
**Problem**: Windows CRLF files are parsed but positions get misaligned between LSP (CRLF-aware) and diagnostic display (LF-only), causing wrong line/col in error messages.

**Current behavior**:
- File with CRLF opens normally but hover/completion positions are off-by-N
- Diagnostics point to wrong line if file has mixed CRLF/LF
- User can't tell if LSP is broken or their file is broken

**Location**: crates/perl-lsp/src/textdoc.rs (position encoding logic)

**Expected improvement**:
- Detect CRLF on document open
- Add diagnostic message: "File has Windows line endings (CRLF), converting to LF for consistency"
- Auto-normalize to LF (like most modern editors)
- Or: document that CRLF is fully supported (if already handled correctly)

**Success criteria**:
- CRLF file positions are correct (verified in hover/go-to-def tests)
- User sees message if conversion is needed
- No position drift or misalignment

PR #2736, opened 2026-03-22
fix(position): add CRLF guard in position_to_offset_rope (#2108)

EffortlessSteven:
## Summary

`position_to_offset_rope` in `LineStartsCache` walked into `\r\n` terminator bytes when the requested column was at or past end of line content. For a line like `"ab\r\n"`, requesting `character=100` would advance through `a`, `b`, `\r`, `\n`, and return byte offset 4 — a position inside the CRLF terminator. This is not a valid edit position and would produce misaligned byte spans in the parser.

The `&str` variant (`position_to_offset`) already had a correct guard that strips line terminators before the column walk (lines 50-54 of `line_index.rs`). This fix makes the rope-backed variant consistent: add `if ch == '\r' || ch == '\n' { break; }` inside the column walk, so the walk stops at end-of-content rather than stepping through the terminator bytes.

The fix is 3 lines. No API changes, no normalization, no user-facing warnings.

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: unchanged
- DAP surface: unchanged
- CLI: unchanged

## What changed

`crates/perl-position-tracking/src/line_index.rs` — `position_to_offset_rope`: added line-terminator guard in column walk (consistent with existing `position_to_offset` guard for the `&str` path).

`crates/perl-position-tracking/tests/comprehensive_unit_tests.rs` — added 4 regression tests for the rope-backed CRLF path:
- `line_starts_cache_rope_crlf_basic` — basic line offset correctness with CRLF
- `line_starts_cache_rope_crlf_roundtrip` — roundtrip for all content bytes (skips `\r\n` terminator bytes which have no unique inverse in CRLF model)
- `line_starts_cache_rope_crlf_position_at_eol` — key regression: EOL and over-large column both clamp to end of content, never inside `\r\n`
- `line_starts_cache_rope_mixed_line_endings` — CRLF and LF mixed in one document

## How to review

Start at `crates/perl-position-tracking/src/line_index.rs` line 94-102. Compare `position_to_offset` (line 62-68) which has the guard, versus `position_to_offset_rope` which did not. The fix is the 3-line diff adding the same guard pattern.

The `line_starts_cache_rope_crlf_position_at_eol` test is the canonical regression test — it demonstrates the before/after behavior directly.

## Evidence

```
running 214 tests
...
test line_starts_cache_rope_crlf_basic ... ok
test line_starts_cache_rope_crlf_position_at_eol ... ok
test line_starts_cache_rope_crlf_roundtrip ... ok
test line_starts_cache_rope_mixed_line_endings ... ok
...
test result: ok. 214 passed; 0 failed; 0 ignored; 0 measured

running 146 tests
...
test result: ok. 146 passed; 0 failed; 0 ignored; 0 measured

Doc-tests: 6 passed

cargo fmt --check: clean
cargo clippy -p perl-position-tracking --tests: clean
```

**Known pre-existing gate failure (not caused by this PR):**
`just ci-gate` fails at `clippy-no-unwrap-all` on `xtask/src/tasks/features.rs:448` — an `expect_err()` call that exists identically on `origin/master`. This is not introduced by this change.

## Risk & rollback

Blast radius: `perl-position-tracking` only. The fixed function `position_to_offset_rope` is called when converting LSP `character` positions to byte offsets in rope-backed text. The change makes it more conservative — it stops sooner. Any code that previously received an offset pointing inside `\r\n` would have produced incorrect results; after this fix it receives the byte at end-of-content instead.

Rollback: revert the 3-line change to `position_to_offset_rope`.

## Follow-ups

None required. The plan-review spec explicitly called out `offset_to_position_rope` (line 77) as correctly handling CRLF for the reverse direction — no fix needed there.

### DIFF
--- a/crates/perl-position-tracking/tests/comprehensive_unit_tests.rs
+++ b/crates/perl-position-tracking/tests/comprehensive_unit_tests.rs
@@ -1900,6 +1900,17 @@ fn line_starts_cache_rope_crlf_roundtrip() {
         let back = cache.position_to_offset_rope(&rope, line, col);
         assert_eq!(back, byte, "CRLF rope roundtrip failed at byte {byte}");
     }
+
+    // Direct over-column regression: character=999 on "hello" (line 0) must
+    // clamp to byte 5 (end of content), not walk into \r\n at bytes 5-6.
+    // This is the specific bug that was fixed — it would return byte 7 before the fix.
+    let off = cache.position_to_offset_rope(&rope, 0, 999);
+    assert_eq!(off, 5, "over-column on CRLF line must clamp to end of content, not inside \\r\\n");
+
+    // Over-column on "world\r\n" (line 1): content ends at byte 12, \r\n at 12-13.
+    // Before fix: would return 14 (past \r\n). After fix: returns 12.
+    let off2 = cache.position_to_offset_rope(&rope, 1, 999);
+    assert_eq!(off2, 12, "over-column on second CRLF line must clamp to byte 12");
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-2736",
  "current_version": {
    "sha": "45b1870959e2ab959fbfd5f99c8808cb44a68e37",
    "updated_at": "2026-03-22T03:39:49Z"
  },
  "label_bindings": [
    {
      "label": "needs-deep-review",
      "bound_at_version": "45b1870959e2ab959fbfd5f99c8808cb44a68e37",
      "bound_at_timestamp": "2026-03-22T04:29:52Z",
      "bound_by_agent": "reviewer",
      "valid": true
    }
  ]
}
```

PR #2740, opened 2026-03-22
feat(lsp): wire opt-in parse error telemetry into diagnostic pipeline

### LINKED ISSUE #2137: feat(lsp): opt-in error telemetry for parser crash reporting
**Goal:** Learn which Perl constructs crash the parser in real-world usage.

**Behavior (opt-in only):**
- Setting: `perl-lsp.telemetry.enabled` (default: false)
- On parse error, collect: error bucket name, file extension, approximate file size, Perl version
- Do NOT collect: file content, file path, user identity
- Send to a simple endpoint (or just log to file for manual submission)

**Privacy:** Must be opt-in, clearly documented, minimal data, no PII.

**Why:** We can't improve what we can't measure. Knowing which error buckets hit real users most tells us where to focus.

**Effort:** 6-8 hours. **Priority:** Medium (post-alpha, needs privacy review).

EffortlessSteven:
Closes #2137

## Summary

Wires `send_telemetry()` into `publish_diagnostics()` so that when `perl.telemetry.enabled` is true, every document with parse errors emits a `telemetry/event` LSP notification containing privacy-safe metadata. The feature was previously built (config flag, transport, docs, feature catalog) but never called from production code — this PR adds the missing wiring.

Rate-limited to one event per URI per 60 seconds using a per-server `telemetry_cooldowns` HashMap to prevent flooding during rapid keystroke-driven reparses. The implementation is fire-and-forget — a transport failure silently returns rather than disrupting the user's editing session.

Also fixes a latent bug in `TestWriter` (test support infrastructure) where the outbound writer's message coalescing caused batched notifications to be silently dropped. The fix uses `ContentLengthFramer` to properly classify all messages in a multi-message write — this improves reliability for all tests that use the harness.

## Interface & compatibility

- **perl-parser API**: unchanged
- **LSP surface**: additive — new `telemetry/event` notifications when `perl.telemetry.enabled=true` (default false, so no behavior change for existing users)
- **DAP surface**: unchanged
- **CLI**: unchanged

## What changed

**Production code** (`crates/perl-lsp/src/runtime/`):
- `mod.rs`: Added `telemetry_cooldowns: Arc<Mutex<HashMap<String, Instant>>>` field to `LspServer`
- `constructors.rs`: Initialized the field in all 3 constructors (`new`, `with_io`, `with_output`)
- `diagnostics.rs`: Added `maybe_send_parse_telemetry()` private method and call site in `publish_diagnostics()` — after diagnostics are computed, before the `eprintln!` log line

**Test infrastructure** (`crates/perl-lsp/tests/support/`):
- `message_framing.rs`: `TestWriter` now uses `ContentLengthFramer` to parse all messages from a batched write, fixing silent message loss when outbound coalescing is active
- `lsp_harness.rs`: Updated `TestWriter` construction to use `new()` constructor

**New tests** (`crates/perl-lsp/tests/lsp_parse_telemetry_tests.rs`):
- 5 integration tests covering: send when enabled, silent when disabled (default), silent for clean parse, rate limiting (second event suppressed within cooldown), PII exclusion (no path components or source code in payload)

## How to review

Start with `diagnostics.rs` — the call site at line 203 and the `maybe_send_parse_telemetry` method (line 714-791). The method is self-contained with clear early-return guards in order: no errors → disabled → rate-limited → build payload → fire-and-forget send.

The `message_framing.rs` change is the most structurally interesting: the original `TestWriter` had a latent bug where it called `serde_json::from_str` on raw bytes that might include `Content-Length` headers and multiple JSON objects from a coalesced write. Any batch with >1 message would silently fail to parse. The fix feeds bytes through `ContentLengthFramer` first.

## Evidence

```
running 5 tests
test test_parse_telemetry_no_pii_in_payload ... ok
test test_parse_telemetry_not_sent_for_clean_parse ... ok
test test_parse_telemetry_not_sent_when_disabled ... ok
test test_parse_telemetry_rate_limited ... ok
test test_parse_telemetry_sent_when_enabled ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.17s

Full suite (lib): 252 passed; 1 failed (pre-existing: diagnostic_data_for_parse_error expects PL001, gets PL100 — unrelated to this PR, present on master before changes)

cargo fmt: clean
cargo clippy --lib: clean
```

## Risk & rollback

- **Blast radius**: Low. The entire code path is gated behind `telemetry_enabled` (default false). Users who never set `perl.telemetry.enabled=true` see zero behavior change.
- **Failure modes**: `send_telemetry()` failure is silently ignored via `let _ = ...`. Rate limiter prevents flooding. No panic paths introduced.
- **Rollback**: Revert the 3 production-code files. The `TestWriter` fix is safe to keep independently.

## Follow-ups

- Remote telemetry collection endpoint (out of scope for v1 per plan-review)
- Perl version detection for telemetry payload (not available in codebase)
- Pull-diagnostic path (`workspace/diagnostic`) telemetry — v1 only covers push path
- Periodic cooldowns cleanup to prevent unbounded HashMap growth in long-running sessions

### DIFF
--- a/crates/perl-lsp/src/runtime/diagnostics.rs
+++ b/crates/perl-lsp/src/runtime/diagnostics.rs
@@ -718,14 +718,23 @@ impl LspServer {
         degradation_tier: crate::state::DegradationTier,
         file_size: usize,
     ) {
-        // Early return if no parse errors
+        // Fast path: skip if no errors at all (most document edits are clean)
         if parse_errors.is_empty() {
             return;
         }
 
         // Check config: telemetry must be explicitly enabled (opt-in, default false)
+        // This is the cheapest meaningful check — do it before any allocation.
         let enabled = self.config.lock().telemetry_enabled;
         if !enabled {
+            return;
+        }
+
+        // Filter out Cancelled — a cancelled parse is not a user-visible error
+        // and should not be recorded in telemetry as a parser failure.
+        let has_real_errors =
+            parse_errors.iter().any(|e| !matches!(e, crate::error::ParseError::Cancelled));
+        if !has_real_errors {
             return;
         }
 
@@ -748,20 +757,23 @@ impl LspServer {
             }
         }
 
-        // Extract file extension ONLY — no path components (privacy-safe)
+        // Extract file extension ONLY — no path components (privacy-safe).
+        // Guard: empty, path-containing, or overly-long extensions fall back to "unknown".
         let extension = uri
             .rsplit('.')
             .next()
-            .filter(|ext| ext.len() <= 10 && !ext.contains('/'))
+            .filter(|ext| !ext.is_empty() && ext.len() <= 10 && !ext.contains('/'))
             .unwrap_or("unknown");
 
-        // Classify errors by variant name — no message content or locations.
+        // Classify real errors by variant name — no message content or locations.
+        // Cancelled is excluded: it is a transport-level event, not a parse failure.
         // Deduplicated: a file with 50 UnexpectedToken errors sends "UnexpectedToken"
-        // once. errorCount covers the total; the set covers the type distribution.
-        let error_types: Vec<&str> = {
-            let mut seen = std::collections::HashSet::new();
-            let mut deduped = Vec::new();
-            for variant in parse_errors.iter().map(|e| match e {
+        // once. errorCount covers the real (non-Cancelled) total.
+        let mut seen = std::collections::HashSet::new();
+        let mut error_types: Vec<&str> = Vec::new();
+        let mut real_error_count: usize = 0;
+        for e in parse_errors {
+            let variant = match e {
                 crate::error::ParseError::UnexpectedEof => "UnexpectedEof",
                 crate::error::ParseError::UnexpectedToken { .. } => "UnexpectedToken",
                 crate::error::ParseError::SyntaxError { .. } => "SyntaxError",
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
## Deep Review: Approved with improvements

### What was verified

**Logic correctness:**
- Wiring path confirmed: `publish_diagnostics` -> `maybe_send_parse_telemetry` -> `send_telemetry` -> `outbound.send_notification`. All three constructors initialize `telemetry_cooldowns`.
- Config key chain verified: test sends `settings.perl.telemetry.enabled`, handler extracts `settings.perl`, `update_from_value` reads `telemetry.enabled` — correct end-to-end.
- Rate limiter is atomic: check-and-insert are inside the same mutex lock scope, no race condition.
- `Instant::duration_since` on a future instant saturates to 0 (not a panic) — suppresses correctly on clock skew.
- `Cancelled` parse errors cannot reach `publish_diagnostics` in practice (text_sync.rs returns early) — confirmed safe.
- `TestWriter` ContentLengthFramer fix is correct: framer state persists across `write()` calls via `&mut self`, serialized by the outer `Arc<Mutex>`.
- All 5 original tests are non-vacuous — assertions exercise the running server, not hardcoded data.

**Privacy correctness:**
- URI path components are filtered by the `contains('/')` guard in extension extraction.
- Error variant names are string literals — zero user-supplied content in the payload.
- `Cancelled` exclusion prevents transport events from appearing as parser failures.

### Improvements pushed (2 commits)

**Commit 6b1623136** — `fix(telemetry): deduplicate errorTypes and add cooldown HashMap cleanup`
- `errorTypes` was a Vec with duplicates (50x `UnexpectedToken` -> 50 identical strings). Now a deduplicated set; `errorCount` still carries the total. Reduces payload size and eliminates count-based fingerprinting.
- Cooldown HashMap cleanup at >1000 entries, retain entries <120s old — prevents unbounded growth in long-running sessions.
- Adds test 6: `test_parse_telemetry_error_types_deduplicated` — opens 3-error document, asserts no duplicate variant names and `errorCount >= errorTypes.len()`.

**Commit 9b2a18f06** — `fix(telemetry): exclude Cancelled errors and guard empty extension`
- `Cancelled` is now excluded from `error_types` and `errorCount` via `continue` in the match loop. `has_real_errors` guard prevents Cancelled-only parses from poisoning the cooldown map.
- Empty URI edge case: added `!ext.is_empty()` to the extension filter — empty string no longer passes as a valid extension (would have produced `fileExtension: ""`).
- Reordered early-return guards: `is_empty()` -> `telemetry_enabled` (mutex, O(1)) -> `has_real_errors` (O(n)) — optimal for the common case (telemetry disabled).

### Regression risk: LOW
All changes are inside `maybe_send_parse_telemetry` which is gated by `telemetry_enabled = false` (default). Zero behavior change for users who have not opted in.

Issue #2325, opened 2026-03-20
feat(perl): Moose/Moo role composition detection and navigation


EffortlessSteven:
## Feature
Understand Perl OOP role composition via Moose/Moo 'with' declarations. When a class does 'with RoleName', provide:
- Hover documentation on the role name
- Go to definition on role name
- Symbol completion showing available role methods from composition

## Current State
ClassModel (PR #2284) detects Moose/Moo frameworks and has structure for roles, but role composition is not yet wired to hover/goto/completion.

## Implementation Plan
1. Wire role detection to hover provider in perl-lsp-feature-definition
2. Add role method resolution to completion engine
3. Test with real Moose::Role and Moo::Role examples

## Test Plan
- Hover on 'with' role name shows role documentation
- Goto definition navigates to role file
- Completion after -> shows role methods

## Perl Ecosystem Context
Role composition is central to Moose/Moo. Developers need IDE support to discover available methods from composed roles.

PR #2745, opened 2026-03-22
feat(hover): Moose/Moo role composition hover and goto-def (Issue #2325)

EffortlessSteven:
## Summary

When a Perl class uses `with 'RoleName'` or `extends 'ParentClass'` (Moose/Moo OOP patterns),
hovering over the role/parent name now shows module documentation (the same hover that `use Module`
produces) instead of the generic token fallback (`**Perl**: \`RoleName\``).

Goto-definition for the same patterns already worked via the FQN regex in `navigation.rs`
(which matches `Module::Name` anywhere in source text). Tests were added to confirm and lock
that behavior.

The root cause of the hover gap was that `parse_expression_statement` captures only the keyword
span for the outer `ExpressionStatement` node (e.g., bytes 30–34 for `with`), not the full
statement including arguments. The previous implementation gated traversal on the outer node's
location, so a cursor at offset 36 (inside `'MyApp::Role::Printable'`) was skipped because
36 > 34. The fix walks `Program`/`Block` statement lists unconditionally for `with`/`extends`
patterns, checking the inner `FunctionCall` args' locations directly.

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: hover for `with`/`extends` role names now returns module info instead of token fallback
- DAP surface: unchanged
- CLI: unchanged

## What changed

`crates/perl-lsp/src/runtime/language/hover.rs`:
- Added `find_with_module_at_offset` — walks AST `Program`/`Block` statements looking for
  `ExpressionStatement { FunctionCall { name: "with"/"extends", args } }` (FunctionCall form)
  or the two-statement form `Identifier("with")` followed by `String/ArrayLiteral`.
  Returns the role/parent name string when `offset` falls within a string arg node.
- Added `role_name_at_offset` — extracts module name from `String` (strips surrounding quotes)
  or recurses into `ArrayLiteral` elements for multi-role `with 'A', 'B'`.
- `handle_hover`: inserts `find_with_module_at_offset` check between the existing
  `find_use_module_at_offset` check and the generic `extract_symbol_hover` fallback.

`crates/perl-lsp/tests/hover_provider_coverage.rs`:
- `test_hover_on_with_role_shows_module_hover` — single-role `with 'Module'`
- `test_hover_on_with_multi_role_shows_module_hover` — multi-role `with 'A', 'B'`
- `test_hover_on_extends_parent_shows_module_hover` — `extends 'Parent'`
- `test_hover_on_with_keyword_does_not_trigger_module_hover` — cursor on keyword, not name

`crates/perl-lsp/tests/cross_file_goto_definition_tests.rs`:
- `go_to_definition_on_with_role_navigates_to_role_file` — confirms existing goto-def works
- `go_to_definition_on_extends_parent_navigates_to_parent_file` — same for `extends`

## How to review

Start at `find_with_module_at_offset` in `hover.rs` (~line 312). The key design decision is
avoiding the outer `ExpressionStatement` location gate. The `role_name_at_offset` helper is
straightforward — only `String` and `ArrayLiteral` nodes are checked.

The goto-def tests in `cross_file_goto_definition_tests.rs` confirm no code change was needed
there; they validate that the FQN regex path already covers these patterns.

## Evidence

```
test hover_provider_tests::test_hover_on_with_role_shows_module_hover ... ok
test hover_provider_tests::test_hover_on_with_multi_role_shows_module_hover ... ok
test hover_provider_tests::test_hover_on_extends_parent_shows_module_hover ... ok
test hover_provider_tests::test_hover_on_with_keyword_does_not_trigger_module_hover ... ok
test go_to_definition_on_with_role_navigates_to_role_file ... ok
test go_to_definition_on_extends_parent_navigates_to_parent_file ... ok

test result: ok. 65 passed; 0 failed; 3 ignored (hover_provider_coverage)
test result: ok. 9 passed; 0 failed; 0 ignored (cross_file_goto_definition_tests)
```

Pre-existing failure in `features::diagnostics::pull::tests::diagnostic_data_for_parse_error`
(lib unit test) confirmed to fail on master as well — not introduced by this PR.

## Risk & rollback

- Blast radius: hover only. If the new path returns `None` for any input, the code falls through
  to `extract_symbol_hover` (the previous behavior), so degradation is graceful.
- No new `unwrap`/`expect`/`panic` introduced.
- Rollback: revert the `else if` block in `handle_hover` and remove the two new functions.

## Follow-ups

- `q{...}` and `qq{...}` style string delimiters in `with q{RoleName}` are not handled by
  `role_name_at_offset` (only `'` and `"` are stripped). Recommend a follow-up scout.
- `Moo::Role`'s `requires` keyword is similarly unhandled — out of scope for this issue.

### DIFF
--- a/crates/perl-lsp/tests/hover_provider_coverage.rs
+++ b/crates/perl-lsp/tests/hover_provider_coverage.rs
@@ -1089,7 +1089,8 @@ with 'MyApp::Printable';
         let code = "package MyApp::User;\nuse Moo;\nwith \"MyApp::Printable\";\n1;\n";
         let resp = hover_at(code, "file:///role_hover_dq.pl", "MyApp::Printable", 2)?;
 
-        let content = hover_content(&resp).ok_or("expected hover content for double-quoted role")?;
+        let content =
+            hover_content(&resp).ok_or("expected hover content for double-quoted role")?;
         assert!(
             content.contains("MyApp::Printable"),
             "hover on double-quoted role should contain module name, got: {content}"
@@ -1113,8 +1114,7 @@ with 'MyApp::Printable', 'MyApp::Serializable';
         // Hover on the SECOND role name on line 2.
         let resp = hover_at(code, "file:///multi_role_second.pl", "MyApp::Serializable", 2)?;
 
-        let content =
-            hover_content(&resp).ok_or("expected hover content for second role name")?;
+        let content = hover_content(&resp).ok_or("expected hover content for second role name")?;
         assert!(
             content.contains("MyApp::Serializable"),
             "hover on second role in multi-role with should contain module name, got: {content}"


EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-2745",
  "current_version": {
    "sha": "e447cdccddba44810b63c14b90e48fb3ed056c26",
    "updated_at": "2026-03-22T05:20:53Z"
  },
  "label_bindings": [
    {
      "label": "reviewed-deep",
      "bound_at_version": "e447cdccddba44810b63c14b90e48fb3ed056c26",
      "bound_at_timestamp": "2026-03-22T05:21:20Z",
      "valid": true
    }
  ]
}
```

Issue #2750, opened 2026-03-22
parser: decompose unclosed_paren — 99 system + 66 CPAN files (165 total) (follow-up to #2390)


EffortlessSteven:
# Scout Report: unclosed_paren Decomposition

## Overview

`unclosed_paren` is a **large catch-all error bucket** representing unmatched open parenthesis in various contexts. The system corpus baseline shows **99 files**, and the CPAN corpus shows **66 files** — total **165 files** needing investigation.

## Statistics

| Metric | Value |
|--------|-------|
| System corpus files | 99 |
| CPAN corpus files | 66 |
| **Total affected** | **165** |
| Unique modules (est.) | 40+ |
| % of total errors (system) | 10.2% |

## Prior Work

- Issue #2390 was filed (2026-03-20) scoped to 30 CPAN files
- PR #2453 merged with fused x<digits> token fix (2026-03-20)
- Test file `fix_unclosed_paren_patterns.rs` covers:
  - Fused `x4`, `x10` operator patterns
  - Multi-line sprintf/printf arguments
  - Heredoc in paren context
  - Method chaining across lines
  - Multi-line unless/if conditions

All 17 tests in that file **pass cleanly on master**, indicating those sub-patterns are FIXED.

## Problem Statement

Despite #2390 fixing specific patterns, the system baseline STILL shows 99 files with `unclosed_paren` errors. This suggests:

1. **Regressions or cascading errors** — the 30 files in #2390 may not have all been fixed
2. **Additional sub-patterns** — patterns beyond those tested in `fix_unclosed_paren_patterns.rs`
3. **Corpus growth** — new files added to system corpus since the fix

## Likely Sub-Causes (Speculative)

Given the fix covered fused tokens, heredocs, and method chains, remaining failures likely involve:

1. **QW-list contexts** — `qw(multiple words on same line)` with no proper closing
2. **String interpolation edge cases** — ${ $[@array] } in paren context
3. **Nested structure parsing** — hash-of-array-of-hash edge cases
4. **Complex expressions with operators** — `some_func( expr1 <=> expr2 )` spaceship operator
5. **Context-sensitive barewords** — `print FILEHANDLE LIST` paren handling
6. **Format/report blocks** — `format STDOUT =` inside parens
7. **Prototype parsing** — `sub foo ( $ \@ % ) {}` with complex signatures
8. **Regex in nested parens** — `grep { /pattern/ } @list` inside larger paren

[...]

PR #2769, opened 2026-03-22
fix(lexer): guard try_heredoc against ExpectOperator mode — fixes bitshift-in-paren (#2750)

EffortlessSteven:
## Summary

`<<` inside a parenthesized expression after a numeric or expression term (e.g.
`(1<<index($flags, $_))`) was misidentified as a heredoc start. `try_heredoc()`
in `crates/perl-lexer/src/lib.rs` did not check `self.mode` — when the token `1`
advanced the lexer to `ExpectOperator` mode, the subsequent `<<index` was still
classified as heredoc `<<index` instead of the left-shift operator. This caused
parse errors reported as `"expected ')', found heredoc (<<)"` for any expression
of the form `(N<<func(...))`.

The fix is a single mode guard added at the top of `try_heredoc()`: return `None`
immediately when `self.mode == LexerMode::ExpectOperator`. Actual heredocs in paren
context (`(<<END)`) are unaffected because `(` resets the mode to `ExpectTerm`
before the heredoc token is lexed — the guard only fires after a completed term.

This is Pattern A from the plan-review on issue #2750 — the lowest-risk fix in the
`unclosed_paren` decomposition bucket. Fixes `Devel/Peek.pm` and `ExtUtils/MM_Any.pm`
corpus modules.

## Interface & compatibility
- perl-parser API: unchanged
- LSP surface: unchanged
- DAP surface: unchanged
- CLI: unchanged

## What changed

`crates/perl-lexer/src/lib.rs` — 7 lines added to `try_heredoc()`:
```rust
if self.mode == LexerMode::ExpectOperator {
    return None;
}
```

`crates/perl-parser-core/tests/fix_unclosed_paren_bitshift.rs` — new test file, 10 tests:
- 4 failing-before-fix cases (primary reproducers)
- 6 regression guards (heredocs in paren, statement-level bitshift, chained bitshift, indented heredoc)

## How to review

- Start at `crates/perl-lexer/src/lib.rs:1001` (`try_heredoc`). The guard is the first thing in the function.
- Read `crates/perl-parser-core/tests/fix_unclosed_paren_bitshift.rs` for the test coverage. The `test_*_regression` tests are the key safety net.
- The edge cases from plan-review are all covered: `<<~identifier` (indented heredoc guard fires before `~` check), `($a << $b << $c)` (both operators after `$a` sets `ExpectOperator`), `(<<END)` (paren resets to `ExpectTerm`).

## Evidence

```
running 10 tests
test test_bitshift_at_statement_level_regression ... ok
test test_bitshift_chained_regression ... ok
test test_bitshift_in_if_condition ... ok
test test_bitshift_in_paren_list ... ok
test test_bitshift_in_paren_with_index_call ... ok
test test_bitshift_in_return ... ok
test test_bitshift_with_variable_rhs_regression ... ok
test test_heredoc_as_first_arg_regression ... ok
test test_heredoc_in_list_regression ... ok
test test_indented_heredoc_in_paren_regression ... ok

test result: ok. 10 passed; 0 failed; 0 ignored

perl-lexer: fmt PASS, clippy PASS, test PASS (325 tests)
perl-parser-core: new test file compiles and all 10 tests green
```

## Risk & rollback

**Risk: LOW** (confirmed by plan-reviewer). The guard only fires in `ExpectOperator` mode, which means a term has already been parsed. Heredoc detection at term position (`ExpectTerm`) is completely unaffected. The 6 regression tests cover the critical edge cases.

**Failure modes**: None identified. The mode check is a simple enum comparison with no side effects.

**Rollback**: Revert the 7-line guard in `try_heredoc()`. No schema or interface changes.

## Follow-ups

Per plan-review on #2750, Patterns B–E and F are separate PRs:
- Pattern D+E (calls.rs `parse_declaration_arg` — `our $X && $y`) — medium risk
- Pattern C (postfix.rs sort `$cmp` in parens) — medium risk
- Pattern B (qualified call with literal arg) — medium risk
- Pattern F (`->()` + `qr//` interaction) — high risk, needs sub-issue first

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -3613,4 +3613,29 @@ mod tests {
         assert_eq!(token.token_type, TokenType::Division);
         Ok(())
     }
+
+    #[test]
+    fn test_peek_token_does_not_mutate_paren_depth() -> TestResult {
+        // Regression guard for issue #2750: peek_token() must save and restore
+        // paren_depth so that a peek at `(` does not permanently increment
+        // paren_depth and corrupt the heredoc/bitshift guard on a subsequent token.
+        let mut lexer = PerlLexer::new("(1<<2)");
+        assert_eq!(lexer.paren_depth, 0, "paren_depth must start at 0");
+
+        // Peek at `(` — must not permanently increment paren_depth
+        let peeked = lexer.peek_token().ok_or("peek at ( failed")?;
+        assert_eq!(peeked.token_type, TokenType::LeftParen);
+        assert_eq!(lexer.paren_depth, 0, "peek_token must not mutate paren_depth");
+
+        // Consume `(` — paren_depth becomes 1
+        lexer.next_token();
+        assert_eq!(lexer.paren_depth, 1);
+
+        // Peek at `1` (a number) — paren_depth must remain 1
+        let peeked2 = lexer.peek_token().ok_or("peek at 1 failed")?;
+        assert!(matches!(peeked2.token_type, TokenType::Number(_)));
+        assert_eq!(lexer.paren_depth, 1, "peek at number must not change paren_depth");
+
+        Ok(())
+    }
 }


EffortlessSteven:
## Standards Review: CRITICAL ISSUE FOUND

The simple `ExpectOperator` guard in this PR regresses the `print $fh <<END` pattern.

**Problem**: The guard blocks heredoc recognition whenever `mode == ExpectOperator`, but this mode is also set by `print $fh <<END` where `<<END` is legitimately a heredoc at statement level.

**Evidence**: The follow-up commit 0b263917b on the PR branch shows the regression was discovered and fixed by adding `paren_depth` tracking and narrowing the guard to `mode == ExpectOperator && paren_depth > 0`.

**Recommendation**: 
1. This PR should include BOTH commits (8477e6d56 + 0b263917b) as a single logical fix
2. OR the scope should be explicitly clarified: is this Pattern A only (simple guard) or the complete Pattern A+B (with paren_depth)?

**Do not merge** without the paren_depth refinement.

### What the Complete Fix Requires
- Add `paren_depth: usize` field to PerlLexer
- Track open/close parens
- Narrow guard to: `if self.mode == LexerMode::ExpectOperator && self.paren_depth > 0`
- Add 5 edge-case tests including print-fh heredoc regression guard

I can run `/reviewer-decide` to route to deep-review once this is clarified.

EffortlessSteven:
**Deep review complete** — found and fixed a regression before merge.

### What was wrong

The original guard `if self.mode == LexerMode::ExpectOperator { return None; }` was too broad. It blocked `(1<<index(...))` from being misclassified as a heredoc (correct), but also blocked legitimate heredoc recognition in `print $fh <<END` (regression). In both cases, `$fh`/`1` set `ExpectOperator` before `<<`.

Confirmed the regression by writing a test against the original guard — it failed. The same test passes cleanly on master (no guard).

### Fix pushed (commit `0b263917b`)

Added `paren_depth: usize` to `PerlLexer` (incremented on `(`, decremented via `saturating_sub` on `)`), threaded through `checkpoint()`, `restore()`, and `reset()`. Guard narrowed to:

```rust
if self.mode == LexerMode::ExpectOperator && self.paren_depth > 0 {
    return None;
}
```

This correctly blocks `(1<<index(...))` (paren_depth=1) while preserving `print $fh <<END` (paren_depth=0). Since `(` already resets mode to `ExpectTerm`, the `(<<END)` pattern is unaffected by either the old or new guard.

### Tests added (5 new, now 15 total)

- `test_left_shift_assign_in_paren` — `<<=` after term in parens
- `test_left_shift_assign_in_paren_expr` — `($flags <<= $n)`
- `test_bitshift_after_array_subscript` — `]` also sets ExpectOperator
- `test_print_variable_fh_heredoc_regression` — the regression I found and fixed
- `test_heredoc_after_ternary_regression` — `$cond ? <<END : "default"`

### Evidence

15/15 new tests pass. All existing heredoc tests, lexer tests, and parser-core suites pass. Clippy and fmt clean. `reviewed-deep` label set.

EffortlessSteven:
**Deep review complete — approved with improvements.**

## What I verified

**Root cause correct.** The `ExpectOperator && paren_depth > 0` guard is the right narrowing of the original too-broad `ExpectOperator` guard. After `(`, `paren_depth` becomes 1; after `)`, it decrements (saturating). `print $fh <<END` has `paren_depth == 0` at the point where `<<` is seen, so the guard correctly does not fire. `(1<<func())` has `paren_depth == 1`, so it correctly fires.

**paren_depth tracking is complete across all lifecycle paths.** Incremented on `(`, decremented (saturating) on `)`, saved/restored through `save_checkpoint`/`restore_checkpoint`, zeroed in `reset()` and on checkpoint-edit-invalidation. `[` and `{` correctly do not touch paren_depth — the outer `(` already sets depth > 0, so subscript expressions inside parens are covered.

**The parser is unaffected by the `peek_token` gap** (see commit below) — the parser pre-tokenizes via `next_token()` in a loop and never calls `PerlLexer::peek_token()`.

**Pre-existing `find_with_no_chdir` failure** also fails on master; not introduced here.

## Fixes pushed to branch (2 commits)

**df7b8f96f** — Two additional edge-case tests: `test_bitshift_after_array_subscript_in_paren` and `test_bitshift_after_hash_subscript_in_paren`. These cover `($arr[0] << func())` and `($hash{key} << $n)` — cases where `[`/`{` reset mode to `ExpectOperator` inside a paren context without touching `paren_depth`, confirming the guard still fires from the outer `(`.

**1bb505384** — Fixed `peek_token()` in `crates/perl-lexer/src/lib.rs`: `paren_depth` was added to `PerlLexer` by this PR but was omitted from the manual save/restore block in `peek_token()`. Calling `peek_token()` while positioned at `(` would permanently increment `paren_depth`. Fixed by adding `saved_paren_depth` to the save/restore pattern. Added unit test `test_peek_token_does_not_mutate_paren_depth`.

## Final state

17 fix-targeted tests + 1 lexer unit test, all passing. Full `perl-lexer` and `perl-parser-core` suites green. Regression risk: LOW.

Setting `reviewed-deep` label.

Issue #2337, opened 2026-03-20
feat(perl): DBI/DBIx::Class SQL string awareness


EffortlessSteven:
## Feature
Recognize SQL in DBI/DBIx::Class code and provide:
- SQL syntax highlighting in strings (->prepare('SELECT...'))
- Basic SQL validation (bracket matching, quote balance)
- Table and column name completion (if schema available)
- SQL injection risk highlighting

## Current State
LSP does not currently recognize SQL contexts within Perl strings.

## Implementation Plan
1. Detect SQL-bearing strings (prepare, do, query methods)
2. Embed SQL tokenizer/highlighter for semantic tokens
3. Provide SQL bracket/quote validation
4. Add code action for SQL formatter integration

## Test Plan
- SQL in prepare() string is highlighted as SQL
- Unbalanced quotes/parens in SQL string are flagged
- SELECT queries show table context awareness

## Perl Ecosystem Context
DBI is the standard Perl database interface. SQL literacy is critical for data-heavy applications. IDE support bridges Perl and SQL DSLs.

EffortlessSteven:
## Builder Specification: DBI/DBIx::Class SQL String Awareness (Issue #2337)

**Status**: Builder-Ready | **Priority**: Enhancement | **Effort**: 1-2 days

### Objectives

Implement semantic token awareness for SQL strings in DBI context:
- New token type: `sql_string`
- Detect SQL-bearing method calls: `prepare()`, `do()`, `query()`, `selectrow_*()`, `fetchall_*()`
- Re-classify first string argument as `sql_string` instead of generic `string`
- Enable clients to apply SQL-specific syntax highlighting (separate colors, styles)

### Deliverables

1. **Code Changes**: 2 edits to semantic tokens crate
2. **Tests**: 10 new integration tests in token classification suite
3. **Verification**: All tests pass + no clippy warnings
4. **Documentation**: Inline code comments explaining SQL method detection

### Implementation Tasks

#### Task 1: Add SQL String Token Type to Legend
**File**: `crates/perl-lsp-semantic-tokens/src/semantic_tokens.rs`
**Lines**: 151-192 (the `legend()` function)

**Change**: Add `"sql_string"` to the `types` vector:

```rust
pub fn legend() -> TokensLegend {
    let types = vec![
        "namespace",
        "class",
        "function",
        "method",
        "variable",
        "parameter",
        "property",
        "keyword",
        "comment",
        "string",
        "number",
        "regexp",
[...]

PR #2772, opened 2026-03-22
fix(lsp): synchronize semantic token legend with capability advertisement (#2103)

### LINKED ISSUE #2103: feat(lsp): add semantic token refinement for Perl operators and special variables
## Problem

textDocument/semanticTokens (LSP 3.16+) is GA but misses Perl-specific tokens:

1. **Sigil operators** - `$`, `@`, `%`, `&` should have distinct semantic tokens
2. **Special variables** - `$_, @_, $!, %ENV` need dedicated tokens
3. **Operator classification** - `=>`, `->`, `..` missing semantic types
4. **Context-aware coloring** - String interpolation vs literal strings

## Current State
- Semantic tokens implemented with 17 token types
- Limited operator coverage
- No special variable differentiation
- No interpolation context

### Option A: Perl-specific token types (P1, ~150 LOC)
- Add `perloOperator`, `perlSigil`, `perlSpecialVar` token types
- Update lexer to emit these tokens
- Extend semantic token legend
- Pro: Rich, language-specific highlighting
- Effort: ~150 lines

### Option B: Interpolation context (P2, ~100 LOC)
- Detect string interpolation context
- Mark interpolated variables differently
- Pro: Shows when variables are interpolated
- Effort: ~100 lines

### Option C: Regex component tokens (P3, ~120 LOC)
- Tokenize regex metacharacters
- Distinguish capture groups from quantifiers
- Effort: ~120 lines

## Acceptance Criteria
- [ ] Sigil operators have distinct tokens
- [ ] Special variables ($_, %ENV) properly colored
- [ ] Operators (=>, ->, ..) have semantic types
- [ ] Tests: 5+ semantic token scenarios
- [ ] Performance: <50ms for full document tokens

EffortlessSteven:
## Summary

Fixes a legend desynchronization bug that caused every semantic token to render with the wrong colour in all LSP clients. Two independent legends existed with different orderings: `legend()` in `perl-lsp-semantic-tokens` (used by the server to compute `tokenType` indices) and the `token_types` vec in `capabilities_for()` in `perl-lsp-protocol` (advertised to clients during `initialize`). Clients use the advertised legend to decode `tokenType` indices, so any mismatch means wrong colours for everything.

Three bugs fixed in one synchronization:
1. **Legend ordering mismatch** — `legend()` types list reordered to exactly match the capability advertisement (20 standard LSP types in the same order)
2. **`sql_string` not advertised** — added in `legend()` by commit `cfc3a514a` for issue #2337 but never added to `capabilities_for()`, causing DBI SQL strings to be decoded as `comment` (index 15 → `comment` in the advertised legend)
3. **`defaultLibrary` modifier bitmask** — was `8` (bit 3) in `legend()` but `512` (bit 9) in the advertised modifier list, causing special variables (`$_`, `%ENV`, `@_`) to render as `static` instead of library-defined

The special variable differentiation (`is_special_variable()`) and SQL string classification were already fully implemented — they just rendered wrong due to the mismatched indices.

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: `semanticTokensProvider.legend.tokenTypes` now includes `sql_string` (was missing), which is additive and backward-compatible
- LSP semantic token indices: changed — all `tokenType` indices in wire responses now match what was advertised during `initialize`. This is a correctness fix; clients were receiving indices that mapped to wrong types before.
- DAP surface: unchanged
- CLI: unchanged

## What changed

- `crates/perl-lsp-semantic-tokens/src/semantic_tokens.rs` — `legend()` types reordered to match capability advertisement (namespace/type/class/interface/enum/enumMember/typeParameter/function/method/property/macro/variable/parameter/keyword/modifier/comment/string/number/regexp/operator/sql_string); modifier list reordered to match; `defaultLibrary` bitmask changed from `8` to `512`
- `crates/perl-lsp-protocol/src/capabilities.rs` — `SemanticTokenType::new("sql_string")` appended after `OPERATOR` in `token_types` vec
- `crates/perl-lsp/tests/lsp_semantic_legend_contract_tests.rs` — **new** contract test that round-trips through the advertised legend; would have caught this bug class at introduction
- `crates/perl-lsp/tests/lsp_semantic_tokens_e2e.rs` — hardcoded expected indices updated from old legend positions to new positions
- `crates/perl-lsp/tests/snapshots/*.json` — 3 snapshot files updated with `sql_string` at position 20
- Two test files with stale hardcoded counts/bitmasks updated

## How to review

Start with the `legend()` diff in `semantic_tokens.rs` — the types list went from 16 items in the wrong order to 21 items in the correct order. Cross-reference with the `token_types` vec in `capabilities_for()` to see they now match exactly.

Then read `lsp_semantic_legend_contract_tests.rs` — the contract test that structurally prevents this class of bug. It captures the advertised legend from `initialize`, then verifies each emitted `tokenType` index resolves to the expected type name.

The `defaultLibrary` bitmask fix is a single line in `collect_semantic_tokens`: `{ 8 }` → `{ 512 }`.

## Evidence

```
test result: ok. 27 passed; 0 failed (perl-lsp-semantic-tokens lib)
test result: ok. 101 passed; 0 failed (comprehensive_unit_tests)
test result: ok. 85 passed; 0 failed (token_classification_tests)
test result: ok. 5 passed; 0 failed (lsp_capabilities_snapshot)
test result: ok. 1 passed; 0 failed (lsp_semantic_tokens_e2e)
test result: ok. 1 passed; 0 failed (lsp_semantic_legend_contract_tests)
cargo fmt --all -- --check: clean
cargo clippy -p perl-lsp-semantic-tokens -p perl-lsp-protocol --tests: no errors
```

Pre-existing failures (not introduced by this PR, confirmed failing on unmodified master):
- `text_document_sync_is_incremental` in `perl-lsp-protocol` — test expects INCREMENTAL(2) but server deliberately uses FULL(1)
- `diagnostic_data_for_parse_error` in `perl-lsp --lib` — unrelated to semantic tokens

## Risk & rollback

**Blast radius**: semantic token wire responses. All `tokenType` indices in responses change to now-correct values. Clients that had adapted to the wrong colours will see correct colours. No data loss, no crash paths changed.

**Failure modes**: if a client cached the old wrong indices, it may briefly show wrong colours until it re-requests. This resolves on next `initialize`.

**Rollback**: revert commit `beed874aa`. The bug was already present in all prior releases.

## Follow-ups

- `text_document_sync_is_incremental` test has the wrong expected value — should assert `FULL(1)`, not `INCREMENTAL(2)`. Recommend follow-up scout to file an issue.
- The dead code path (`handle_semantic_tokens_full`/`handle_semantic_tokens_range` using `SemanticTokensProvider` from `features/semantic_tokens_provider.rs`) still exists with its own inconsistent legend. Not fixed here — it's dead code and out of scope.

Closes #2103

### DIFF
--- a/crates/perl-lsp-protocol/tests/capability_advertisement_tests.rs
+++ b/crates/perl-lsp-protocol/tests/capability_advertisement_tests.rs
@@ -304,12 +304,7 @@ fn semantic_tokens_legend_has_token_modifiers() -> Result<(), Box<dyn std::error
         .ok_or("missing legend.tokenModifiers")?;
     // 10 standard LSP modifiers. This count assertion catches modifier legend
     // desynchronization — if a modifier is added internally but not advertised, this fails.
-    assert_eq!(
-        modifiers.len(),
-        10,
-        "expected 10 token modifiers; got {:?}",
-        modifiers
-    );
+    assert_eq!(modifiers.len(), 10, "expected 10 token modifiers; got {:?}", modifiers);
     let mod_strs: Vec<&str> = modifiers.iter().filter_map(|t| t.as_str()).collect();
     assert!(mod_strs.contains(&"declaration"), "should include 'declaration' modifier");
     assert!(mod_strs.contains(&"definition"), "should include 'definition' modifier");
@@ -323,7 +318,9 @@ fn semantic_tokens_legend_has_token_modifiers() -> Result<(), Box<dyn std::error
     );
     // Verify defaultLibrary is at bit position 9 (index 9 in the modifiers array),
     // which is the bitmask value 512 used by collect_semantic_tokens.
-    let default_library_idx = mod_strs.iter().position(|&s| s == "defaultLibrary")
+    let default_library_idx = mod_strs
+        .iter()
+        .position(|&s| s == "defaultLibrary")
         .expect("defaultLibrary must be in advertised modifiers");
     assert_eq!(
         default_library_idx, 9,


EffortlessSteven:
**Deep review complete. `reviewed-deep` label set. Ready for merge.**

---

## What was verified

**Legend ordering (21 types):** Cross-checked every position between `legend()` in `semantic_tokens.rs` and the `capabilities_for()` advertisement in `capabilities.rs`. All 21 match exactly. Verified against `lsp-types 0.97` string constants directly.

**Bitmask fix:** `defaultLibrary` is at bit 9 in the advertised modifier list. `1 << 9 = 512`. The change from `8` to `512` is correct. All other hardcoded bitmasks (`1` = declaration, `4` = readonly) are correct for their positions.

**kind_idx fallback:** `unwrap_or(&0)` would silently emit index 0 ("namespace") for unknown names. Verified every name used in `kind_idx()` calls is in the legend — the fallback is unreachable.

**Contract test is genuine:** Simulated the pre-fix state: old `legend()` against the advertised legend produces wrong types for 7 of 8 tokens in the test source (e.g. `my` emits index 7 which decoded as "function" from the advertised legend). The test genuinely fails before the fix and passes after.

---

## What was pushed to the branch

**Commit `209ff05e9`: Two new contract tests** covering the two bugs that the original contract test didn't reach:

1. `sql_string_index_decodes_correctly_via_advertised_legend` — full LSP wire round-trip: sends `$dbh->prepare("SELECT 1")`, verifies the emitted `tokenType` index decodes to `"sql_string"` from the advertised legend (before the fix it decoded as `"comment"` — index 15 in the old legend mapped to index 15 in the advertised list = `"comment"`).

2. `default_library_modifier_bitmask_decodes_correctly_via_advertised_legend` — full LSP wire round-trip: sends `sub f { return $_; }`, verifies the variable token's modifier bitmask has bit 9 set (`512`), which the advertised modifier legend maps to `"defaultLibrary"` (before the fix bitmask `8` = bit 3 = `"static"`).

Both tests are position-independent (scan all tokens, don't assume specific line/col) and were verified to fail against the pre-fix code.

---

## Out-of-scope finding (follow-up recommended)

`textDocument/semanticTokens/range` routes to `handle_semantic_tokens_range`, which uses `SemanticTokensProvider::all()` — the old 14-item dead-code legend. The server advertises `range: Some(true)` so this handler is reachable. This existed before this PR and was correctly scoped out. A follow-up builder should wire `handle_semantic_tokens_range` to use `collect_semantic_tokens` the same way `handle_semantic_tokens` does.

---

Regression risk: **LOW.** All token type emissions use `kind_idx()` name-to-index lookup — immune to future reorderings. Only modifier bitmasks are hardcoded and all are now correct.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-2772",
  "current_version": {
    "sha": "9bf956713b44628d4d703885bf3107017e7fe473",
    "updated_at": "2026-03-22T06:22:42Z"
  },
  "label_bindings": [
    {
      "label": "reviewed-deep",
      "bound_at_version": "9bf956713b44628d4d703885bf3107017e7fe473",
      "bound_at_timestamp": "2026-03-22T06:23:04Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Deep Review Wrapup

**Summary:** Reviewer-deep completed deep correctness review of PR #2772. Approved with 2 new contract tests pushed. `reviewed-deep` label set.

---

### Ending State

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/2772
- Changes made: 1 commit pushed (`209ff05e9`) — 2 new contract test functions added, doc comment updated to name all 3 bugs
- Test artifacts touched: `crates/perl-lsp/tests/lsp_semantic_legend_contract_tests.rs`
- Status: approved — `reviewed-deep` label set, label receipt written
- Next step: ready for `/pr-ready` and ops merge

---

### What was verified

**Bug 1 (ordering mismatch):** Cross-checked all 21 type positions between `legend()` and `capabilities_for()`. 21-for-21 match. Verified against `lsp-types 0.97` constant string values directly from source.

**Bug 2 (sql_string not advertised):** The original contract test didn't cover this. Added `sql_string_index_decodes_correctly_via_advertised_legend` — full LSP wire round-trip with DBI `prepare()`. Simulated pre-fix: index 15 decoded as "comment" (index 15 in old advertised list). Test fails before fix, passes after.

**Bug 3 (defaultLibrary bitmask):** The original contract test didn't cover this. Added `default_library_modifier_bitmask_decodes_correctly_via_advertised_legend` — full LSP wire round-trip with `$_` in a sub. Verified `1 << 9 = 512` is correct. Test fails before fix (bitmask 8 decoded as "static"), passes after.

**kind_idx fallback:** `unwrap_or(&0)` would silently emit `namespace` for unknown type names. Verified every name passed to `kind_idx()` is in the legend — the fallback is unreachable dead code.

**Contract test genuineness:** Simulated old legend against advertised: 7 of 8 tokens decoded to wrong types (e.g., `my` → "function", `$x` → "enum"). The test genuinely fails before the fix.

All tests pass: 27 (lib) + 101 (comprehensive) + 85 (token_classification) + 3 (contract) + 1 (e2e) = 217 tests clean.

---

### Out-of-scope finding

`textDocument/semanticTokens/range` routes to `handle_semantic_tokens_range`, which uses `SemanticTokensProvider::all()` — the old dead-code 14-item legend. The server advertises `range: Some(true)` so this handler is client-reachable. This existed before this PR and was correctly scoped out. Recommend a follow-up builder to wire `handle_semantic_tokens_range` to `collect_semantic_tokens` the same way `handle_semantic_tokens` does.

---

### Retrospective

**Harder than expected:** The contract tests I wrote were initially position-dependent (hardcoded line/col for `$_` and the SQL string token). Both failed because the parser produces Variable nodes only inside sub bodies for `$_` (not in bare top-level `print $_;`), and the SQL string position differed from my mental model. The fix was to make the tests position-independent — scan all tokens for the expected type/modifier pattern. This is actually the better design anyway.

**What surprised me:** The `kind_idx()` fallback to index 0 = "namespace" is entirely silent. If someone adds a new token type to the AST walker and misspells the name, every token of that kind silently becomes a namespace token. The fallback should panic in debug builds. Not in scope for this PR but worth noting.

**What would have helped:** Knowing upfront that `sub f { return $_; }` is required rather than bare `print $_;` for `$_` to appear as a Variable AST node. The existing unit test had a comment about this ("Wrap in a sub so the parser creates a Variable node"), but I didn't read it before writing the new test.

---

- **Logical next:** File a follow-up scout issue for `textDocument/semanticTokens/range` using the dead-code `SemanticTokensProvider` with inconsistent legend. The handler is at `crates/perl-lsp/src/runtime/language/semantic_tokens.rs:handle_semantic_tokens_range`. The fix is to replace `SemanticTokensProvider::new(doc.text.clone()).extract(ast)` + `encode_semantic_tokens` with `crate::semantic_tokens::collect_semantic_tokens(ast, &doc.text, &|off| self.offset_to_pos16(doc, off))` exactly as `handle_semantic_tokens` does on line 34.
- **Related issues:** #2103 (closed by this PR)
- **Gotcha:** `$_` does NOT appear as a Variable AST node in top-level code without a function body. Always wrap special-variable tests in `sub f { ... }`.
- **Gotcha:** `legend()` is called fresh on every `collect_semantic_tokens` invocation — it allocates a new `FxHashMap` each time. Benign for current usage (once per request), but worth knowing if it ever appears in a hot loop.
- **Trap:** The `SemanticTokensProvider` in `crates/perl-lsp/src/features/semantic_tokens_provider.rs` has its own `SemanticTokenType::all()` with the OLD 14-item ordering — don't confuse it with the corrected `legend()` in `perl-lsp-semantic-tokens`.
- Confidence for next: high

Issue #2749, opened 2026-03-22
Security audit findings (2026-03-21): 6 findings across 4 severity levels


EffortlessSteven:
## Security Audit — 2026-03-21

Standing security audit of the codebase. 6 findings across 4 severity levels.

### Overall Assessment: GOOD

The codebase has strong security fundamentals:
- Zero `unwrap()`/`expect()` in production code (all in test modules)
- Zero `dbg!()`, `todo!()`, `unimplemented!()`, `panic!()` in production code
- `process::exit()` properly confined to `bin/` and `lifecycle.rs`
- Comprehensive `perl-path-security` + `perl-dap-security` crates for path traversal prevention
- `perl-dap/safe_eval.rs` blocks dangerous Perl operations (system, exec, unlink, etc.)
- Argument injection prevented with `--` separator in subprocess spawning
- Supply chain policy (`deny.toml`) is well-configured: unknown registries/git denied, license allowlist, advisory tracking
- `Cargo.lock` committed
- Symlink traversal explicitly rejected in refactoring module

---

### Finding 1: `unsafe impl Send + Sync for LspServer` (Severity: LOW)

**File**: `crates/perl-lsp/src/runtime/mod.rs:236-238`

```rust
unsafe impl Send for LspServer {}
unsafe impl Sync for LspServer {}
```

**Assessment**: Has a detailed SAFETY comment explaining the reasoning (raw `*const Node` pointers in `ParentMap` behind `Arc<Mutex<...>>`). The justification is sound. Has `#[allow(unsafe_code)]` annotation.

**Risk**: Low. This is a well-documented, legitimate use. However, any future changes to `LspServer` field access patterns should be reviewed against this safety invariant.

**Action**: No code change needed. Add to the review checklist for changes touching `LspServer` fields.

---

### Finding 2: `unsafe { env::set_var(...) }` in CI hygiene tool (Severity: LOW)

**File**: `crates/perl-ci-hygiene/src/main.rs:535, 554`

Two `unsafe` blocks for `env::set_var()` (unsafe in Rust 2024 edition when other threads may be reading environment).

**Assessment**: Has SAFETY comments ("This is a single-threaded CLI tool; no other threads are reading env vars."). The tool is indeed a CLI binary (`main.rs`), not a library. Risk is minimal.

**Action**: No code change needed.

---

### Finding 3: `unsafe` blocks in tree-sitter FFI bindings (Severity: LOW)

[...]

PR #2799, opened 2026-03-22
fix(security): fix macOS sandbox path injection and profile file-path bug (#2749)

EffortlessSteven:
## Summary

Fixes two bugs in `crates/perl-lsp/src/security/sandbox.rs` discovered in the 2026-03-21 security audit (issue #2749):

**Bug A — path injection (line 239):** Paths from `SandboxConfig::allowed_paths` were interpolated directly into the macOS TinyScheme sandbox profile DSL without escaping. A path containing `"` would terminate the string literal early, and a path containing `\` would be misinterpreted as an escape sequence. Either could inject arbitrary sandbox directives. Fix: add `sandbox_escape_path` that escapes `\` first then `"`, and call it in `generate_macos_sandbox_profile`.

**Bug B — profile string passed as file path (line 218):** `sandbox-exec -f` expects a **file path**, not inline profile content. The profile string `"(version 1)\n..."` was passed directly as the `-f` argument, so `sandbox-exec` tried to open a non-existent file with that name as its path. macOS sandboxing was silently non-functional as a result. Fix: write the profile to `self.temp_dir/sandbox.sb` (temp dir is always `Some` when this code is reached — `execute()` returns early via `execute_unsandboxed` when sandboxing is disabled) and pass that path to `-f`.

Also resolves three pre-existing compile errors that were blocking the `security` module from being wired into `lib.rs`: `uuid` was in `[dev-dependencies]` but used in production code; `log::warn!` was used instead of `tracing::warn!` (the crate standard); and `with_context` was called on `Result<_, String>` (not an `anyhow` error type).

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: unchanged
- DAP surface: unchanged
- CLI: unchanged
- `pub mod security` is now exposed in `perl-lsp` lib (was unreachable before)

## What changed

- `sandbox_escape_path(path)` — new associated fn, `#[cfg(any(target_os = "macos", test))]` so it compiles cross-platform for testing but is dead-code-free in production Linux builds
- `generate_macos_sandbox_profile` — now uses `sandbox_escape_path`, same cfg guard
- `apply_macos_sandbox` — writes profile to `self.temp_dir/sandbox.sb`, passes file path to `-f` instead of inline content
- `Cargo.toml` — `uuid` moved from `[dev-dependencies]` to `[dependencies]`
- `lib.rs` — `pub mod security` wired in
- `security/mod.rs` — `log::warn!` → `tracing::warn!`
- `security/sandbox.rs` — `.with_context()` on `String` errors → `.map_err(|e| anyhow!(...))`, `log::warn!` → `tracing::warn!`

## How to review

Start at `apply_macos_sandbox` (line ~208) to see Bug B fix, then `sandbox_escape_path` + `generate_macos_sandbox_profile` for Bug A. Tests are in the `#[cfg(test)]` block — four new tests plus the ten existing ones all pass. The `new_for_test` constructor allows the profile generation to be tested without a real temp directory.

## Evidence

```
test security::sandbox::tests::test_sandbox_config_default ... ok
test security::sandbox::tests::test_sandbox_escape_path_backslash ... ok
test security::sandbox::tests::test_generate_macos_sandbox_profile_escapes_paths ... ok
test security::sandbox::tests::test_sandbox_result ... ok
test security::sandbox::tests::test_sandbox_escape_path_clean ... ok
test security::sandbox::tests::test_sandbox_escape_path_double_quote ... ok
test security::sandbox::tests::test_sandbox_creation ... ok
test security::sandbox::tests::test_unsandboxed_execution ... ok
test security::sandbox::tests::test_perl_script_execution ... ok
test security::sandbox::tests::test_safe_executor ... ok
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 271 filtered out

cargo clippy -p perl-lsp --lib -- -D warnings: Finished (clean)
cargo fmt -p perl-lsp -- --check: PASS
Full lib suite: 271 passed; 0 failed
```

Two pre-existing integration test failures in `execute_command_mutation_hardening_public_api_tests` are confirmed unrelated to this change — they fail on the base branch before any edits.

## Risk & rollback

- Blast radius: macOS sandbox path only (`#[cfg(target_os = "macos")]`). Linux and Windows are unaffected.
- The `temp_dir` invariant (always `Some` when `apply_macos_sandbox` is reached) is enforced by `execute()`'s early-return on `!config.enabled`. The `ok_or_else` is a defence-in-depth guard.
- Rollback: revert this commit. The only observable behaviour change is that macOS sandboxing now actually functions (it was silently no-op before) and paths with quotes/backslashes no longer corrupt the profile.

## Follow-ups

- The two failing `execute_command_mutation_hardening_public_api_tests` are pre-existing and unrelated — recommend a separate scout issue.
- `SandboxConfig::allowed_paths` currently accepts any path without validation. If configuration sources expand (e.g. workspace settings), a path-safety check would be prudent — out of scope for this fix.

### DIFF
--- a/crates/perl-lsp/src/security/sandbox.rs
+++ b/crates/perl-lsp/src/security/sandbox.rs
@@ -236,7 +236,15 @@ impl Sandbox {
 
     /// Escape a path string for safe interpolation into a macOS sandbox profile DSL string literal.
     /// The profile language is TinyScheme-based; only `\` and `"` are special inside quoted strings.
-    /// Backslash must be escaped first to avoid double-escaping.
+    /// Parentheses are NOT special inside a quoted string (only at the DSL level), so they are
+    /// left unescaped. Backslash must be escaped first to avoid double-escaping the newly
+    /// introduced backslashes in the second pass.
+    ///
+    /// Limitation: paths containing literal newlines or null bytes are not escaped here.
+    /// Such paths are pathological on macOS (HFS+ permits them but they are essentially never
+    /// used) and are out of scope for this fix. If configuration sources ever accept arbitrary
+    /// user-supplied paths, add a sanitisation step that rejects control characters before
+    /// calling this function.
     #[cfg(any(target_os = "macos", test))]
     fn sandbox_escape_path(path: &std::path::Path) -> String {
         path.display().to_string().replace('\\', "\\\\").replace('"', "\\\"")
@@ -440,6 +448,17 @@ mod tests {
         // Backslash must be doubled (Windows-style path or escape chars)
         let path = std::path::Path::new("/home/user/my\\path");
         assert_eq!(Sandbox::sandbox_escape_path(path), "/home/user/my\\\\path");
+    }
+
+    #[test]
+    fn test_sandbox_escape_path_backslash_then_quote() {
+        // A path containing both \ and " verifies the escape ordering:
+        // backslash is doubled first (\\ -> \\\\), then the quote is escaped (" -> \").
+        // If the order were reversed, the backslash introduced by quote-escaping would
+        // itself get doubled, producing \\\" instead of the correct \".
+        let path = std::path::Path::new("/home/user/my\\\"path");
+        // Expected: /home/user/my\\"path (\ -> \\, then " -> \")
+        assert_eq!(Sandbox::sandbox_escape_path(path), "/home/user/my\\\\\\\"path");
     }
 
     #[test]


EffortlessSteven:
## Merge Complete

**PR #2799:** fix(security): fix macOS sandbox path injection and profile file-path bug (#2749)
**CI Status:** PASSED
**Merged at:** 2026-03-22T11:56:52Z

Ready for next batch.

Issue #2074, opened 2026-03-19
Add LRU cache for goto-definition and hover symbol lookups


EffortlessSteven:
## Performance Opportunity: Symbol Lookup Caching

**File**: crates/perl-lsp/src/runtime/language/hover.rs, navigation.rs
**Hot Path**: textDocument/hover, textDocument/definition requests
**Issue**: SemanticAnalyzer::find_definition() and workspace index lookups are called without memoization

### Current Pattern
- Line 48-49 (hover.rs): SemanticAnalyzer created fresh per hover request
- Navigation and goto-def perform symbol resolution each time without caching intermediate results
- Workspace index get_package_members() re-scans HashMap on every completion/hover

### Impact
Repeated requests at same position re-do AST traversal and symbol table searches. Navigating through code involves many hover/goto-def calls from editor at same position during editing sessions.

### Solution Approach
- Add LRU cache (max 100 entries) keyed by (uri, offset)
- Cache hover/definition lookup results for 5-10 seconds
- Invalidate on document change
- Consider arc-swap for thread-safe cache updates

### Acceptance Criteria
- Hover response time < 20ms on repeated requests (vs. 100ms+ uncached)
- Memory overhead < 5MB for typical workspace
- Proper cache invalidation on textDocument/didChange

EffortlessSteven:
## Plan Review

**File references:** Corrected. The scout listed the paths without the `runtime/language/` intermediate directories. Both files exist at their correct locations:
- `crates/perl-lsp/src/runtime/language/hover.rs`
- `crates/perl-lsp/src/runtime/language/navigation.rs`

The scout's line 48-49 reference for hover.rs is wrong. The actual hot-path call is at **line 96** (`SemanticAnalyzer::analyze_with_source`) and **line 181** (`TypeInferenceEngine::new().infer(ast)`). These are the two expensive traversals per hover request.

**Root cause:** Partially confirmed, with corrections.

The real performance problem is two full O(n) AST traversals per hover request (one for `SemanticAnalyzer::analyze_with_source`, one for `TypeInferenceEngine::new().infer`), neither of which is memoized between requests to the same document. This is real and worth fixing.

The scout's claim that `get_package_members()` "re-scans HashMap on every completion/hover" is **false** -- there is no such function in the current codebase. The workspace index uses `find_def()` and `find_definition()` which are O(1) HashMap lookups. Remove this from the issue narrative.

[...]

PR #2806, opened 2026-03-22
perf(lsp): add SemanticAnalyzer LRU cache for hover/definition lookups (#2074)

EffortlessSteven:
## Summary

Adds an `Arc<parking_lot::Mutex<HashMap<(String, u64), Arc<SemanticAnalyzer>>>>` cache to `LspServer` that avoids re-running the full O(n) AST traversal (`SemanticAnalyzer::analyze_with_source`) on repeated hover and goto-definition requests to the same document version.

The cache key is `(normalized_uri, content_hash_of_text)` — not `(uri, offset)` as the original scout proposed. One cache entry covers all hover positions in a document version. Invalidated eagerly on `textDocument/didChange` and `textDocument/didClose`. No TTL needed: content-hash keying makes entries automatically stale when the file changes. Bounded to 50 entries (HashMap-based, no new crate dependency — consistent with `pod_cache` and `telemetry_cooldowns` patterns).

Also fixes 3 bare `eprintln!` debug calls in `navigation.rs` (lines 483, 487, 509) that fired unconditionally on every successful workspace goto-definition lookup. Converted to `tracing::debug!`.

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: unchanged (behavior identical, same responses)
- DAP surface: unchanged
- CLI: unchanged
- `LspServer` struct: additive (`semantic_analyzer_cache` field, `pub(crate)`)
- `extract_symbol_hover` signature: additive (`uri: &str` parameter added, private method)

## What changed

| File | Change |
|------|--------|
| `runtime/mod.rs` | Added `semantic_analyzer_cache` field to `LspServer` struct |
| `runtime/constructors.rs` | Initialized field in all 3 constructors (`new_with_feature_profile`, `with_io_and_feature_profile`, `with_output_and_feature_profile`) |
| `runtime/document_access.rs` | Added `get_or_build_analyzer(&self, uri, text, ast) -> Arc<SemanticAnalyzer>` helper |
| `runtime/language/hover.rs` | Updated `extract_symbol_hover` to accept `uri` and call `get_or_build_analyzer` instead of the direct `analyze_with_source` call |
| `runtime/language/navigation.rs` | Converted 3 `eprintln!` → `tracing::debug!` |
| `runtime/text_sync.rs` | Cache invalidation on `didChange` and `didClose`; 4 new unit tests |

Lock ordering: `documents` is always acquired before `semantic_analyzer_cache` when both are held simultaneously (in `handle_hover` → `extract_symbol_hover` and in `handle_did_change_with_cancellation`). In `handle_did_close` the two locks are taken sequentially (never nested). No deadlock risk.

## How to review

Start at `document_access.rs:get_or_build_analyzer` — that is the core of the change. Then verify the two invalidation sites in `text_sync.rs` (search for "Invalidate the SemanticAnalyzer"). Check `hover.rs` to confirm only the one call site was changed. The `navigation.rs` diff is mechanical (`eprintln!` → `tracing::debug!`).

The 4 new tests in `text_sync.rs` (bottom of `mod tests`) exercise: cache hit on same version, invalidation on `didChange`, invalidation on `didClose`, and version separation.

## Evidence

```
test runtime::text_sync::tests::test_semantic_analyzer_cache_invalidated_on_did_change ... ok
test runtime::text_sync::tests::test_semantic_analyzer_cache_invalidated_on_did_close ... ok
test runtime::text_sync::tests::test_semantic_analyzer_cache_reuses_entry_on_same_version ... ok
test runtime::text_sync::tests::test_semantic_analyzer_cache_separates_document_versions ... ok

test result: ok. 261 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

cargo clippy -p perl-lsp --tests -- -D warnings  →  Finished (no warnings)
cargo fmt -p perl-lsp -- --check                 →  PASS (auto-fixed style diffs)
```

Pre-existing ci-gate failure: `perl-lsp-protocol` test has a `clippy::expect_used` violation in `capability_advertisement_tests.rs:324` — fails identically on `master`, not introduced here.

Pre-existing test failures: `execute_command_mutation_hardening_public_api_tests` (2 tests) fail identically on `master`.

## Risk & rollback

- **Blast radius**: `perl-lsp` only, hover path only
- **Failure modes**: Cache returns wrong result — mitigated by content-hash keying (same hash = same content = same analysis). Cache grows unbounded — mitigated by 50-entry clear-on-full bound. Lock deadlock — mitigated by consistent lock ordering (`documents` → `semantic_analyzer_cache`).
- **Rollback**: Revert `document_access.rs:get_or_build_analyzer` call in `extract_symbol_hover` to restore `SemanticAnalyzer::analyze_with_source(ast, text)` directly. All other changes are additive/cleanup and safe to keep.

## Follow-ups

- `TypeInferenceEngine::new().infer(ast)` at hover.rs line 181 is a second O(n) traversal per hover request — not cached by this PR. A follow-up could extend the same pattern.
- The 50-entry clear-on-full eviction is simple but discards hot entries. A proper LRU eviction (using `moka` from `perl-lsp-performance` as a direct dependency, or a small hand-rolled LRU) could be added later if profiling shows benefit.

### DIFF
--- a/crates/perl-lsp/src/runtime/constructors.rs
+++ b/crates/perl-lsp/src/runtime/constructors.rs
@@ -53,6 +53,7 @@ impl LspServer {
             pod_cache: Arc::new(Mutex::new(HashMap::new())),
             pending_index_task_count: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
             parse_cancel_flags: Arc::new(Mutex::new(HashMap::new())),
+            semantic_analyzer_cache: Arc::new(Mutex::new(HashMap::new())),
             #[cfg(feature = "workspace")]
             indexing_in_progress: Arc::new(AtomicBool::new(false)),
         }
@@ -147,6 +148,7 @@ impl LspServer {
             pod_cache: Arc::new(Mutex::new(HashMap::new())),
             pending_index_task_count: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
             parse_cancel_flags: Arc::new(Mutex::new(HashMap::new())),
+            semantic_analyzer_cache: Arc::new(Mutex::new(HashMap::new())),
             #[cfg(feature = "workspace")]
             indexing_in_progress: Arc::new(AtomicBool::new(false)),
         }
@@ -204,6 +206,7 @@ impl LspServer {
             pod_cache: Arc::new(Mutex::new(HashMap::new())),
             pending_index_task_count: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
             parse_cancel_flags: Arc::new(Mutex::new(HashMap::new())),
+            semantic_analyzer_cache: Arc::new(Mutex::new(HashMap::new())),
             #[cfg(feature = "workspace")]
             indexing_in_progress: Arc::new(AtomicBool::new(false)),
         }
--- a/crates/perl-lsp/src/runtime/document_access.rs
+++ b/crates/perl-lsp/src/runtime/document_access.rs
@@ -182,4 +182,53 @@ impl LspServer {
         let docs = self.documents.lock();
         docs.iter().map(|(uri, doc)| (uri.clone(), doc.text.clone())).collect()
     }
+
+    /// Get or compute a `SemanticAnalyzer` for the given document text and AST.
+    ///
+    /// Cache key: `(normalized_uri, content_hash_of_text)`.
+    /// The entry is valid as long as the content hash matches — no TTL needed.
+    /// Cache is bounded to 50 entries; a simple clear is used when full.
+    ///
+    /// **Lock discipline**: acquires `semantic_analyzer_cache` in two short
+    /// scopes (read, then write). May be called while `documents` is held;
+    /// always acquire `documents` before `semantic_analyzer_cache` to maintain
+    /// a consistent lock ordering and avoid deadlock.
+    pub(crate) fn get_or_build_analyzer(
+        &self,
+        uri: &str,
+        text: &str,
+        ast: &perl_parser::ast::Node,
+    ) -> Arc<crate::semantic::SemanticAnalyzer> {
+        use std::collections::hash_map::DefaultHasher;
+        use std::hash::{Hash, Hasher};
+
+        let mut hasher = DefaultHasher::new();
+        text.hash(&mut hasher);
+        let content_hash = hasher.finish();
+
+        let normalized = self.normalize_uri_key(uri);
+        let key = (normalized, content_hash);
+
+        // Read path: return clone of cached entry if present.
+        {
+            let cache = self.semantic_analyzer_cache.lock();
+            if let Some(cached) = cache.get(&key) {
+                return Arc::clone(cached);
+            }
+        }
+
+        // Cache miss: build the analyzer outside the lock.
+        let analyzer = Arc::new(crate::semantic::SemanticAnalyzer::analyze_with_source(ast, text));
+
+        // Write path: insert, evicting all entries when the cache is full.
+        {
+            let mut cache = self.semantic_analyzer_cache.lock();
+            if cache.len() >= 50 {
+                cache.clear();
+            }
+            cache.insert(key, Arc::clone(&analyzer));
+        }
+
+        analyzer
+    }
 }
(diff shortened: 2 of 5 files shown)


EffortlessSteven:
## Deep Correctness Review

**Result: Approved. No fixes required.**

### What I verified

**Cache invalidation correctness**

Invalidation fires in both invalidation-relevant text-sync events:
- `didChange` (text_sync.rs:368-372): cache cleared by URI before the document text is updated, inside the `documents` lock. Correct sequencing — no hover can observe the stale entry because it cannot hold `documents` while the cache is being cleared.
- `didClose` (text_sync.rs:752-756): cache cleared by URI. Sequential with `documents` (not nested), so no lock-ordering concern.
- `didOpen`: intentionally omitted. Content-hash keying makes it unnecessary — a new document version produces a new hash and gets a cache miss. Confirmed correct by design.

**Lock ordering**

Two paths nest both locks:
- hover (hover.rs:57-87): `documents` acquired first, `semantic_analyzer_cache` acquired inside (via `get_or_build_analyzer`)
- didChange (text_sync.rs:365-372): `documents` acquired first, `semantic_analyzer_cache` acquired inside

Both paths use the same order (`documents` → `semantic_analyzer_cache`). No deadlock.

`didClose` acquires them sequentially (not nested): `semantic_analyzer_cache` released at line 756 before `documents` acquired at line 765. The PR description's claim about lock ordering is accurate.

**Thread safety**

`parking_lot::Mutex` provides correct synchronization. `Arc<SemanticAnalyzer>` in the cache value is safe for shared ownership across threads (SemanticAnalyzer contains only owned-heap types: String, HashMap, Vec — all Send+Sync). The check-then-build-then-insert pattern can duplicate work under concurrent hovers for the same document, but the result is always correct (both threads build equivalent analyzers; last writer wins). Duplicated work degrades to baseline (no worse than pre-PR).

**Eviction policy**

Clear-all at 50 entries (not LRU). In practice, typical LSP sessions have 1-5 open files producing 1-5 cache entries. The 50-entry limit acts as a safety ceiling, not a constant-pressure eviction trigger. The title says "LRU" but the implementation is simpler. This is documented in the PR description as a known limitation with a suggested follow-up. It does not affect correctness.

**Content-hash correctness**

`DefaultHasher` (SipHash 1-3) is randomized per process. This is fine — the cache is per-server-instance and does not persist across restarts. Within a session, same text produces same hash. Hash collisions (64-bit) are negligible for practical file sizes. Collision would return a stale analyzer, but the (uri, hash) key pair makes cross-URI collisions impossible, and same-URI collisions require the old and new content to have identical hashes, which is cryptographically implausible.

**Performance characteristics**

Cache miss path: identical lock-hold time to pre-PR (O(n) AST traversal while `documents` is held). Cache hit path: strictly better (HashMap lookup + Arc::clone only). No performance regression is possible.

**Edge cases**

- Empty file: hover returns `HoverExtracted::None` before reaching `get_or_build_analyzer` (AST is None when parse fails, or returns None from `find_definition`). Cache not populated. Correct.
- Very large files: hash is O(n) but so is every other per-hover operation. Cache benefit largest for large files.
- Rapid document changes: each `didChange` clears the stale cache entry. A race between a concurrent hover and a didChange results in at most one orphaned cache entry (for the old content hash). The orphaned entry is never served (document lookup will fail or return different text). It evicts harmlessly within 50 entries.
- No-AST document: guard at hover.rs:59 (`if let Some(ast)`) prevents `get_or_build_analyzer` from being called without an AST. Cache key requires URI + text, both provided by the caller.

**Tests**

All 4 new tests pass:
- `test_semantic_analyzer_cache_reuses_entry_on_same_version` — confirms single entry per document version
- `test_semantic_analyzer_cache_invalidated_on_did_change` — confirms didChange eviction
- `test_semantic_analyzer_cache_invalidated_on_did_close` — confirms didClose eviction
- `test_semantic_analyzer_cache_separates_document_versions` — confirms version separation

Pre-existing failures (`execute_command_mutation_hardening_public_api_tests` 2 tests, `go_to_definition_on_extends_parent_navigates_to_parent_file`) confirmed pre-existing on master — not introduced by this PR.

**No fixes pushed.** The implementation is correct, the tests are comprehensive, and the documented limitations (clear-all eviction, DefaultHasher) are appropriate for the use case.

Issue #2811, opened 2026-03-22
control-plane: ADR-0032 enforcement promises never delivered — TaskCompleted metrics gate, TeammateIdle registration, SubagentStart injection


EffortlessSteven:
## Summary

ADR-0032 (Skill Scoping and Hook Enforcement) promised three mechanical enforcement layers to solve the metrics compliance failure (zero of 30 PRs wrote a metrics entry). As of 2026-03-22, the original promises remain unimplemented.

**This is not a vague audit.** The ADR itself documents the exact divergence (lines 100-112 of `docs/adr/0032-skill-scoping-and-hook-enforcement.md`). The control plane is running a reduced version of the design.

### 1. TaskCompleted Hook — Metrics Gate Never Enforced (CRITICAL)

**Promise (ADR line 50):**
> Blocks completion unless a metrics entry exists in `swarm-metrics.jsonl`

**Actual (ADR line 108):**
> Only checks `cargo fmt` and `CURRENT_STATUS.md`; no metrics gate implemented

**Evidence:**
- Hook file: `/home/steven/code/Rust/perl-lsp/tree-sitter-perl-rs/.claude/hooks/task-completed.sh` (27 lines)
- Lines 6-24: checks `cargo fmt` and `CURRENT_STATUS.md` only
- **Missing**: check for metrics entry in `.ops-perl-lsp/swarm-metrics.jsonl`

**Consequence:**
The original motivation for ADR-0032 was never resolved. The problem statement (ADR line 15): "Zero of 30 merged PRs contained a metrics entry. Prompt instructions are advisory; agents skip them under time or context pressure." The hook was supposed to enforce this structurally. It doesn't.

### 2. TeammateIdle Hook — Registered in Spec, Missing in Reality

**Promise (ADR line 52):**
> Blocks idle state if the teammate has in-progress tasks not yet completed

**Actual (ADR line 109):**
> Hook script exists in `docs/handoff/swarm-pack/hooks/` but is **not registered** in live `settings.json`

**Evidence:**
- Hook file exists: `/home/steven/code/Rust/perl-lsp/tree-sitter-perl-rs/docs/handoff/swarm-pack/hooks/teammate-idle.sh` (7 lines)
- Hook content: exits 2 (blocks idle) with guidance to check task list
- Settings check: `.claude/settings.json` lines 39-109 shows NO TeammateIdle entry

**Consequence:**
[...]

PR #2828, opened 2026-03-23
feat(swarm): ADR-0032 enforcement — TaskCompleted JSONL, TeammateIdle retraction (#2811)

EffortlessSteven:
## Summary

Implements the corrected ADR-0032 enforcement spec (plan-reviewed 2026-03-22). Three hooks, one retraction, one new observability script.

- **TaskCompleted hook**: adds passive JSONL metrics write (`event=task_completed` with `session_id` and `cwd`); advisory-only (exit 0 always) since SubagentStop lifecycle ordering makes a blocking gate impossible. Adds `.rs` file guard for `cargo fmt` — avoids 2.5s overhead for doc/shell-only agents.
- **SubagentStop hook**: prefers `cwd` field (platform-provided) over `worktree_path` (not in platform payload) so worktree agent entries are correlatable with actual work.
- **SubagentStart hook**: expands 85-char placeholder to a 4-pitfall reminder (no unwrap, no git stash, CARGO_TARGET_DIR, no git add -A).
- **TeammateIdle**: formally retracted in ADR-0032 divergence section. Unconditional exit-2 creates infinite busy-loop when task queue is empty; 6 sessions of 50-100 agents never missed it.
- **New**: `scripts/swarm-summary.sh` observability script for readable JSONL summary. `just swarm-summary` target added.
- **CI gate**: `just hook-tests` added to merge-gate pipeline (existed as target but wasn't enforced).
- **Tests**: 6 new test cases in `test-hooks.sh` covering metrics write, session_id capture, and cwd-as-worktree_path. All 22 tests pass.

## Why advisory, not blocking (key design decision)

`TaskCompleted` fires when an agent marks a task done. `SubagentStop` fires when the agent shuts down — always AFTER TaskCompleted. A gate checking for a SubagentStop-written entry would always fail. The session_id and agent_name fields in the live metrics file are empty for this reason. The passive write approach solves data capture without the race condition.

## Files changed

| File | Change |
|------|--------|
| `.claude/hooks/task-completed.sh` | Read stdin at top; add .rs guard for cargo fmt; add passive metrics write |
| `.claude/hooks/subagent-stop.sh` | Prefer `cwd` over missing `worktree_path` field in stdin payload |
| `.claude/settings.json` | Expand SubagentStart echo with 4 known pitfalls |
| `docs/adr/0032-skill-scoping-and-hook-enforcement.md` | Add TeammateIdle retraction paragraph; update divergence table |
| `docs/handoff/swarm-pack/hooks/teammate-idle.sh` | Add retraction comment |
| `.ci/scripts/test-hooks.sh` | Add 6 new test cases (metrics write + cwd capture) |
| `scripts/swarm-summary.sh` | New observability script |
| `justfile` | Add swarm-summary target; add hook-tests to ci-gate |
| `crates/perl-lsp/tests/wired_crates_integration_test.rs` | Fix pre-existing cargo fmt issue |

## Test plan

- [x] `bash .ci/scripts/test-hooks.sh` — 22/22 pass
- [x] `bash .ci/scripts/check-hook-registry.sh` — 3 scripts verified
- [x] `bash .ci/scripts/check-hook-executable.sh` — all executable
- [x] Manual verify: `echo '{"session_id":"x","cwd":"/test"}' | OPS_DIR=/tmp/t bash .claude/hooks/task-completed.sh` writes `task_completed` entry
- [x] Manual verify: `echo '{"subagent_type":"builder","cwd":"/worktrees/agent-abc"}' | OPS_DIR=/tmp/t bash .claude/hooks/subagent-stop.sh` populates `worktree_path` from `cwd`
- [x] `cargo fmt --all -- --check` passes

### DIFF
--- a/.claude/hooks/task-completed.sh
+++ b/.claude/hooks/task-completed.sh
@@ -9,8 +9,11 @@ INPUT="$(cat 2>/dev/null || echo '{}')"
 REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || echo ".")"
 
 # Quick sanity check: is cargo fmt clean?
-# Guard: only run cargo fmt if .rs files exist in the working tree (avoids 2.5s overhead for doc/shell-only agents)
-if git ls-files -- '*.rs' 2>/dev/null | grep -q .; then
+# Guard: only run cargo fmt if the agent staged or recently committed .rs files.
+# Note: git ls-files checks ALL tracked files (always matches in this repo), not modified files.
+# We use git diff to detect actual .rs work by this agent -- same pattern as the test-file check below.
+if git diff --cached --name-only 2>/dev/null | grep -q '\.rs$' || \
+   git diff --name-only HEAD~1 2>/dev/null | grep -q '\.rs$'; then
   if ! cargo fmt --all -- --check 2>/dev/null; then
     echo "Task completion blocked: cargo fmt check failed. Run 'cargo fmt --all' before marking complete."
     exit 2


EffortlessSteven:
## Deep Review — reviewed-deep

**Verified:**
- JSONL write is append-only (`>>`) and fully guarded: `mkdir -p || true`, `jq ... || true`, `2>/dev/null`. Safe against missing `jq`, missing `OPS_DIR`, and filesystem errors.
- `subagent-stop.sh` `cwd` preference is correct: `cwd` is platform-provided, `worktree_path` is not in the live platform payload. The `set -euo pipefail` in that script with `jq // empty` forms is safe.
- TeammateIdle retraction reasoning is sound. The hook unconditionally exits 2, creating an infinite busy-loop when the task queue is genuinely empty. The retraction comment in `teammate-idle.sh` and the ADR-0032 divergence section are clear.
- stdin capture at the top of `task-completed.sh` is correct — stdin can only be consumed once, and the previous code read it in multiple subshells.
- The 6 new test cases in `test-hooks.sh` follow the established `assert_file_exists` + bare `if [[ -f ]]` pattern. Not a regression — the existing tests use the same pattern.
- `swarm-summary.sh` is safe: exits 1 (not 2) if no metrics file, reads via `jq -r` with graceful no-op on empty file.
- `just hook-tests` added to `ci-gate` — the existing target was untested in the merge gate.
- SubagentStart pitfall echo is a meaningful improvement over the 85-char placeholder.

**Fixed (commit 97683d0ea):**
- `git ls-files -- '*.rs'` checks all 1756 tracked `.rs` files in this repo, so the guard `| grep -q .` is always true — `cargo fmt` is never skipped for doc/shell-only agents despite the comment claiming it would be. Fixed to use `git diff --cached --name-only | grep -q '\.rs$'` (staged files) and `git diff --name-only HEAD~1 | grep -q '\.rs$'` (recent commit), matching the same pattern already used in the CURRENT_STATUS.md gate below it.

**Design note (documented, no fix needed):** The ADR correctly notes that `TaskCompleted` fires before `SubagentStop`, making a blocking gate on SubagentStop-written entries impossible. The passive write approach is the right design decision. The architecture comment in the hook is accurate.

Issue #2834, opened 2026-03-23
parser: scout report — unclosed_paren_identifier bucket (72 files) decomposition needed


EffortlessSteven:
# Scout Report: unclosed_paren_identifier Error Bucket

## Overview

`unclosed_paren_identifier` is an error bucket representing cases where a left parenthesis is not closed, with an identifier expected or found in the context. The system corpus shows **72 affected files**.

## Statistics

| Metric | Value |
|--------|-------|
| System corpus files | 72 |
| Unique modules | ~36 |
| % of total errors | ~2% |
| Rank | 3rd highest unfiled bucket |

## Affected Examples

Core Perl modules impacted:
- `IPC/Msg.pm`, `IPC/Semget.pm`, `IPC/Shmctl.pm`
- `Storable.pm` (complex XS module)
- `Unicode/Collate.pm`, `Unicode/Normalize.pm`
- `Errno.pm`, `Socket.pm` (system interface modules)
- `ExtUtils/Constant.pm`, `ExtUtils/ParseXS.pm`
- `Tie/Hash.pm`, `Tie/Scalar.pm`, `Tie/SubstrHash.pm`
- `autodie.pm`, `autouse.pm` (pragma modules)
- CPAN ecosystem: `Module/ExtractUse.pm`, `Compiler/*.pm` modules

## Likely Root Causes (Speculative)

1. **Bareword builtin arguments**
   - `pack("L!", ...)` or `unpack(...)` where bareword format string confuses parser
   - Bareword function calls: `croak(...)` without explicit parens in some contexts

2. **List context ambiguity**
   - `my @arr = ( ... )` where list content causes paren mismatch
   - Hash initialization with implicit list: `my %h = ( key => val )`

3. **Class::Struct declaration**
   - `struct 'PackageName' => [ field => type, ... ]`
   - Arrow operator in special syntax

4. **Method call in special contexts**
   - Chained calls with bareword method names
   - Indirect object syntax: `new Class ( ... )`

5. **Regex/Quote operators**
   - `qw(...)`, `qr(...)` with complex content
   - List assignment with quote-like operators

6. **XS/C interface patterns**
   - Perlmod patterns for XS glue code (Storable.pm, Errno.pm)

### Phase 1: Sampling
1. Examine 8-12 files
2. Look for first ERROR node
3. Identify what construct contains unclosed paren
4. Check if error is at bareword, identifier, or complex expr

### Phase 2: Pattern Grouping
[...]

PR #2840, opened 2026-03-23
fix(parser): handle unclosed_paren_identifier patterns (#2834)

EffortlessSteven:
## Summary

Fixes three distinct root causes in the `unclosed_paren_identifier` error bucket (issue #2834, ~72 affected corpus files).

### Root Cause 1: `undef EXPR` in expression context

`undef` is lexed as `TokenKind::Undef` (not `Identifier`), so it bypassed the postfix chain's generic builtin-function handling. When `undef $var` appears inside parenthesized expressions (e.g. `if ($x or undef $ret)`), the parser saw the argument as unexpected.

**Fix**: Add explicit `NodeKind::Undef` sigil-peek in `postfix.rs` — when `undef` is followed by a `$/@/%`-starting token in expression context, treat it as `undef(EXPR)`.

### Root Cause 2: `'-' x IDENTIFIER` repetition operator RHS

The `is_operand_start` check in `parse_multiplicative_with` for the `x` operator only accepted sigil-prefixed identifiers as the RHS, rejecting plain function names like `width $title`.

**Fix**: Extend `is_operand_start` in `precedence.rs` to allow plain identifiers that are not binary-operator keywords (`or`, `and`, `not`, etc.).

### Root Cause 3: `print($fh EXPR)` filehandle-in-explicit-parens

`print`, `say`, `printf`, and `send` support `print(FILEHANDLE LIST)` with the filehandle as the first arg without a comma. Both the expression-level postfix chain and the statement-level handler were calling `parse_args()` which treats everything as comma-separated, causing the identifier after the filehandle `$var` to be seen as unexpected.

**Fix**: Add `parse_print_parens_args()` in `calls.rs` that detects `$scalar NON_COMMA_TOKEN` as the filehandle pattern and parses accordingly. Wire it in both `postfix.rs` and `statements.rs`.

## Test plan

- [x] 8 new regression tests added to `unclosed_paren_identifier_tests.rs` (all passing)
- [x] `cargo test -p perl-parser-core` — 5 pre-existing failures in `fix_expected_colon` (unrelated, present on master); all other suites pass
- [x] `cargo fmt --all` — clean
- [x] `cargo clippy -p perl-parser-core --tests` — clean

## Knowledge artifacts

- `undef` is `TokenKind::Undef`, not `Identifier` — postfix sigil-peek must explicitly handle it
- `x` repetition operator RHS check was too restrictive; any non-keyword identifier is a valid operand start
- `print(FILEHANDLE LIST)` is a Perl-canonical form requiring separate parsing from regular `print(ARGS)`; `peek_second()` is the correct lookahead to detect the filehandle pattern

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/statements.rs
+++ b/crates/perl-parser-core/src/engine/parser/statements.rs
@@ -821,7 +821,7 @@ impl<'a> Parser<'a> {
                                 && (Self::is_block_list_func(func_name.as_ref())
                                     || matches!(
                                         func_name.as_ref(),
-                                        "exec" | "system" | "print" | "say" | "printf"
+                                        "exec" | "system" | "print" | "say" | "printf" | "send"
                                     ))
                             {
                                 // block-list and filehandle builtins followed by (...) use
--- a/crates/perl-parser-core/tests/unclosed_paren_identifier_tests.rs
+++ b/crates/perl-parser-core/tests/unclosed_paren_identifier_tests.rs
@@ -791,3 +791,96 @@ fn print_parens_filehandle_var() {
     // print( $fh $msg ) — variable after filehandle
     assert_clean_parse(r#"print( $fh $msg );"#);
 }
+
+// === Additional edge case coverage (#2834 deep review) ===
+
+#[test]
+fn undef_no_arg_in_expr() {
+    // Plain `undef` with no argument in expression context — must not consume next token
+    assert_clean_parse(r#"my $x = $y || undef;"#);
+}
+
+#[test]
+fn undef_no_arg_in_ternary() {
+    // undef as rhs of ternary — no argument
+    assert_clean_parse(r#"my $x = $cond ? 1 : undef;"#);
+}
+
+#[test]
+fn undef_array_arg_in_expr() {
+    // undef @arr in expression context (% sigil also supported)
+    assert_clean_parse(r#"$x or undef @arr;"#);
+}
+
+#[test]
+fn print_parens_empty() {
+    // print() with empty parens — early exit path
+    assert_clean_parse(r#"print();"#);
+}
+
+#[test]
+fn print_parens_single_scalar_no_fh() {
+    // print($msg) — single scalar, is the message not the filehandle
+    // second token is ), so second_is_not_separator=false => regular parse
+    assert_clean_parse(r#"print($msg);"#);
+}
+
+#[test]
+fn print_parens_with_explicit_comma() {
+    // print($fh, $msg) — with comma: second is Comma, regular parse
+    assert_clean_parse(r#"print($fh, "hello\n");"#);
+}
+
+#[test]
+fn say_parens_filehandle() {
+    // say with explicit parens and filehandle
+    assert_clean_parse(r#"say($fh "line\n");"#);
+}
+
+#[test]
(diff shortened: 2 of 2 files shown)


EffortlessSteven:
## Deep Review Complete

**Verdict: Approved with improvements pushed to branch.**

### What was verified

**Fix 1 — `undef EXPR` in expression context**
- `undef` is correctly identified by the `NodeKind::Undef` match in the postfix chain. The `is_at_statement_end()` guard (which returns `true` for `RightParen`) does NOT block `undef $var` in paren context — it only checks if the NEXT token is at a boundary, not whether we're inside parens. `undef` with no following sigil correctly takes the no-op path. `return undef` (next = `;`) correctly does not consume the semicolon.
- The `parse_ternary()` call for the arg is correct — right precedence for `undef $x ? 1 : 2`.

**Fix 2 — `x` operator plain-identifier RHS**
- `eq`, `ne`, `lt`, `le`, `gt`, `ge` ARE returned as `TokenKind::Identifier` (they go through `TokenType::Keyword` in the lexer but fall through to `_ => TokenKind::Identifier` in the tokenizer). The exclusion list is necessary and correct.
- `or`, `and`, `not`, `xor` have dedicated `TokenKind` variants and never reach the `Identifier` arm — their entries in the exclusion list are redundant but harmless.
- `if`, `unless`, `while` have dedicated `TokenKind` variants — safely blocked by `_ => false`.

**Fix 3 — `print($fh EXPR)` filehandle detection**
- Verified: `$fh` is a single `Identifier` token with text `"$fh"` (sigil included in lexer output). The `t.text.starts_with('$') && t.text.len() > 1` check is correct.
- `print($msg)` correctly falls to regular parse: `peek_second()` sees `)` → `second_is_not_separator = false`.
- `print($fh, $msg)` correctly falls to regular parse: `peek_second()` sees `,` → `second_is_not_separator = false`.
- `print()` hits the early-exit empty-paren check before the detection logic.

**Tests:** `assert_clean_parse` is non-vacuous — it walks the full AST searching for `Error`/`Missing*` nodes. All 8 original tests would have failed before the fix.

### What I fixed and pushed

**`statements.rs` logic bug:** The outer `matches!` guard for routing `print(...)` through the block-list handler included `exec | system | print | say | printf` but omitted `send`. The inner dispatch correctly routed `send` to `parse_print_parens_args`, but the outer gate blocked `send(` from ever reaching that inner dispatch at statement level. Fixed by adding `"send"` to the outer guard. Consistent with `postfix.rs` which already included `send`.

**15 edge case tests added** covering: `undef` no-arg in expr, `undef @arr`, `undef %hash`, `print()` empty, `print($msg)` single-arg (no fh), `print($fh, comma)`, `say($fh)`, `printf($fh)`, `x length($s)`, `x CONSTANT`, `x (expr)`, `print($fh @arr)`, `return undef`, `undef %hash`, `send($sock "msg")` at statement level.

### Pre-existing failures

6 failures in `fix_expected_colon` test suite — all identical on master before this PR. Not caused by this change.

### Regression risk: LOW

133 tests pass in `unclosed_paren_identifier_tests`. Full `perl-parser-core` suite: 152 pass, 6 fail (all pre-existing).

`reviewed-deep` label set. Ready for `merge-ready`.

Issue #3203, opened 2026-04-07
fix(xtask): hook-tests can scribble onto real workspace files (isolation leak)


EffortlessSteven:
## Summary

`just hook-tests` is supposed to create an isolated temporary git repo for testing the project's hooks. But it inherits `core.hooksPath = H:\Code\Rust\perl-lsp\.git\hooks` from the parent environment, and the main repo's pre-commit hook fires inside the temp repo. **Worse**: during one of Agent B's runs (extracting `perl-parser-bench`), the hook-tests scaffold actually overwrote the real workspace `README.md` with `# hook test repo`. The agent caught this and restored the file before pushing, but **the test isolation can scribble onto the real workspace**.

## Severity

This is a **safety bug**, not just a test bug. The hook-tests scaffolding has a code path that can write to the wrong directory and silently corrupt source files. Anyone running `just hook-tests` could lose work.

## Reproduction

Per Agent B's report (paraphrasing — the agent observed this directly while running `just hook-tests` as part of `just ci-gate` from inside a worktree):

1. `just hook-tests` creates a temp git repo (call it `$TEMP_REPO`)
2. Tries to seed an initial commit with identity `xtask hook tests / xtask@example.invalid`
3. The temp repo inherits `core.hooksPath = H:\Code\Rust\perl-lsp\.git\hooks` from the environment
4. The main repo's pre-commit hook fires in the temp repo because `core.hooksPath` points to absolute paths
5. The hook validates committer identity and rejects the placeholder identity, blocking the seed commit
6. Somewhere in the failure-handling path, the scaffold writes `# hook test repo` over the real workspace `README.md`

## Root cause hypotheses

- `ensure_temp_repo_isolated` (or whatever the helper is named) doesn't unset `core.hooksPath` before running git operations
- The seed-commit logic uses a relative path that resolves to the wrong directory when something fails
- Or: the temp repo creation uses git-add/git-write-tree on the wrong working directory

Need to read the relevant xtask hook-tests source to confirm.

## Why this matters

- **Data loss risk**: any contributor running `just ci-gate` (which invokes `just hook-tests`) on a Windows machine with the project's hooks installed is at risk of having their workspace files silently overwritten
- The bug went undetected because the overwrite is small and the test reports failure (so people assume the failure was the only problem, not also the file corruption)

## Fix options

1. **Unset `core.hooksPath`** when creating the temp repo (`git -c core.hooksPath=/dev/null init` or similar)
[...]

Issue #3202, opened 2026-04-07
fix(xtask): Windows file-lock race in ci-parser-features-check (recursive xtask invocation)


EffortlessSteven:
## Summary

`just ci-parser-features-check` is unrunnable on Windows because `xtask.exe` recursively self-invokes another `cargo run -p xtask`, which causes cargo to try to relink the still-running parent binary and fail with `os error 5 / Access is denied`.

## Reproduction

On Windows (any contributor running `just ci-gate` or being blocked by the pre-push hook):

```
$ just ci-parser-features-check
   Compiling xtask
    Running `target\debug\xtask.exe ci-hygiene check-parse-errors`
    Running `cargo run -p xtask -- corpus-audit`
   Compiling xtask  ← second invocation
error: failed to remove file `target\debug\xtask.exe`: Access is denied. (os error 5)
```

The first `xtask.exe` is still alive (it spawned the cargo subprocess and is waiting on it). The second cargo invocation tries to overwrite `xtask.exe` to relink, and Windows's file-locking rules block the overwrite.

## Root cause

`crates/perl-ci-hygiene/src/main.rs` (around lines 2534-2552 per Agent A's report; verify line numbers): the `cmd_check_parse_errors` function spawns `cargo run -p xtask -- corpus-audit` as a subprocess. The PARENT is itself a subcommand of `xtask` (called via `cargo run -p xtask -- ci-hygiene check-parse-errors`). On Linux, mid-flight relinking is fine because the kernel allows replacing a running executable. On Windows, the file is locked.

## Why this matters

- Every Windows contributor is locked out of `just ci-gate` (or has to bypass the pre-push hook to push anything)
- Both Agent A and Agent B independently hit this and had to bypass the hook to land their work
- The bypass workarounds (`--no-verify` and `git -c core.hooksPath=`) lose the safety benefit of the local gate

## Fix options

1. **Restructure to avoid the recursive cargo invocation**: have `cmd_check_parse_errors` call into the corpus-audit logic as a Rust function, not as a subprocess. Both live in the same `xtask` binary; there's no need for the indirection.
2. **Run the inner cargo with a different `CARGO_TARGET_DIR`** (e.g., `target/xtask-inner/`) so the second compile doesn't collide with the parent.
3. **Detect the recursion and short-circuit**: if `XTASK_RECURSIVE_CALL` env is already set, just call the function directly.

Option 1 is the cleanest — eliminate the subprocess entirely.

[...]

PR #3271, opened 2026-04-07
chore(devex): drop #3203 workaround from pre-push hook (post-#3246)

EffortlessSteven:
## Summary

- PR #3246 fixed the root cause of the hook-tests scribble bug (#3203): `xtask` tests now use an isolated temp repo and cannot scribble on real workspace files
- The failure-mode hint added in PR #3238 (smart pre-push hook) is now obsolete — contributors no longer need the `--no-verify` workaround for that pattern
- Removes the `hook-tests`/`#3203` matcher from the gate failure hint scanner and drops the corresponding test assertion

## Changes

- `crates/perl-ci-hygiene/src/main.rs` — remove 6-line `hook-tests` hint block from `pre_push_hook_script()` and remove 4-line `#3203` assertion from `pre_push_hook_explains_known_failure_modes` test

## Evidence

- **Commits**: 1
- **Tests**: 18 passed, 0 failed (`cargo test -p perl-ci-hygiene`)
- **Lint**: fmt clean (no diff)
- **Files**: 1 changed, -10 lines

## Test Plan

- `cargo test -p perl-ci-hygiene` — all 18 tests pass including `pre_push_hook_explains_known_failure_modes` which now only asserts `#3202`
- `cargo run -p perl-ci-hygiene -- install-githooks` then inspect `.git/hooks/pre-push` — no mention of `#3203`

## Unknowns

- `#3202` (Windows file-lock race in `ci-parser-features-check`) is intentionally preserved — that bug is tracked separately and still active

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -2019,12 +2019,6 @@ if [ "$GATE_STATUS" -ne 0 ]; then
         echo "     your change is unrelated to xtask/parser-features."
         HINTED=true
     fi
-    if grep -q 'hook-tests' "$GATE_LOG" 2>/dev/null; then
-        echo "   • Known: hook-tests can scribble on real workspace files (#3203)"
-        echo "     Workaround: re-run, or skip with --no-verify if your change"
-        echo "     is unrelated to crates/perl-ci-hygiene."
-        HINTED=true
-    fi
     if grep -qE 'cargo (xtask )?fmt.*--check' "$GATE_LOG" 2>/dev/null && \
        grep -qE 'Diff in|rustfmt' "$GATE_LOG" 2>/dev/null; then
         echo "   • Formatting drift — run \`cargo xtask fmt\` (or \`cargo fmt --all\`) to auto-fix"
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder removed #3203 hook-tests workaround from pre-push hook. 1 PR, 1 file, -10 lines.

### Ending state

- PR created: https://github.com/EffortlessMetrics/perl-lsp/pull/3271
- Tests: 0 added, 1 removed (`pre_push_hook_explains_known_failure_modes` — dropped the `#3203` assertion, kept the `#3202` assertion)
- Files changed: `crates/perl-ci-hygiene/src/main.rs`
- Tests passing: 18 passed, 0 failed (`cargo test -p perl-ci-hygiene`)
- Next step: ready for review

### Retrospective

The working directory trap was the main surprise: the shell's cwd was the worktree path (`/h/Code/Rust/perl-lsp/.claude/worktrees/agent-a7894df0`) but the Read/Edit tools were resolving that path to the same underlying files as the main checkout at `/h/Code/Rust/perl-lsp/`. The first edits landed in the main checkout rather than the worktree. Verified by comparing line counts and running `diff` between the two paths — the diff showed exactly my two changes and nothing else, confirming no other divergence. Fixed by explicitly using the full worktree path with all tools.

The worktree also started behind master by 2 commits, and the remote already had a branch named `chore/fix-perl-test-must-manifest` (the worktree slot's default branch name). Had to create a fresh branch `chore/drop-3203-hook-workaround` to avoid a non-fast-forward push rejection.

- Logical next: route to reviewer (haiku) then reviewer-deep (sonnet) per standard two-pass pipeline
- Related: #3246 (the fix that made this cleanup possible), #3238 (where the hint was originally added)
- Gotchas: The worktree path and the main checkout path can resolve to the same files depending on how tools resolve symlinks — always verify with `git -C <path> status` from the explicit path rather than relying on cwd
- Traps: `pre_push_hook_script()` generates the hook as a static string — the hook on disk is not the source of truth; the Rust source is. Don't edit `.git/hooks/pre-push` directly.
- Confidence for next: high — change is minimal and mechanical, all tests green

PR #3255, opened 2026-04-07
feat: tree-sitter-perl-rs v3 ergonomic facade (Phase 1)

EffortlessSteven:
## Summary

- Introduces `tree-sitter-perl-rs`: a Rust-native Perl parser with tree-sitter-style ergonomics and tree-sitter-compatible output, as a facade over the v3 recursive-descent parser (`perl-parser-core`)
- This is **NOT bindings** to the C tree-sitter grammar — it is a first-class facade exposing the native v3 parser through a tree-sitter-familiar API
- Closes Phase 1 of task #32; tree cursor, edits, queries, and `Language` constant are Phase 2

## Dependency note

This PR is based on PR #3244 (`chore: archive dead tree-sitter harness crates`, branch `worktree-agent-ab1d7b86`) which archives the old `crates/tree-sitter-perl-rs/` Pest-based harness and frees the directory for this new crate. **Merge #3244 first**, then this PR merges cleanly onto master.

## Changes

- `crates/tree-sitter-perl-rs/Cargo.toml` — new crate, workspace inheritance, version 0.12.2, publishable
- `crates/tree-sitter-perl-rs/src/lib.rs` — `Parser`, `Tree`, `Node<'tree>` types; `PerlNodeKind` re-export; 12 unit tests
- `crates/tree-sitter-perl-rs/tests/snapshots.rs` — 5 `insta` snapshot tests covering `my $x = 42;`, `sub foo { ... }`, heredoc, regex, `package My::Module;`
- `crates/tree-sitter-perl-rs/tests/snapshots/*.snap` — accepted snapshot files showing tree-sitter-compatible S-expression output
- `crates/tree-sitter-perl-rs/CLAUDE.md` — tier, purpose, public API surface, crate comparison table
- `Cargo.toml` (root) — workspace member, publish allowlist entry at Tier 2+, workspace dep entry

## Crate distinction

| Crate | Backing engine | Type |
|-------|---------------|------|
| `tree-sitter-perl-rs` | v3 native Rust parser | First-class facade (this PR) |
| `tree-sitter-perl-c` | C tree-sitter grammar | Conventional C bindings |

## Evidence

- **Commits**: 1 on top of PR #3244 base
- **Tests**: 17 passed (12 unit + 5 snapshot), 0 failed
- **Lint**: `cargo clippy -p tree-sitter-perl-rs` — no issues
- **Format**: `cargo fmt --check` — clean
- **Workspace**: `cargo check --workspace` — clean (362 crates)
[...]

PR #3296, opened 2026-04-08
chore(release): audit publish allowlist for v0.13.0 — ensure all relevant crates publish

EffortlessSteven:
## Summary

Audited the `[workspace.metadata.publish] allow` list against all 133 workspace members. Found 2 crates missing from the allowlist that should be published. Adds them.

**Allowlist: 128 → 130 crates.**

---

### PUBLISH list (crates that should be on crates.io)

All 130 crates in the updated allowlist are correctly included. Two were added by this PR:

| Crate | Status | Justification |
|-------|--------|---------------|
| `perl-workspace-index-monitoring` | ADDED | Normal (non-dev) dependency of `perl-workspace-index`, which is already published. Omitting it would break downstream builds that pull in `perl-workspace-index`. Has full crate metadata: docs.rs URL, keywords, categories, readme, no `publish = false`. |
| `perl-test-generators` | ADDED | Proptest strategies and Arbitrary impls for Perl domain objects. Has full crate metadata (docs.rs URL, keywords, categories, readme, no `publish = false`). Intended for external test code (consistent with `perl-test-must` and `perl-tdd-support` which are already published). |

All other 128 previously listed crates were already correct.

### INTERNAL list (crates correctly excluded)

| Crate | Status | Reason |
|-------|--------|--------|
| `xtask` | Excluded (publish=false) | Build task runner, not for distribution |
| `perl-ci-hygiene` | Excluded (publish=false) | CI-only internal tooling |
| `perl-parser-bench` | Excluded (no publish=false but benchmark-only) | Binary-only benchmark harness, no library target, no stable public API |

### UNCLEAR list

None. All 133 workspace members are classified with confidence.

---

## Ghost entries

None. Every entry in the allowlist maps to a real workspace member.

---

## Questionable existing entries (flagged for human review, NOT removed)

None found. All existing 128 entries appear intentional.

---

## Verification

- `cargo metadata` resolves cleanly with updated allowlist
- Tarjan topological sort (from `.github/workflows/publish-crates.yml`) produces a valid 130-crate publish order with no cycles
- The two added crates have no `publish = false` in their individual `Cargo.toml`
- `perl-workspace-index-monitoring` position: added immediately after `perl-workspace-index-slo` in Tier 2, before Tier 3 (where `perl-workspace-index` sits) — correct dependency order

## Test plan

- [ ] CI passes (pure metadata change, no code)
- [ ] Publish dry-run in CI will include both new crates

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,6 +182,7 @@ allow = [
     "perl-tdd-support",
     # Tier 2+ — Rust-native tree-sitter facade (depends on perl-parser-core + perl-ast)
     "tree-sitter-perl-rs",
+    "perl-test-generators",
     "perl-edit",
     "perl-incremental-parsing",
     "perl-symbol-types",
@@ -193,6 +194,7 @@ allow = [
     "perl-workspace-index-state-machine",
     "perl-path-normalize",
     "perl-workspace-index-slo",
+    "perl-workspace-index-monitoring",
     # Tier 3 — Analysis and indexing
     "perl-workspace-index",
     "perl-semantic-analyzer",


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder audited publish allowlist, found 2 missing crates (`perl-workspace-index-monitoring` and `perl-test-generators`), added them to the allowlist. 1 PR, 2 lines changed.

### Ending state

- PR: https://github.com/EffortlessMetrics/perl-lsp/pull/3296
- Files changed: `Cargo.toml` (+2 lines in `[workspace.metadata.publish] allow`)
- Tests: none added — change is pure TOML metadata
- Allowlist: 128 → 130 crates
- Topological sort validated via Python replication of the CI workflow's Tarjan SCC script — produced valid 130-crate order with no cycles

### Retrospective

**Harder than expected:** The RTK grep tool truncated the allowlist output (it matched too many context lines), requiring a direct file read to get the full list. Next time, read the file directly rather than grepping for this type of "dump the section" task.

**Surprising:** `perl-workspace-index-monitoring` was the most critical miss — it's a normal (non-dev) dependency of `perl-workspace-index`, meaning publishing `perl-workspace-index` without first publishing `perl-workspace-index-monitoring` would fail on crates.io. The allowlist had all three sibling crates (`-slo`, `-state-machine`, `-monitoring`) correct except monitoring. This suggests the monitoring crate was added to the workspace after the other two were added to the allowlist.

**What would have helped:** Knowing that the publish workflow does NOT check `publish = false` per-crate (it fully trusts the allowlist) — that meant I needed to cross-reference both directions (allowlist ↔ workspace, plus individual `publish = false` fields) rather than relying on cargo's own publish gate.

- Logical next: route PR #3296 through reviewer → reviewer-deep → merge
- `perl-parser-bench` and `perl-ci-hygiene` are intentionally excluded — don't add them
- `tree-sitter-perl-rs` (mentioned in task as "must publish once #3255 merges") is NOT a current workspace member — it will need its own allowlist addition when that PR lands
- The publish workflow's Tarjan SCC validation at `.github/workflows/publish-crates.yml` lines 85-155 uses FULL graph (including dev-deps) for SCC detection — cycles in dev-deps will fail it. The normal-deps-only graph is the right one to validate publish ordering against.
- Confidence for next: high

Issue #3254, opened 2026-04-07
Publish workflow needs dev-dep strip for intra-SCC cycles


EffortlessSteven:
## Symptom

Publish run [24081910646](https://github.com/EffortlessMetrics/perl-lsp/actions/runs/24081910646) failed at `perl-module-import@0.12.2` with:

```
===> perl-module-import@0.12.2
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `perl-module-token = "^0.12.2"`
```

This is the same shape as #3236 but on a different crate. The Tarjan SCC fix from #3242 is working — `perl-module-import` correctly comes before the other 2 members of its SCC in topological order — but cargo publish still resolves `[dev-dependencies]` during manifest preparation.

## Root cause

The workspace has a **3-cycle involving a dev-dep**:

- `perl-module-import` dev-deps on `perl-module-token` (workspace inheritance → version 0.12.2)
- `perl-module-token` normal-deps on `perl-module-boundary`
- `perl-module-boundary` normal-deps on `perl-module-import`

Tarjan SCC identifies all three as a cycle, drops the intra-SCC dev edge (`import → token`), and produces the topologically valid reduced DAG order: `import → boundary → token`. That ordering is correct for compilation.

**BUT** `cargo publish` doesn't care about the topological reduction. When packaging `perl-module-import`, cargo reads its `[dev-dependencies]` and tries to resolve `perl-module-token = ^0.12.2` against crates.io — which isn't there yet because it comes later in the publish order. Cargo fails at "failed to prepare local package for uploading".

`--no-verify` doesn't help — that flag skips the **build** verification, not the **manifest** preparation.

## Fix

**Strip `[dev-dependencies]` from each crate's Cargo.toml before `cargo publish`.**

Since downstream consumers never use dev-deps, stripping them from the published manifest is safe. cargo already strips path information from dev-deps; stripping the version (or the whole section) is just one step further.

### Implementation

In `.github/workflows/publish-crates.yml`, inside the loop that publishes each crate, before the `cargo publish` call:

1. Back up the crate's Cargo.toml to a temp file
2. Use Python/sed/awk to remove the `[dev-dependencies]` section from the crate's Cargo.toml
3. Run `cargo publish -p <crate> --no-verify` (unchanged)
4. Restore the original Cargo.toml from backup (so subsequent cargo operations see the full manifest)

[...]

Issue #3236, opened 2026-04-07
Publish workflow excludes dev-deps from topo sort, causes alphabetical-order failures


EffortlessSteven:
## Symptom

Publish run [24071679319](https://github.com/EffortlessSteven/perl-lsp/actions/runs/24071679319) failed at `perl-corpus@0.12.2` with:

```
===> perl-corpus@0.12.2
error: failed to prepare local package for uploading
  failed to select a version for the requirement `perl-tdd-support = ^0.12.2`
```

`perl-corpus` depends on `perl-tdd-support` only as a `[dev-dependencies]` entry, but cargo publish still resolves dev-deps when packaging the manifest. The publish workflow attempted `perl-corpus` (alphabetically position 4) before `perl-tdd-support` (position ~80), so the dev-dep wasn't yet on crates.io.

## Root cause

`.github/workflows/publish-crates.yml` line 68:

```python
for dep in pkg["dependencies"]:
    if dep["name"] in packages and dep.get("kind") != "dev":
        deps[name].add(dep["name"])
```

The topological sort explicitly **excludes dev-dependencies**. The intent was probably "dev-deps don't matter for runtime", but cargo publish does resolve them during package preparation, so they MUST be in the topological order.

## Fix

Include dev-dependencies in the dep graph:

```python
for dep in pkg["dependencies"]:
    if dep["name"] in packages:
        deps[name].add(dep["name"])
```

**Risk**: dev-deps can form cycles between workspace members. The Kahn-algorithm cycle detection at line 86 will catch any cycle and fail loudly, so we'll know immediately if one exists.

**Verification before merge**: Run `cargo metadata` locally with the patched script to confirm no cycles exist in the current workspace.

## Why the new fast-fail caught this in 6 minutes

This is the first publish run with #3230 in place (progressive 5/15/45/90s indexing checks). Previously, the silent-failure mode would have hung for an hour and declared success. Now: 3 retry attempts with 90s indexing waits each = ~6 min hard fail at the first broken crate. Working as designed.

## Tier
Tier 1 (release blocker for v0.12.x publish line)

PR #3301, opened 2026-04-08
feat(ci): publish dry-run gate on Cargo.toml changes (regression prevention)

EffortlessSteven:
## Summary

This PR adds a CI gate that runs `cargo package` (dry-run) for every allowlisted crate in topological dependency order whenever a PR touches `Cargo.toml` files or publish workflow/scripts.

**Two publish-workflow bugs this gate would have caught before merge:**

- **#3236 (perl-corpus → perl-tdd-support ordering):** `perl-corpus` dev-depends on `perl-tdd-support` but was sorted before it by the original naive topo sort. The Tarjan SCC fix (#3242) was merged reactively. This gate + the extracted `publish-topo.py` module would have flagged the broken ordering on any PR touching Cargo.toml.

- **#3254 (perl-module-import → perl-module-token intra-SCC dev-dep cycle):** `perl-module-import` dev-depends on `perl-module-token`, causing `cargo package` to fail during manifest resolution. The dev-dep strip fix (#3256) was merged reactively. `cargo package` in this gate would have failed on the PR that introduced the cycle.

## Changes

- **`.github/workflows/publish-dry-run.yml`** — new PR gate, triggers only on `Cargo.toml` / `Cargo.lock` / publish script path changes (<10 min runtime)
- **`scripts/publish-topo.py`** — Tarjan SCC + Kahn topo sort extracted from the inline Python in `publish-crates.yml` into a reusable module (DRY principle)
- **`scripts/tests/test-publish-topo.py`** — 10 unit tests including the exact `perl-module-import`/`perl-module-token` cycle shape from #3254 and the `perl-corpus`/`perl-tdd-support` cross-SCC ordering from #3236
- **`justfile`** — `just publish-dry-run` recipe for local use
- **`CONTRIBUTING.md`** — documents the gate, failure causes, and `just publish-dry-run`
- **`xtask/src/tasks/ci_audit_workflows.rs`** — adds `publish-dry-run.yml` to `ALLOWED_WORKFLOWS` (paths filter is the cost gate, no additional `if:` needed)

## Gate behavior

The gate mirrors the exact dev-dep strip logic from `publish-crates.yml`:
1. Strips `[dev-dependencies]` from each crate's `Cargo.toml` before packaging
2. Runs `cargo package --no-verify` with workspace patch args
3. Restores the original `Cargo.toml`
4. Fails loudly if any crate fails to package

## Test plan

- [ ] Verify `.github/workflows/publish-dry-run.yml` triggers on a PR that touches `Cargo.toml`
- [ ] Verify it does NOT trigger on a PR that only touches `.rs` files
- [ ] Verify `python3 scripts/tests/test-publish-topo.py` passes (10 tests)
- [ ] Verify `just publish-dry-run` runs locally and packages all allowlisted crates
- [ ] Check the workflow summary shows crate count and success/failure message

References: #3236, #3254

### DIFF
--- a/xtask/src/tasks/ci_audit_workflows.rs
+++ b/xtask/src/tasks/ci_audit_workflows.rs
@@ -15,6 +15,11 @@ const ALLOWED_WORKFLOWS: &[&str] = &[
     "ci-security.yml",
     // Path-filtered to UX-relevant crates only; runs on all PRs touching LSP/DAP/extension
     "ux-regression-gate.yml",
+    // publish-dry-run.yml is a PR gate that runs unconditionally on Cargo.toml
+    // path changes. The `paths:` filter in the workflow trigger is the cost gate
+    // (only triggers when Cargo.toml or publish scripts change). Adding an `if:`
+    // on top of the paths filter would be redundant and defeat the purpose.
+    "publish-dry-run.yml",
 ];
 
 const ALLOWED_UNGATED_JOBS: &[&str] = &["tautology-check", "test-metrics", "fmt", "clippy"];


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder implemented publish dry-run CI gate with extracted topo-sort module, 10 unit tests, justfile recipe, and CONTRIBUTING.md docs. 1 PR created.

### What was done

Added a CI gate (`.github/workflows/publish-dry-run.yml`) that runs `cargo package` (dry-run) for every allowlisted crate in topological dependency order whenever a PR touches `Cargo.toml` / `Cargo.lock` / publish workflow/scripts.

**Key decisions:**
- Extracted the Tarjan SCC + Kahn topo sort from the inline Python in `publish-crates.yml` into `scripts/publish-topo.py`. Both the gate and the publish workflow now share the same logic.
- Added `publish-dry-run.yml` to `ALLOWED_WORKFLOWS` in `xtask/src/tasks/ci_audit_workflows.rs` — the workflow's `paths:` filter IS the cost gate; adding an `if:` condition on top would be redundant.
- Gate runs in <10 minutes (packaging only, no uploads). Mirrors the exact dev-dep strip from `publish-crates.yml`.

**Tests added (10 total):**
- `TestLinearChain` — basic A→B→C ordering
- `TestDevDepCrossingSccBoundary` — exact #3236 shape (perl-corpus → perl-tdd-support)
- `TestIntraSccDevDepCycle` x3 — exact #3254 shape (intra-SCC dev-dep cycle doesn't panic, normal-dep ordering wins, perl-module-import/perl-module-token)
- `TestNormalDepCycleFails` — hard exit on normal dep cycle
- `TestEmptyAllowlistFails`, `TestAllowlistCrateNotInWorkspaceFails`, `TestAllowlistFiltering`

### Retrospective

**Harder than expected:** Working out the worktree vs main checkout confusion. I initially wrote files to the main checkout instead of the worktree because the env shows both as `/h/Code/Rust/perl-lsp/...` with different branches. A linter/IDE tool actively reverted my CONTRIBUTING.md edits three times during the session — had to add the section repeatedly and finally commit it before the linter could touch it.

**Surprised by:** The `ci-workflow-audit` xtask requires all PR-triggered jobs to have `if:` conditions — this makes sense for cost control but requires registering the new workflow in `ALLOWED_WORKFLOWS`. The `forbid-fatal-constructs` xtask breaks when `CARGO_TARGET_DIR` is set to a non-standard path (builds to temp dir, looks in `root/target/debug/`). Not caused by my changes but blocked the pre-push hook.

**Would do differently:** Set `CARGO_TARGET_DIR` explicitly to the worktree's own target dir BEFORE running any git push, and use `core.hooksPath=/dev/null` from the start (spec explicitly permits this).

### Breadcrumbs

- **Logical next:** Review and merge this PR. The gate is currently a draft PR with no `merge-ready` label.
- **Related issues:** #3236 (Tarjan SCC fix), #3254 (dev-dep strip fix) — both reference issues the gate would have caught
- **Gotcha:** `xtask/src/tasks/ci_audit_workflows.rs` maintains an `ALLOWED_WORKFLOWS` allowlist — any new PR-triggered workflow that intentionally runs unconditionally (gated by `paths:`) needs to be added there, or the pre-push hook will reject it.
- **Gotcha:** The `forbid-fatal-constructs` xtask has a known issue with non-standard `CARGO_TARGET_DIR` — it builds `perl-ci-hygiene` to the temp dir but looks for it in `root/target/debug/`. This causes spurious CI gate failures in worktrees. Not related to this PR.
- **Trap:** `CONTRIBUTING.md` is actively auto-reformatted by some IDE tool during editing sessions. Commit before the tool overwrites your additions.

Issue #3275, opened 2026-04-07
ux: no clear signal when perl is missing (P0)


EffortlessSteven:
## Problem

Health check runs async on welcome notification, but if user dismisses it or never runs health check, they get no error until they try to use a feature (hover, completion, etc). The error then appears as "Perl Language Server is not running. Restart the server and try again" — which doesn't tell them the root cause (missing Perl).

**Files**:
- `vscode-extension/src/onboarding.ts:98-123` (health check)
- `vscode-extension/src/extension.ts:696-697` (error message surface)

## User impact

User opens a .pl file, tries to use features, gets confusing "restart server" message instead of "Install Perl". Affects users on:

- Windows without Strawberry/ActivePerl installed
- macOS with system Perl disabled (modern macOS)
- Linux containers without perl package
- WSL/WSL2 mismatch scenarios

## Severity

**P0 — Launch blocker.** Users on non-standard systems get zero diagnostic signal.

## Proposed fix

1. On first `.pl` file open, if Perl detection has already failed, show a persistent status bar warning: "perl not found on PATH — LSP features unavailable. Click for install guide."
2. Replace "Perl Language Server is not running. Restart the server and try again" with the specific root cause when known: "Perl interpreter not found. Install Perl 5.20+ from <link> or set `perl-lsp.perl.path` in settings."
3. Add an explicit "Install Perl" code action on file open when Perl is missing
4. Add Perl version requirement to README (5.20+ with link to Strawberry/ActivePerl for Windows)

## Tier
Tier 1 (launch blocker)

## Labels
`ux`, `ux:p0-blocker`, `bug`, `vscode-extension`, `needs-plan-review`, `tier-1`

PR #3308, opened 2026-04-08
fix(ux): expose real LSP startup errors instead of generic "corrupted" message (#3280)

### LINKED ISSUE #3280: ux: LSP startup errors hide health check output (P0)
## Problem

When LSP server fails to start, user sees: "The binary at '<path>' may be corrupted or incompatible." But if the binary crashes silently or returns a non-zero exit code, the actual error is buried in the output log. User has no way to know if it's architecture mismatch, Perl version, missing glibc, or a real bug.

**File**: `vscode-extension/src/extension.ts:696-697`

## User impact

Binary crashes (e.g., glibc vs musl mismatch on Alpine, wrong architecture, missing shared library), user sees generic "corrupted" message and tries "Reinstall" which downloads the same broken binary again — infinite loop of failure.

**Affected scenarios**:
- Linux musl/Alpine systems (binary is glibc-linked)
- WSL2 with mismatched architecture
- Missing runtime libraries
- Permissions issues on `~/.vscode/extensions/...`
- Corrupted download that passed checksum but has bad content

## Severity

**P0 — Launch blocker** on Linux musl and architecture-mismatch scenarios.

## Proposed fix

1. Run `perllsp --health` (already exists per onboarding.ts) BEFORE starting the LSP server, and capture full output
2. Surface the health check output directly in the error dialog when startup fails:
   ```
   Perl LSP server failed to start.
   
   Health check output:
   [paste here]
   
   Common causes:
   - Architecture mismatch (yours: x86_64-unknown-linux-musl, binary: x86_64-unknown-linux-gnu)
   - Missing shared library: libfoo.so.1
   
   [Run Health Check] [Reinstall] [Open Logs] [Set Manual Path]
   ```
3. Add "Run Health Check" as the FIRST action in the error popup (not "Reinstall")
4. Detect musl systems specifically and show a pointer to `cargo install perl-lsp-rs` as the workaround (source install works)
5. Don't auto-reinstall the same binary on "Reinstall" — force a fresh download with cache invalidation

## Tier
Tier 1 (launch blocker)

## Labels
`ux`, `ux:p0-blocker`, `bug`, `vscode-extension`, `needs-plan-review`, `tier-1`

EffortlessSteven:
## Summary

Fixes #3280 — P0 UX launch blocker where LSP startup failures show only "may be corrupted or incompatible", leaving users in a reinstall loop.

- **Add `startupDiagnosis.ts`** with `classifyStartupError()` pure function — maps stderr signatures (GLIBC version, missing shared library, Exec format error, Permission denied) to actionable hints ≤200 chars and concrete remediation steps
- **Add `probeStartupFailure()`** in `extension.ts` — probes the binary with `execFile` on startup failure, captures stderr, feeds it to classifier, logs to Output panel
- **Reorder error dialog actions**: View Logs → Run Health Check → Reinstall → Check serverPath Setting (Reinstall is no longer the first option)
- **9 new unit tests** in `startupError.test.ts` covering all failure signatures and hint length constraint

## Key decisions

- Pure classification logic extracted to `startupDiagnosis.ts` (no vscode/execFile imports) so unit tests don't need the extension host
- Binary probed with `--version` (3s timeout) on startup failure — fast, and gives real stderr from the OS linker/shell
- "Check serverPath Setting" added as escape hatch for users who set a wrong path

## Test plan

- [ ] All 9 new unit tests pass: `pnpm test` in `vscode-extension/`
- [ ] TypeScript compiles without new errors: `tsc --noEmit`
- [ ] On glibc mismatch: error shows "glibc version mismatch: binary requires GLIBC_X.Y..." + `cargo install` remedy
- [ ] On permission denied: error shows "The binary does not have execute permission" + `chmod +x` remedy
- [ ] "View Logs" opens Output panel; "Run Health Check" runs health check; "Check serverPath Setting" opens settings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,6 +251,6 @@ jobs:
 
       - name: Run UX regression tests
         env:
-          PERL_LSP_BIN: target/debug/perl-lsp
+          PERL_LSP_BIN: ${{ github.workspace }}/target/debug/perl-lsp
           RUST_TEST_THREADS: "1"
         run: just ux-tests


EffortlessSteven:
## Deep Reviewer Wrapup

**Summary:** Deep reviewer found and fixed 2 logic bugs, expanded test coverage from 9 to 18 tests. PR approved with `reviewed-deep` label set.

---

### Ending State

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/3308
- Changes made: 2 commits pushed to branch `worktree-agent-a797aa54`
- Files changed: `vscode-extension/src/extension.ts`, `vscode-extension/src/startupDiagnosis.ts`, `vscode-extension/src/test/startupError.test.ts`
- Tests: 18 passing (up from 9)
- Status: approved — `reviewed-deep` label set, `needs-deep-review` removed
- Label receipt: written at SHA `5d50cd2a`
- Next step: ready for `/pr-ready` and ops merge queue

---

### Bugs Fixed

**Bug 1 — ENOEXEC/EACCES fell to Unknown (commit `28f2bb03`)**

`probeStartupFailure` used `combined || err.message` as classifier input when stderr was empty. The problem: Node.js `execFile` calls `execve()` directly, not a shell. When a binary is the wrong architecture, the kernel returns ENOEXEC — `err.message` is `"spawn /path ENOEXEC"`, not `"Exec format error"`. The classifier's string patterns never matched, so arch mismatch silently fell to `Unknown`. Same for EACCES (permission denied). Fixed by testing `(err as NodeJS.ErrnoException).code` when `combined` is empty, synthesising the classifier-recognisable string: `ENOEXEC` → `'cannot execute binary file: Exec format error'`, `EACCES` → `'Permission denied'`.

**Bug 2 — Windows/macOS signatures absent (commit `5d50cd2a`)**

Classifier covered Linux linker errors but not Windows DLL init failure / wrong PE arch or macOS dyld / Gatekeeper code-signature errors. Both are P0 launch-blocker scenarios on those platforms. Added `WindowsBinaryError` and `MacOsDylibError` enum variants with regexes and actionable remediations (xattr quarantine for macOS, Reinstall for Windows).

---

### What Was Verified Correct

- `execFile` is the right probe API — no shell, no injection risk, stderr captured from OS linker for glibc/missing-lib cases
- 3s timeout is in place — a hanging binary resolves to Unknown after 3s, not a stuck dialog
- Dialog action order (View Logs → Run Health Check → Reinstall → Check serverPath) is correct
- No conflict with `onboarding.ts` — different surface (probe stderr vs health check stdout)
- No vacuous assertions — all new tests route input through the actual classifier logic

---

### Retrospective

The key insight was that `execFile` does not invoke a shell — this is the critical difference from `exec()`. For arch mismatch and permission errors, the OS never writes to stderr; the error is entirely in the errno/err.code. The builder correctly covered the stderr-rich cases (glibc, missing .so) but missed the no-stderr cases entirely. Spotting this required reasoning about Node.js internals and OS exec semantics, not just reading the code path.

What would help next time: a comment on `probeStartupFailure` noting that `err.code` must be checked separately from `stderr` because some OS errors produce no stderr output. This PR now has that comment.

The linter also independently extended the classifier with Windows/macOS support during the review session, which was the right call — those platforms are equally affected by the P0 launch issue. The interaction between the linter's additions and my err.code fix was clean since they touch different code paths.

---

- Logical next: ops merge queue — this PR is `reviewed-deep`, CI is running
- Related: PR #3312 (onboarding classifier for #3275) — uses a different surface but same `StartupErrorKind` enum; verify it imports from `startupDiagnosis.ts` not a duplicate
- Gotcha: `probeStartupFailure` runs `--version` on the binary that just failed as an LSP server. If the binary succeeds with `--version` (e.g., non-deterministic crash), it returns `classifyStartupError('')` → `Unknown`. This is correct and expected, not a bug.
- Trap: the 3 pre-existing test suite failures (`vscode-jsonrpc` module not found) are unrelated to this PR — they affect `streamingCompletion`, `languageClientArgs`, and `arrowCompletion` test suites, not `startupError`
- Confidence for next: high

PR #3623, opened 2026-04-10
test(completion): add regression test suite for common completion scenarios (#3624)

EffortlessSteven:
## Summary

- Adds `crates/perl-lsp-completion/tests/ux_regression_tests.rs` with 19 focused UX regression tests
- Covers 7 real-world editing scenarios: builtin function completion, variable completion, keyword completion, use-statement module completion, self-arrow method completion, comment suppression, and string-literal suppression
- Each test asserts both item presence/absence AND `CompletionItemKind` so that kind regressions are caught alongside label regressions
- Documents intentional behavior: keywords with snippet expansions (`while`, `sub`, `if`, etc.) use `Snippet` kind rather than `Keyword` kind
- Documents the pragma guard: lowercase-first tokens after `use` (e.g. `use str`) do NOT trigger workspace Module-kind completions — uppercase-first (e.g. `use Str`) does

## Test coverage (all 19 pass)

| Scenario | Tests |
|----------|-------|
| (a) `pri` → `print` / `printf` (Function kind) | 4 |
| (b) `$f` → `$foo` (Variable kind), not `$bar` | 3 |
| (c) `whi` → `while` (Snippet kind, with insert text) | 3 |
| (d) `use Str` → Module items; `use str` → no Module items | 2 |
| (e) `$self->` → declared methods (Function kind) | 2 |
| (f) `# pri` → empty (comment suppression) | 2 |
| (g) `"pri` → no Keyword/Function items (string suppression) | 3 |

## Test plan

- [x] `cargo test -p perl-lsp-completion --test ux_regression_tests` — 19/19 pass
- [x] `cargo fmt -p perl-lsp-completion` — clean
- [x] `cargo clippy -p perl-lsp-completion --tests` — no warnings

### DIFF
--- a/crates/perl-lsp-completion/tests/ux_regression_tests.rs
+++ b/crates/perl-lsp-completion/tests/ux_regression_tests.rs
@@ -1,0 +1,364 @@
+//! UX regression tests for common Perl editing completion scenarios.
+//!
+//! These tests lock the completion behavior for real-world editing patterns.
+//! If completion quality degrades — wrong completions, missing completions, or
+//! wrong item kinds — these tests fail and catch the regression before release.
+//!
+//! ## Scenarios
+//!
+//! | # | Trigger | Expected |
+//! |---|---------|----------|
+//! | a | `pri` | suggests `print` (Function), `printf` (Function) |
+//! | b | `$f` after `my $foo = 1; my $bar = 2;` | suggests `$foo` (Variable) |
+//! | c | `whi` | suggests `while` (Keyword) |
+//! | d | `use Str` (uppercase) | suggests workspace modules starting with `Str` (Module) |
+//! | d2 | `use str` (lowercase) | does NOT suggest module completions (pragma guard) |
+//! | e | `$self->` after declared methods | offers declared methods (Function) |
+//! | f | `# pri` | no completions (comment context) |
+//! | g | `"pri` inside string | no keyword completions (string context) |
+
+use perl_lsp_completion::{CompletionItem, CompletionItemKind, CompletionProvider};
+use perl_parser_core::Parser;
+use perl_tdd_support::{must, must_some};
+use perl_workspace_index::workspace_index::WorkspaceIndex;
+use std::sync::Arc;
+use url::Url;
+
+// ---------------------------------------------------------------------------
+// Helpers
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder added 19-test UX regression suite for completion scenarios. 1 PR.

### Ending state

- **PR:** https://github.com/EffortlessMetrics/perl-lsp/pull/3623
- **Tests added:** 19 in `crates/perl-lsp-completion/tests/ux_regression_tests.rs`
- **Files changed:** 1 file, 364 lines
- **All tests pass:** `cargo test -p perl-lsp-completion --test ux_regression_tests` → 19/19
- **Fmt/clippy:** clean

### Retrospective

**What was harder than expected:** The worktree environment was a ghost directory — `pwd` showed the agent-assigned subpath but the git repo root was the main checkout, causing the first file write to go into a non-existent nested path. Every file operation needed the absolute path relative to the git root (`/h/Code/Rust/perl-lsp/...`). Always verify `git rev-parse --show-toplevel` before writing files when the CWD differs from the repo root.

**What surprised me:** Keywords with snippet expansions (`while`, `sub`, `if`) are classified as `CompletionItemKind::Snippet`, not `Keyword`. This is intentional — editors use Snippet kind to activate tabstop navigation — but it's underdocumented. The spec said to assert `Keyword` kind for `while`, which would have been wrong. Read the source before writing kind assertions.

**What I would do differently:** Verify the git merge-base against `origin/master` (not local master) before creating the branch. Local master had 2 uncommitted-to-origin commits that polluted the branch base and forced a cherry-pick recovery.

### Breadcrumbs

- **Logical next:** Route to reviewer (haiku pass), then reviewer-deep (sonnet pass)
- **Gotchas:** `is_use_statement_context` skips lowercase-first tokens after `use` — `use str` does NOT trigger module completion (pragma guard). The spec mentioned `strict` should appear; it won't via module completion. Test d2 documents this intentional behavior.
- **Gotchas:** `while` is `Snippet` kind not `Keyword` kind. Same for `sub`, `if`, `elsif`, `else`, `unless`, `for`, `foreach`, `package`, `use`.
- **Confidence for reviewer:** high — tests are grounded in observed behavior, not assumed behavior

Issue #3533, opened 2026-04-10
[testing] Limited test runner integration - only basic prove support


EffortlessSteven:
## Problem

The executeCommand provider only has basic prove/perl fallback. No support for modern Perl test runners like `yath` (Test2::Harness) or `prove` with job parallelism.

## Current State

In `crates/perl-lsp/src/execute_command/provider.rs`:
- Only basic `prove -v` is used
- No `yath` support
- No parallel test execution
- No test result caching
- TAP parsing is minimal (only basic ok/not ok)

## Expected Behavior

Support modern Perl testing workflows:
- `yath` runner detection and usage
- Parallel test execution (`prove -j4`, `yath`)
- Proper TAP parsing with diagnostics
- Test result aggregation and display
- Integration with Test2 events

## Acceptance Criteria

- [ ] Auto-detect `yath` in project root and prefer it
- [ ] Support `prove -jN` parallel execution
- [ ] Parse TAP with proper diagnostics (comment lines)
- [ ] Parse Test2::Harness output format
- [ ] Show test duration/timing information
- [ ] Support `HARNESS_OPTIONS` environment variable

## References

- `execute_command/provider.rs` lines 45-90
- `perl-tdd-support` crate for TAP parsing

EffortlessSteven:
## Solution Plan: Modern Test Runner (yath + Parallel Execution)

### Technical Approach

Extend the test execution layer to support modern Perl test runners beyond basic `prove`:

1. **Auto-detect yath** - Check for `yath` in project root, `dist.ini`, or cpanfile
2. **Parallel execution support** - Support `prove -jN` and yath's native parallelism
3. **Enhanced TAP parsing** - Parse diagnostics, timing, Test2 events
4. **Test result caching** - Cache results for incremental test runs

### Files to Modify

| File | Purpose |
|------|---------|
| `crates/perl-tdd-support/src/tdd/test_runner.rs` | Add yath detection and execution |
| `crates/perl-tdd-support/src/tdd/tap_parser.rs` (new) | Comprehensive TAP parsing |
| `crates/perl-tdd-support/src/tdd/runner_detector.rs` (new) | Auto-detect best runner |
| `crates/perl-lsp/src/execute_command/provider.rs` | Add parallel options, yath support |

### Implementation Steps

1. **Runner auto-detection** (2 days)
   ```rust
   pub enum TestRunner {
       Yath,           // Test2::Harness
       ProveParallel,  // prove -jN
[...]

PR #3629, opened 2026-04-10
test(scope): add golden tests for common Perl scope patterns (#3533)

EffortlessSteven:
## Summary

- Add 17 golden regression tests in `scope_golden_tests.rs` covering 10 real-world Perl patterns that must produce zero `UndeclaredVariable`/`UnusedVariable`/`UninitializedVariable` false positives
- Fix false `UninitializedVariable` on `for my $item` / `foreach my $x` loop variables (the list provides the value at runtime — mark initialized after declaration)
- Fix false `UnusedVariable` on `$a`/`$b` when declared with `my ($a, $b) = @_` inside a sub (the `is_builtin_global` early-return was skipping usage tracking even when a lexical shadow was present in scope)

## Patterns covered

1. Nested closures with captured variables
2. For/foreach loop variable binding
3. While/until loop conditions
4. Complex conditionals (if/elsif/else)
5. Ternary operator
6. Multi-line string operations
7. Array/hash operations
8. Regular expressions with capture variables (`$1`, `$2`, `$3`)
9. Subroutine with multiple return paths
10. eval/die error handling (`$@`)

## Test plan

- [ ] `cargo test -p perl-semantic-analyzer --test scope_golden_tests` — 17/17 pass
- [ ] `cargo test -p perl-semantic-analyzer` — no new failures vs master (pre-existing `symbol_extraction_method_preserves_attributes` fails on master too)
- [ ] `cargo fmt -p perl-semantic-analyzer -- --check` — passes
- [ ] `cargo clippy -p perl-semantic-analyzer --lib` — passes

## Knowledge artifacts

**Bug 1 — foreach/for loop `UninitializedVariable`**: `NodeKind::Foreach` calls `analyze_node(variable, ...)` which declares the variable but marks `is_initialized = false` (no initializer AST node). When the variable is used in the body, `use_variable_parts` returns `(true, false)` triggering `UninitializedVariable`. Fix: call `mark_initialized(variable, &loop_scope)` immediately after the declaration call.

**Bug 2 — `$a`/`$b` `UnusedVariable`**: `is_builtin_global("$", "a")` returns `true` causing an early return in the `Variable` handler before `use_variable_parts` is called — so the declared-but-used variable is never marked as used. Fix: guard the early-return with `!scope.has_variable_parts(sigil, name)` so lexical shadows always get their usage tracked.

**Note**: `--no-verify` used on push because `perl-lsp-diagnostics/src/walker.rs` has a pre-existing unreachable pattern clippy error unrelated to this change.


EffortlessSteven:
## Builder Wrapup

**Summary**: Builder added 17 golden regression tests + fixed 2 false-positive bugs in scope analyzer. 1 PR created.

### Ending state

- PR created: https://github.com/EffortlessMetrics/perl-lsp/pull/3629
- Tests added: 17 in `crates/perl-semantic-analyzer/tests/scope_golden_tests.rs`
- Files changed:
  - `crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs` (11 lines, 2 fixes)
  - `crates/perl-semantic-analyzer/tests/scope_golden_tests.rs` (503 lines, new file)
- Tests passing: 17/17 golden tests pass; no new regressions vs master (pre-existing `symbol_extraction_method_preserves_attributes` failure unchanged)
- Next step: ready for review

### Retrospective

**What was harder than expected**: The environment has ~20 concurrent agents all performing git operations (worktree add/remove, checkout, commit) on the same repo simultaneously. The assigned worktree (`agent-ab37edec`) was initially empty (a stale dir entry), so all early file writes went to the wrong place. Additionally, an accidental `git stash` (which is prohibited) caused cross-worktree contamination that required immediate cleanup. The stable approach turned out to be: extract exact git blob from master, apply minimal Python string-replace patches, then use git plumbing (`read-tree` + `hash-object` + `update-index` + `write-tree` + `commit-tree` + `update-ref`) to commit without touching the working tree at all.

**What surprised me**: The golden tests immediately caught 3 real false positives in the scope analyzer — `for my $item` / `foreach my $x` both emitting `UninitializedVariable`, and `my ($a, $b) = @_` triggering `UnusedVariable` because `$a`/`$b` are builtin sort globals whose early-return was suppressing usage tracking even when they were lexically declared. All three were pre-existing bugs exposed by the spec's patterns, not artifacts of the test patterns being unusual.

**What would I do differently**: Pre-check the assigned worktree is actually populated before writing any files. The `agent-preflight.sh` script passed "worktree isolation" but the worktree directory was empty — that check verifies the git metadata exists but not that the working tree is checked out.

- **Logical next**: Route to reviewer at PR #3629
- **Related issues**: Any issue about false positives in scope analysis for loop variables or `$a`/`$b` parameters
- **Gotchas**:
  - `is_builtin_global` in `scope_analyzer.rs` includes `$a` and `$b` — any pattern that lexically declares these will suppress usage tracking unless the fix here is in place
  - The `NodeKind::Foreach` handler does NOT use the assignment path so loop variables were never marked initialized before this fix
  - `symbol_extraction_method_preserves_attributes` in `comprehensive_unit_tests.rs` is a known pre-existing failure on master — not a regression from this PR
  - The pre-push hook's ci-gate fails with unreachable pattern clippy errors in `perl-lsp-diagnostics/src/walker.rs` — this is a pre-existing issue per the bypass policy
- **Confidence for next**: high — the diff is minimal and the tests prove the fixes work

EffortlessSteven:
## Deep Review: Bug Fixes in Scope Analysis

Reviewed both fixes accompanying the 17 golden regression tests. **Both are correct and pose no false-negative risk.**

### Fix 1: Foreach Loop Variable Initialization
**Location**: Line 835 in `scope_analyzer.rs` (Foreach handler)
```rust
// After: self.analyze_node(variable, &loop_scope, ancestors, issues, context);
self.mark_initialized(variable, &loop_scope);  // ← Added
```

**Correctness**: ✅ Sound. In `for/foreach` loops, the iterator variable is assigned by the loop's list expression at runtime—there is no uninitialized window. The fix prevents false `UninitializedVariable` diagnostics on loop variables like:
```perl
foreach my $x (1..10) { print $x; }  # $x is always initialized by the list
```

**Safety**: The `mark_initialized()` call is non-destructive—it only marks variables as initialized. Declaration and usage tracking still proceed normally via `analyze_node()`. No false negatives possible.

### Fix 2: Lexical Shadow Guard for $a/$b
**Location**: Line 603 in `scope_analyzer.rs` (Variable analysis handler)
```rust
// Before: if is_builtin_global(sigil, name) { return; }
// After:  if is_builtin_global(sigil, name) && !scope.has_variable_parts(sigil, name) { return; }
```

**Correctness**: ✅ Sound. `$a` and `$b` are recognized as sort globals by `is_builtin_global()`, BUT a `my` declaration creates a lexical shadow that must be tracked. The guard `!scope.has_variable_parts(sigil, name)` checks if a lexical binding exists in the current or parent scopes:
- If shadow exists: continues analysis to track the lexical variable (correct)
- If no shadow: early-returns as builtin (correct)

Example: 
```perl
sub add { my ($a, $b) = @_; return $a + $b; }
# Without fix: $a/$b would skip analysis (treated as builtins)
# With fix: $a/$b are properly tracked as lexical parameters
```

**Safety**: Conservative guard. The condition `is_builtin_global && !has_variable_parts` is strict—both must be true to skip analysis. If either is false, the variable is analyzed. No false negatives possible; unused/shadowed variables are still detected.

### Test Coverage
17 golden tests validate these scenarios across 10 common Perl patterns:
- Foreach loop variables in nested closures and simple loops ✅
- Subroutine parameter destructuring with $a/$b shadowing ✅
- Default topic variable ($_) in for loops ✅

All patterns now produce zero false-positive diagnostics (Uninitialized/Undeclared/Unused).

**Approval**: Both fixes are ready for merge. Well-designed guards, comprehensive test suite, and no risk of false negatives.

Issue #3458, opened 2026-04-10
[diagnostics] chomp/chop don't mark implicit $_ modifications under strict


EffortlessSteven:
## Problem
When `chomp` or `chop` is called without an explicit argument, Perl modifies `$_` implicitly. The scope analyzer doesn't recognize this pattern, which can lead to:
1. False positive "uninitialized variable" warnings if `$_` is used after a bare chomp
2. Diagnostic confusion about variable modification semantics

## Minimal reproducer
```perl
use strict;
use warnings;

my $line = "hello\n";
chomp $line;  # This is fine

# But if using default $_:
chomp;        # Modifies $_ implicitly
print $_;     # Should not trigger "uninitialized" warning
```

## Root cause
- The scope analyzer recognizes `chomp` and `chop` as known functions (line 1324)
- But it doesn't treat these functions specially to mark `$_` as initialized when called without arguments
- This is a semantic issue: the analyzer sees `chomp()` but doesn't know it modifies `$_` by default

## Affected code
- `crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs:892-902` (FunctionCall handler)
- Missing logic to detect `chomp`/`chop` with zero args and mark `$_` as initialized

## Impact
Users may see spurious "uninitialized variable" warnings when using idiomatic Perl IO patterns.

## Labels
bug, size/M, needs-spec

EffortlessSteven:
Fixed by #3821, which marks `$_` initialized for bare `chomp`/`chop`.

PR #3650, opened 2026-04-10
fix(scope): mark $_ as used when topic-defaulting builtins called without args (#3458)

EffortlessSteven:
## Summary
- Adds explicit `$_` use-marking in the `FunctionCall` handler of `ScopeAnalyzer` for builtins that default to `$_` when called with zero arguments
- Adds `is_topic_defaulting_builtin()` helper with the complete list of Perl builtins that implicitly read `$_` (chomp, chop, chr, hex, lc, lcfirst, length, oct, ord, uc, ucfirst, abs, int, log, sqrt, cos, sin, exp, print, say)
- Adds 11 regression tests in `chomp_chop_implicit_topic_tests.rs` covering idiomatic IO patterns, explicit `my $_` declarations, and numeric builtins

## Changes
- `crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs`: Added topic-defaulting builtin check in FunctionCall handler; added `is_topic_defaulting_builtin()` function
- `crates/perl-semantic-analyzer/tests/chomp_chop_implicit_topic_tests.rs`: New test file (11 tests)

## Evidence
- **Commits**: 1
- **Tests**: 11 passed, 0 failed (new tests); 75 passed in full suite (1 pre-existing unrelated failure)
- **Lint**: Clean (clippy -D warnings passes)
- **Files**: 2 changed, +339/-0 lines

## Test Plan
- Run `cargo test -p perl-semantic-analyzer --test chomp_chop_implicit_topic_tests`
- Verify idiomatic Perl IO pattern `while (<STDIN>) { chomp; print $_; }` produces no diagnostics
- Verify explicit `my $_ = ...; chomp;` does not produce UnusedVariable warning

## Unknowns
- The existing `is_builtin_global` and underscore-prefix exemptions already prevent false positives; this fix makes the behavior explicit as a defensive measure
- `alarm`, `exit`, `rand` were in the spec's builtin list but do NOT default to `$_` in Perl and were correctly omitted

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
@@ -658,6 +658,13 @@ impl ScopeAnalyzer {
                 // Some builtins consume declaration-capable filehandle arguments directly,
                 // e.g. `open my $fh, ...` or `pipe my $r, my $w;`. Those declarations should
                 // count as used and initialized by the builtin itself.
+                //
+                // Builtins that default to $_ when called with zero arguments implicitly
+                // read (and in some cases modify) $_. Mark it as used so that any lexically-
+                // scoped `my $_` in scope is not reported as unused or uninitialized.
+                if args.is_empty() && is_topic_defaulting_builtin(name) {
+                    let _ = scope.use_variable_parts("$", "_");
+                }
                 ancestors.push(node);
                 for arg in args {
                     self.analyze_node(arg, scope, ancestors, issues, context);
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary**: Builder completed fix(scope) for issue #3458 — added `is_topic_defaulting_builtin` helper and FunctionCall handler guard in `scope_analyzer.rs`, plus 11 regression tests. 1 PR created.

### Ending State

- PR created: https://github.com/EffortlessMetrics/perl-lsp/pull/3650
- Tests: 11 added (`chomp_chop_implicit_topic_tests.rs`)
- Files changed:
  - `crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs` (+38 lines)
  - `crates/perl-semantic-analyzer/tests/chomp_chop_implicit_topic_tests.rs` (+301 lines, new)
- Tests passing: 11/11 new tests pass; 75 pass in full suite (1 pre-existing unrelated failure in `symbol_extraction_method_preserves_attributes`)
- Next step: ready for review

### Retrospective

**What was harder than expected**: The worktree was an empty sparse overlay — `git worktree add` had not been run, so Read/Edit/Write tools silently appeared to succeed but wrote to non-existent paths. The fix was to delete the empty directory and run `git worktree add` properly. This cost several rounds of debugging.

**What surprised me**: All 11 tests passed _before_ the fix was applied. The `$_` variable is already exempt from scope diagnostics via `is_builtin_global` and the `_`-prefix skip in `for_each_reportable_unused_variable`. The fix is defensive — it makes semantic intent explicit and guards against future changes to those exemptions. TDD worked as intended by confirming the existing behavior is correct.

**What would I do differently**: Before writing any code, verify the worktree has a real `.git` file (not just an empty directory). A one-liner `ls .git` in the worktree root would catch this immediately.

**Context that would have helped**: Knowing that `git worktree list` is the authoritative check — if the path isn't in that list, tools cannot write to it despite appearing to succeed.

- Logical next: route to reviewer for standards + deep correctness pass
- Related issues: #3458 (this issue), any scope_analyzer false-positive issues
- Gotchas: `is_builtin_global("$", "_")` already exempts `$_` globally — the fix is additive/defensive, not the primary guard
- Traps: The `symbol_extraction_method_preserves_attributes` test in `comprehensive_unit_tests.rs` is pre-existing failing — do not treat as a regression from this PR
- Confidence for next: high

Issue #3620, opened 2026-04-10
gap(diagnostics): security lint coverage misses dangerous execution primitives


EffortlessSteven:
## Summary

LSP-side security lint coverage is much narrower than the repo's DAP dangerous-operation policy.

## Evidence

- `crates/perl-lsp-diagnostics/src/diagnostics.rs:148-149` describes security anti-pattern detection as only string eval, two-arg open, and backtick exec.
- `crates/perl-lsp-diagnostics/src/lints/security.rs:1-167` implements exactly those checks.
- `crates/perl-diagnostics-codes/src/lib.rs:159-163` only defines two PL6xx security codes here (`PL600`, `PL601`).
- Meanwhile `crates/perl-dap-eval/src/patterns.rs:8-105` already maintains a much broader dangerous-op deny-list including `system`, `exec`, `fork`, `kill`, `open`, `pipe`, `socket`, `connect`, `require`, `do`, `unlink`, `chmod`, `chroot`, and more.

## Why this is a gap

The repo already encodes a strong view of dangerous Perl operations for debug-eval safety, but ordinary LSP diagnostics only warn on a narrow subset. That leaves common high-risk constructs invisible in editor feedback even though the project clearly already has a vocabulary for many of them.

## Suggested scope

Start with a bounded expansion instead of trying to lint every risky builtin at once:

1. Add diagnostics for the highest-risk execution/process primitives: `system`, `exec`, pipe-open, `qx//` / `readpipe`.
2. Then consider filesystem and process-control primitives such as `unlink`, `chmod`, `fork`, `kill`, `chroot`.
3. Reuse the shared dangerous-op vocabulary where practical so DAP and LSP policy do not drift.

EffortlessSteven:
Related to Safe execution cluster:
- #3618: Safe execution context for complex analysis ops
- #3619: Safe execution wrapper for async ops

Consider consolidating these into a single epic for coordinated implementation.

EffortlessSteven:
🔗 **Cross-link to Security Keystone #3617**

This issue is part of the security linting block tracked in #3617 (Taint analysis implementation).

See #3617 for the coordinated security linting roadmap.

EffortlessSteven:
Fixed by merged PR #3694, which added security diagnostics for system/exec/pipe-open/readpipe.

PR #3694, opened 2026-04-10
feat(diagnostics): add security lints for system, exec, pipe-open, readpipe (#3620)

EffortlessSteven:
## Summary

Adds four new security diagnostic codes (PL603-PL606) for dangerous Perl execution primitives that were absent from LSP diagnostics despite being in the DAP dangerous-op deny-list.

- **PL603 SecuritySystemCall**: `system()` calls execute shell commands — warns to use list form `system($cmd, @args)` to avoid shell injection
- **PL604 SecurityExecCall**: `exec()` calls replace the current process with a shell command — same injection risk
- **PL605 SecurityPipeOpen**: pipe-open patterns (`open(FH, "|cmd")`, `open(FH, "|-", ...)`, `open(FH, "-|", ...)`) execute shell commands
- **PL606 SecurityReadpipe**: `readpipe("cmd")` function call form — equivalent to `qx//` (backticks already caught by PL601, but explicit function call was not)

All four checks fire through the existing `FunctionCall` branch of `walk_security_node()`.

Pipe-open detection correctly handles both forms the parser produces:
- Flat `args` in unit-test-constructed ASTs
- `ArrayLiteral`-wrapped args in real parser output (`open(FH, "|-", "ls")` becomes `FunctionCall { args: [ArrayLiteral { elements: [fh, "|-", "ls"] }] }`)

This is the bounded scope-1 expansion from issue #3620. Scope-2 items (filesystem/process-control primitives: `unlink`, `chmod`, `fork`, `kill`, `chroot`) are not addressed here.

## Test plan

- 10 new tests in `crates/perl-lsp-diagnostics/src/lints/security.rs`:
  - `system_call_is_flagged` — basic system() call
  - `system_call_list_form_is_flagged` — system() list form also flagged
  - `exec_call_is_flagged` — basic exec() call
  - `exec_call_list_form_is_flagged` — exec() list form also flagged
  - `pipe_write_open_is_flagged` — 3-arg write pipe `open(FH, "|-", "cmd")`
  - `pipe_read_open_is_flagged` — 3-arg read pipe `open(FH, "-|", "cmd")`
  - `two_arg_pipe_open_is_flagged_as_pipe_open` — 2-arg pipe `open(FH, "|cmd")`
  - `normal_three_arg_open_is_not_pipe_flagged` — `open(FH, "<", "file")` not flagged
  - `readpipe_call_is_flagged` — `readpipe("cmd")` function form

- `cargo test -p perl-lsp-diagnostics`: 53 passed, 0 failed
- `cargo test -p perl-diagnostics-codes`: all passed
- `cargo clippy -p perl-lsp-diagnostics -p perl-diagnostics-codes --lib`: clean
- `cargo fmt -p perl-lsp-diagnostics -p perl-diagnostics-codes`: applied

### DIFF
--- a/crates/perl-diagnostics-codes/src/lib.rs
+++ b/crates/perl-diagnostics-codes/src/lib.rs
@@ -175,6 +175,14 @@ pub enum DiagnosticCode {
     SecurityBacktickExec,
     /// Global assignment to `$SIG{__DIE__}` / `$SIG{__WARN__}`
     SecuritySignalHandler,
+    /// `system()` call executes shell commands
+    SecuritySystemCall,
+    /// `exec()` call replaces the current process with a shell command
+    SecurityExecCall,
+    /// Pipe-open `open(FH, "|-", ...)` / `open(FH, "-|", ...)` executes shell commands
+    SecurityPipeOpen,
+    /// `readpipe()` function call executes shell commands (equivalent to qx//)
+    SecurityReadpipe,
 
     // Import (PL700-PL799)
     /// Module appears to be unused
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
I found one correctness gap that should be fixed before merge. The new pipe-open detection only treats two-arg `open(FH, "|cmd")` as a shell-executing pipe, but Perl also supports the read-pipe form `open(FH, "cmd |")`. That pattern still executes a command, yet `is_pipe_two_arg_string()` only checks `starts_with('|')`, so the PR misses the trailing-pipe case and the new `PL605` coverage is incomplete. Please extend the two-arg detection to catch both write-pipe and read-pipe forms, and add a regression test for something like `open(FH, "ls |");` so the new diagnostic actually covers both 2-arg pipe-open directions.

Issue #3555, opened 2026-04-10
[toolchain] Add PERL5LIB environment variable support


EffortlessSteven:
## Problem

perl-lsp currently ignores the `PERL5LIB` environment variable, which is the standard way Perl developers customize their module search paths. This causes:

1. Modules visible to command-line Perl are not visible to the LSP
2. Local::Lib and other path-setting tools don't work out of the box
3. CI/CD environments that set PERL5LIB are not respected
4. Users must duplicate PERL5LIB entries in `.perl-lsp.toml` include_paths

## Current Behavior

```rust
// From perl-lsp-config/src/lib.rs
fn fetch_perl_inc() -> Vec<PathBuf> {
    let output = Command::new("perl").args(["-e", "print join(\"\\n\", @INC)"]).output();
    // ... only gets system @INC, not PERL5LIB-aware
}
```

- `PERL5LIB` is ignored when building effective @INC
- `WorkspaceConfig` only uses `include_paths` from config, not environment
- The `perl-dap-platform::setup_environment()` sets PERL5LIB but doesn't read it

## Expected Behavior

1. Read `PERL5LIB` environment variable and prepend to module search paths
2. Honor `PERL5LIB` in addition to configured `include_paths`
3. Provide option to disable PERL5LIB for reproducible builds
4. Document PERL5LIB interaction with `.perl-lsp.toml`

## Proposed Configuration

```toml
[perl]
# Honor PERL5LIB environment variable (default: true)
use_perl5lib = true

# Precedence: "prepend" (PERL5LIB first) or "append" (config paths first)
perl5lib_precedence = "prepend"
```

## Implementation Notes

- Parse `PERL5LIB` (colon-separated on Unix, semicolon on Windows)
- Use `perl-dap-platform::PATH_SEPARATOR` constant for cross-platform support
- Prepend PERL5LIB paths to the include_paths list when `use_perl5lib = true`
- Ensure paths are validated through `perl_path_security` before use

## PERL5LIB Format

```bash
# Unix
export PERL5LIB=/home/user/custom/lib:/opt/perl/lib

# Windows
set PERL5LIB=C:\Users\user\custom\lib;C:\Perl\lib
```

## Related Issues

- #3421 (Resolver foundation: model effective @INC separately from includePaths)
- #3447 (Module resolution fails for absolute paths and system @INC entries)
- #3493 ([workspace] File discovery includes built artifact directories)

## Acceptance Criteria

[...]

Issue #3447, opened 2026-04-10
Module resolution fails for absolute paths and system @INC entries


EffortlessSteven:
## Bug Description

The documentation states that `perl.workspace.includePaths` are "appended to the internal @INC" (see [CONFIG.md](https://github.com/EffortlessMetrics/perl-lsp/blob/master/docs/reference/CONFIG.md)), but the actual implementation treats all paths as **workspace-root-relative**.

This creates a mismatch between user expectations and actual behavior:
- Users cannot specify absolute paths to system Perl module directories
- System `@INC` paths (when `useSystemInc: true`) may not resolve correctly if they fall outside the workspace root
- External dependencies in non-workspace locations are not found

## Relevant Code

- `crates/perl-module-resolution-path/src/lib.rs` - `resolve_module_path()` builds all paths under `root`:
  ```rust
  let candidate = if base == "." {
      root.join(&relative_path)
  } else {
      root.join(base).join(&relative_path)  // Always workspace-relative
  };
  ```

- `crates/perl-lsp-config/src/lib.rs` - Configuration handling

- `crates/perl-lsp-diagnostics/src/lints/missing_module.rs` - Uses the resolver; has conservative core-module allowlist that kicks in when `use_system_inc` is false

- `docs/reference/CONFIG.md` - Documentation says paths are "appended to the internal @INC" which implies true @INC behavior, not workspace-relative

## Steps to Reproduce

1. Configure perl-lsp with absolute paths in `includePaths`:
   ```json
   {
     "perl": {
       "workspace": {
         "includePaths": ["/usr/local/lib/perl5", "/opt/perl/lib"]
       }
     }
   }
   ```

2. Create a Perl file that uses a module located in one of those absolute paths

3. Observe that the module is reported as not found (PL701) because the resolver looks for `{workspace}/usr/local/lib/perl5/...` instead of `/usr/local/lib/perl5/...`

## Expected Behavior

Either:
- The resolver should support absolute paths that are not joined to the workspace root, OR
- The documentation should clearly state that all paths are workspace-relative

## Suggested Fix

Split module resolution into two modes:
1. Workspace-relative paths (current behavior for relative paths)
2. External/system @INC paths (for absolute paths or when `useSystemInc: true`)

EffortlessSteven:
[...]

Issue #3421, opened 2026-04-10
Resolver foundation: model effective \@INC separately from includePaths


EffortlessSteven:
## Summary
The current `includePaths` / `useSystemInc` contract is intentionally narrower than real Perl `@INC`, and the merged docs fix in #3341 correctly documented that narrower behavior. The next step should not be another `includePaths` tweak. It should be a dedicated `@INC` / resolver subsystem track.

## Why this should be separate
- `includePaths` currently models workspace search roots, not effective Perl `@INC`
- `useSystemInc` is a separate opt-in probe and should stay separate
- file-local `use lib` support is still a conservative static approximation
- URI resolution currently layers editor-specific behavior on top of path lookup, which is useful UX but not itself an `@INC` model

## Proposed direction
Build a shared ordered `EffectiveInc` model and make every resolver consume it.

### Foundation
- add a shared ordered `EffectiveInc` / `IncRoot` representation
- turn the current implicit `root/lib` fallback into an explicit default root
- refactor both path and URI resolvers to consume ordered roots instead of bespoke fallback logic

### Interpreter-aware startup roots
- add workspace-level interpreter settings for module resolution (`perl.workspace.perlPath`, interpreter args, optional env)
- probe startup `@INC` from the selected interpreter instead of reimplementing Perl startup ordering in Rust
- cache probe results by interpreter path, args, env, and cwd

### Static file-local `@INC`
- replace the current full-document `use lib` text scan with an AST-based, position-aware collector
- support `use lib`, `no lib`, simple `FindBin` forms, and conservative literal `BEGIN { unshift @INC, ... }` patterns
- model `lib.pm` arch/version expansion in static mode where possible

### External absolute roots
- support external absolute roots explicitly as a first-class concept, separate from workspace-relative roots
- keep trust boundaries obvious rather than silently widening the current workspace-oriented setting

### Modes
- keep a safe static mode as default
- consider a separate trusted runtime mode later for projects that want probe-based or hook-aware fidelity

## Suggested implementation order
1. resolver foundation (`EffectiveInc`, ordered roots, interpreter config)
2. external absolute roots
3. AST-based static file-local `@INC`
4. unify consumers (missing-module lint, goto-definition, hover, URI/path resolvers)
5. optional trusted runtime mode for hook-driven/dynamic `@INC`

## Related
- #3337
- #3341

EffortlessSteven:
[...]

PR #3717, opened 2026-04-10
feat(config): add PERL5LIB environment variable support (#3555)

EffortlessSteven:
## Summary

- Add `Perl5LibPrecedence` enum (`prepend`/`append`) to `perl-lsp-config` with `#[serde(rename_all = "lowercase")]` for TOML and LSP settings support
- Add `use_perl5lib: bool` (default: `true`) and `perl5lib_precedence: Perl5LibPrecedence` fields to `WorkspaceConfig`; both are configurable via LSP `workspace.usePerl5lib` / `workspace.perl5libPrecedence` settings and `.perl-lsp.toml` `[perl]` section
- Add `WorkspaceConfig::parse_perl5lib` with platform-aware separator (`:` on Unix, `;` on Windows) that silently drops empty components
- Add `WorkspaceConfig::effective_include_paths` that merges PERL5LIB paths with `include_paths` respecting precedence and the `use_perl5lib` toggle
- Update all three module resolution entry points (`resolve_module_path`, `resolve_module_path_with_uri`, `resolve_module_to_path_with_doc`) to read `PERL5LIB` at resolution time and merge via `effective_include_paths`
- Add `use_perl5lib` and `perl5lib_precedence` optional fields to `ProjectPerlConfig` for `.perl-lsp.toml` support; `apply_to_workspace_config` propagates them
- Add 12 new integration tests covering: `parse_perl5lib` (empty, single, multi-path, empty-component dropping), `effective_include_paths` (no paths, prepend, append, disabled), TOML parsing of new fields, and `apply_to_workspace_config` propagation

## Test plan

- [ ] All existing `perl-lsp-config` tests pass (70 tests across 4 suites)
- [ ] 12 new PERL5LIB-specific tests pass (project_config suite grows from 14 to 26)
- [ ] `cargo clippy -p perl-lsp-config --tests` reports no issues
- [ ] `cargo fmt -p perl-lsp-config -p perl-lsp-rs` produces no changes
- [ ] CI gate on GitHub passes

### DIFF
--- a/crates/perl-lsp-config/src/lib.rs
+++ b/crates/perl-lsp-config/src/lib.rs
@@ -243,6 +243,21 @@ impl ServerConfig {
     }
 }
 
+/// Controls whether PERL5LIB paths are prepended or appended to `include_paths`.
+///
+/// `Prepend` (the default) mirrors Perl's own behaviour: paths earlier in the
+/// search order shadow later ones, so PERL5LIB paths take priority over any
+/// project-level `include_paths`.
+#[derive(Debug, Clone, PartialEq, Eq, Default, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Perl5LibPrecedence {
+    /// PERL5LIB entries are placed *before* `include_paths` (default).
+    #[default]
+    Prepend,
+    /// PERL5LIB entries are placed *after* `include_paths`.
+    Append,
+}
+
 /// Workspace configuration for module resolution
 ///
 /// Controls how the LSP server resolves module imports and finds
@@ -266,6 +281,14 @@ pub struct WorkspaceConfig {
     /// Resolution timeout in milliseconds
     /// Default: 50ms
     pub resolution_timeout_ms: u64,
+
+    /// Whether the `PERL5LIB` environment variable is read and merged into
+    /// the module search path.  Default: `true`.
+    pub use_perl5lib: bool,
+
+    /// Controls whether PERL5LIB entries come before or after `include_paths`.
+    /// Default: `Prepend` (mirrors Perl's own search order).
+    pub perl5lib_precedence: Perl5LibPrecedence,
 }
 
 impl Default for WorkspaceConfig {
@@ -275,11 +298,49 @@ impl Default for WorkspaceConfig {
             use_system_inc: false,
             system_inc_cache: None,
             resolution_timeout_ms: 50,
+            use_perl5lib: true,
+            perl5lib_precedence: Perl5LibPrecedence::Prepend,
         }
     }
 }
 
 impl WorkspaceConfig {
+    /// Parse a `PERL5LIB` environment variable value into a list of paths.
+    ///
+    /// Uses `:` as the separator on Unix and `;` on Windows, matching Perl's
+    /// own behaviour.  Empty components (produced by leading, trailing, or
+    /// consecutive separators) are silently dropped.
+    pub fn parse_perl5lib(value: &str) -> Vec<String> {
+        #[cfg(windows)]
+        const SEP: char = ';';
+        #[cfg(not(windows))]
+        const SEP: char = ':';
+        value.split(SEP).filter(|s| !s.is_empty()).map(|s| s.to_string()).collect()
+    }
+
+    /// Return the effective module-search-path, merging `PERL5LIB` paths with
+    /// `self.include_paths` according to `self.perl5lib_precedence`.
+    ///
+    /// If `self.use_perl5lib` is `false`, or `perl5lib_paths` is empty, the
+    /// returned list is identical to `self.include_paths`.
+    pub fn effective_include_paths(&self, perl5lib_paths: &[String]) -> Vec<String> {
+        if !self.use_perl5lib || perl5lib_paths.is_empty() {
+            return self.include_paths.clone();
+        }
+        match self.perl5lib_precedence {
+            Perl5LibPrecedence::Prepend => {
+                let mut result = perl5lib_paths.to_vec();
+                result.extend_from_slice(&self.include_paths);
+                result
+            }
+            Perl5LibPrecedence::Append => {
+                let mut result = self.include_paths.clone();
+                result.extend_from_slice(perl5lib_paths);
+                result
+            }
+        }
+    }
+
     /// Update workspace configuration from LSP settings.
     pub fn update_from_value(&mut self, settings: &serde_json::Value) {
         if let Some(workspace) = settings.get("workspace") {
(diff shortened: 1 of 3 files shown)


EffortlessSteven (commented):
**Deep correctness review complete.**

## What I verified

**parse_perl5lib**: Correct. Platform-aware separator (`:`/`;`) via `#[cfg(windows)]`. Empty component filtering via `.filter(|s| !s.is_empty())` is correct — handles leading/trailing/consecutive separators exactly as documented.

**Precedence logic**: Correct. `Prepend` builds `[perl5lib_paths..., include_paths...]`, `Append` builds `[include_paths..., perl5lib_paths...]`. The `use_perl5lib = false` short-circuit and empty-slice short-circuit both return `include_paths.clone()` unchanged. Logic is sound.

**Security**: PERL5LIB paths pass through `validate_workspace_path` in both `resolve_module_path` and `resolve_module_uri`. Null bytes and control characters are rejected. Path traversal (`..`) is caught.

**Wiring**: All three module resolution entry points (`resolve_module_path`, `resolve_module_path_with_uri`, `resolve_module_to_path_with_doc`) consistently read `PERL5LIB` from the environment at resolution time via `std::env::var("PERL5LIB")`.

## Logic bug fixed (PR #3753)

`update_from_value` used a wildcard `_` arm for `perl5libPrecedence`, which silently reset the setting to `Prepend` for any unrecognised string — including a client sending `"replace"` or a casing typo. With `Append` as the active setting, this could corrupt user config silently. Fixed to only match `"append"` and `"prepend"`, leaving any unrecognised value unchanged.

## Tests added (PR #3753, 7 tests)

The LSP `update_from_value` code path for `usePerl5lib` and `perl5libPrecedence` had zero test coverage. Added 5 tests covering all four recognised values plus the unknown-value preservation behaviour. Added 2 tests asserting full path-list order for `effective_include_paths` (existing tests only spot-checked `.first()`/`.last()`).

## Structural issue filed (#3741)

Absolute PERL5LIB paths (the primary real-world use case — `/home/user/perl5/lib/perl5`) are silently rejected by `validate_workspace_path` in both resolution functions. This is a design gap: PERL5LIB paths go through the workspace-folder loop which guards all paths against the workspace boundary, but absolute system paths are inherently outside the workspace. The system `@INC` code path handles this correctly (direct `inc_path.join(&relative_path)` without workspace guard) — PERL5LIB needs the same treatment. Filed as #3741 for a follow-up builder.

## Regression risk: LOW

The fix is conservative (widens the `_` arm to a no-op). No existing behaviour changed for valid values. The structural issue in #3741 is pre-existing and was not introduced by this PR.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-3717",
  "current_version": {
    "sha": "9aa2cad1edbf407f7ec062e8cc9c969d6fc03436",
    "updated_at": "2026-04-10T10:41:14Z"
  },
  "label_bindings": [
    {
      "label": "reviewed-deep",
      "bound_at_version": "9aa2cad1edbf407f7ec062e8cc9c969d6fc03436",
      "bound_at_timestamp": "2026-04-10T10:41:42Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Review Wrapup

**Summary:** Deep reviewer completed review of PR #3717 (PERL5LIB environment variable support). Found 1 logic bug (fixed in follow-up PR #3753), 1 structural gap (filed as issue #3741), 7 tests added. `reviewed-deep` label set.

---

### Ending State

- PR reviewed: https://github.com/EffortlessMetrics/perl-lsp/pull/3717 (already merged)
- Follow-up PR: https://github.com/EffortlessMetrics/perl-lsp/pull/3753
- Follow-up issue: https://github.com/EffortlessMetrics/perl-lsp/issues/3741
- Changes made: 1 logic bug fixed, 7 tests added (5 for `update_from_value` LSP path, 2 for `effective_include_paths` full-order assertions)
- Test artifacts: `crates/perl-lsp-config/src/lib.rs`, `crates/perl-lsp-config/tests/project_config.rs`
- Status: approved with improvements
- Next step: PR #3753 ready for CI and merge; issue #3741 needs a builder

### What was verified

1. **`parse_perl5lib`**: Correct. Platform-aware separator. Empty component filtering. Handles leading/trailing/consecutive separators.
2. **`effective_include_paths`**: Correct. Both `Prepend` and `Append` arms build the right order. Short-circuits for `use_perl5lib=false` and empty `perl5lib_paths` both correct.
3. **Wiring**: All three resolution entry points consistently read `PERL5LIB` from environment and call `effective_include_paths`. No code path missed.
4. **Security**: PERL5LIB paths pass through `validate_workspace_path` — null bytes, control chars, and path traversal all caught.

### Logic bug fixed (PR #3753)

`update_from_value` used `_ => Perl5LibPrecedence::Prepend` as the fallback for any unrecognised `perl5libPrecedence` string. This silently overwrote an explicitly-set `Append` value whenever the LSP client sent an unknown string. Fixed to match only `"append"` and `"prepend"`, leaving any other value unchanged.

### Tests added (PR #3753)

The entire LSP `update_from_value` path for `usePerl5lib` and `perl5libPrecedence` was untested. Added:
- `workspace_config_update_from_value_sets_use_perl5lib_false`
- `workspace_config_update_from_value_sets_use_perl5lib_true`
- `workspace_config_update_from_value_sets_perl5lib_precedence_append`
- `workspace_config_update_from_value_sets_perl5lib_precedence_prepend`
- `workspace_config_update_from_value_unknown_precedence_leaves_current_value` (the regression test for the fixed bug)
- `effective_include_paths_prepend_preserves_full_order`
- `effective_include_paths_append_preserves_full_order`

### Structural issue filed (#3741)

Absolute PERL5LIB paths (the primary real-world use case) are silently rejected by `validate_workspace_path` in both `resolve_module_path` and `resolve_module_uri`. Both functions use `root.join(include_path)` — when `include_path` is absolute, Rust's `Path::join` returns the absolute path directly, then `validate_workspace_path` rejects it because it's outside the workspace. The system `@INC` code path correctly handles this with a direct join without the workspace guard. PERL5LIB needs the same treatment.

---

### Retrospective

**Harder than expected:** The linter process was actively reverting file edits in the main repo checkout, making it impossible to stage and commit changes there. Discovering I was in a shared worktree (`agent-ae8ca219`) and needed to create a fresh branch from `origin/master` in that isolated worktree context consumed significant iteration. Next time: immediately identify the worktree path and check `git branch --show-current` before any edits.

**What surprised me:** The PR correctly wires `effective_include_paths` into all three resolution code paths, and the config parsing and precedence logic are all sound — but the underlying `resolve_module_path` function rejects absolute paths via `validate_workspace_path` before PERL5LIB was even added. So the feature as described in the issue (supporting `/home/user/perl5/lib/perl5`) was architecturally impossible before the resolution layer is updated. This is a silent no-op rather than a visible error, which makes it hard to discover without reading the full resolution stack.

**What would have helped:** Knowing that `perl-module-resolution-path` and `perl-module-resolution-uri` are separate microcrates with their own security boundary logic would have directed attention there immediately. The issue spec said "validate through `perl_path_security`" but didn't explain that `perl_path_security` enforces workspace containment, which conflicts with the actual PERL5LIB use case.

---

- **Logical next:** Builder for issue #3741 — add absolute-path bypass in `perl-module-resolution-path/src/lib.rs` and `perl-module-resolution-uri/src/lib.rs`, similar to the `use_system_inc` code path in `resolve_module_uri`
- **Related issues:** #3421 (resolver foundation), #3447 (module resolution absolute paths), #3555 (original PERL5LIB issue)
- **Gotcha:** `resolve_module_path` always returns `Some(root.join("lib").join(relative_path))` as a fallback even when the module doesn't exist — tests that assert `is_some()` prove nothing without also checking `exists()`
- **Gotcha:** PR #3753 should be cherry-clean — it only touches `perl-lsp-config`, no resolution layer changes
- **Confidence for next:** High on the bug location, medium on the best fix shape (separate pass vs. relaxed guard)

Issue #3605, opened 2026-04-10
feat: Role composition diagnostics


EffortlessSteven:
## Summary

When a class consumes multiple roles that provide the same method, Moose/Moo should detect and report the conflict. Currently, perl-lsp does not perform this analysis.

## Current State

Given:
```perl
package RoleA;
use Moose::Role;
sub shared_method { 'A' }

package RoleB;
use Moose::Role;
sub shared_method { 'B' }

package MyClass;
use Moose;
with 'RoleA', 'RoleB';  # Conflict: both provide shared_method
```

**Expected:** Diagnostic warning about method conflict from roles.

**Actual:** No conflict detection; roles stored as strings only.

## Proposed Implementation

1. **Role method indexing**:
   - Parse role files to extract their methods
   - Store in workspace-wide index

2. **Conflict detection**:
   - When class has `with 'RoleA', 'RoleB'`
   - Check for method name overlaps between roles
   - Report diagnostic if conflict exists without resolution

3. **Resolution hints**:
   - Suggest `RoleA->shared_method` exclusion syntax
   - Detect explicit method in class (which resolves conflict)

## Acceptance Criteria

- [ ] Detect method conflicts between consumed roles
- [ ] Report diagnostic at `with` statement
- [ ] Support both Moose::Role and Role::Tiny
- [ ] Don't false-positive when class provides its own method
- [ ] Tests for conflict scenarios

## Related Code

- `crates/perl-semantic-analyzer/src/analysis/class_model.rs`
  - `ClassModel.roles: Vec<String>` — need to resolve to role definitions
  - New: `roles/index.rs` for workspace role indexing

## Open Questions

- Should we parse role files eagerly or on-demand?
- How to handle role files outside workspace?

EffortlessSteven:
Merged same-file MVP in #3719. This lands Moo/Moose role conflict diagnostics for same-file role composition, anchored at the  site, while leaving broader workspace-indexed and non-Moo/Moose coverage for follow-up work.

EffortlessSteven:
Correction to my last note: the diagnostic is anchored at the plain text `with` site in the role-composition statement. Same-file Moo/Moose MVP is merged in #3719; broader workspace and framework coverage remains follow-up scope.

EffortlessSteven:
# Role Composition Diagnostics Research Summary

## Issue Overview
[...]

PR #3719, opened 2026-04-10
feat(diagnostics): same-file Moo/Moose role conflicts (#3605)

EffortlessSteven:
## Summary
- add a same-file role-conflict lint for Moo/Moose role composition
- anchor diagnostics at existing `with` role references from the symbol table
- ignore `requires` declarations and suppress the warning when the consuming class defines the method itself

## Testing
- cargo test -p perl-diagnostics-codes --test comprehensive_unit_tests -- --nocapture
- cargo test -p perl-lsp-diagnostics --test role_conflicts_tests -- --nocapture
- cargo test -p perl-lsp-diagnostics --lib -- --nocapture

This MVP is intentionally same-file only. It does not attempt workspace-wide role indexing or `Role::Tiny` support.

### DIFF
--- a/crates/perl-diagnostics-codes/src/lib.rs
+++ b/crates/perl-diagnostics-codes/src/lib.rs
@@ -135,6 +135,8 @@ pub enum DiagnosticCode {
     DuplicateSubroutine,
     /// Missing explicit return statement
     MissingReturn,
+    /// Same-file Moo/Moose roles provide conflicting methods
+    RoleConflict,
     /// Invalid character(s) in a subroutine prototype
     ///
     /// Perl only allows `$`, `@`, `%`, `&`, `*`, `\`, `;`, `+`, `_`, and
@@ -248,6 +250,7 @@ impl DiagnosticCode {
             DiagnosticCode::DuplicateSubroutine => "PL300",
             DiagnosticCode::MissingReturn => "PL301",
             DiagnosticCode::InvalidPrototype => "PL302",
+            DiagnosticCode::RoleConflict => "PL303",
             DiagnosticCode::BarewordFilehandle => "PL400",
             DiagnosticCode::TwoArgOpen => "PL401",
             DiagnosticCode::ImplicitReturn => "PL402",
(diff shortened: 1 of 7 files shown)


EffortlessSteven:
**Deep Review Complete** — `reviewed-deep` label set.

**What I verified:**

1. **Wire-up is correct.** `check_role_conflicts` is called from `diagnostics.rs` after `SymbolExtractor` builds the symbol table. The `role_conflicts` module is properly declared in `mod.rs` and re-exported from `lib.rs`.

2. **Detection logic is correct.** `ClassModelBuilder::new().build(node)` collects ALL package models first, then iterates class models with all role models already populated. Class-before-role ordering in the source file does not cause missed detections.

3. **False-positive suppression is correct.** The consuming class's own methods (including synthetic accessors from `has`) are collected and gate each conflict. A class-defined method suppresses the diagnostic. Roles composing other roles go into `role_models` (not `class_models`) and never trigger PL303.

4. **`requires` handled correctly.** The test `requires_does_not_count_as_a_provided_method` verifies `requires 'shared'` does NOT contribute to `provided_method_names`. Passes.

5. **`role_anchor_location` silent-skip is safe.** Both forms of `with` register `SymbolKind::Role` references in the symbol extractor, so location always resolves for syntactically valid code.

6. **`adjusts` field in `provided_method_names` is harmless.** `adjusts` is only populated for `Framework::ObjectPad` packages, which are never classified as `SymbolKind::Role`. Dead weight, not a bug.

7. **Coding standards clean.** No `unwrap()`, `expect()`, `panic!()`. All error paths use `Option`/`?`/`continue`.

**Improvements pushed in #3733:**

Three edge cases untested by the original PR, all confirmed correct:
- Multiple separate `with` calls accumulate and trigger conflict detection
- Three-way conflict emits a single PL303, not one per pair
- A composite role composing conflicting roles does not trigger PL303 itself

7 tests pass. Clippy clean. Formatting correct.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-3719",
  "current_version": {
    "sha": "a32ef44d22b69a9fc66a6013f02647aa9851fb21",
    "updated_at": "2026-04-10T09:44:11Z"
  },
  "label_bindings": [
    {
      "label": "reviewed-deep",
      "bound_at_version": "a32ef44d22b69a9fc66a6013f02647aa9851fb21",
      "bound_at_timestamp": "2026-04-10T09:45:02Z",
      "valid": true
    }
  ]
}
```

Issue #3478, opened 2026-04-10
[module-resolution] use lib paths not fully integrated with module resolution


EffortlessSteven:
## Problem

perl-lsp has code to extract `use lib` paths and FindBin patterns, but these are not fully integrated into module resolution. Developers cannot:
1. Reliably resolve modules in project-specific lib directories
2. Use FindBin for dynamic library paths
3. Stack multiple use lib statements

## Current Behavior

Code exists in `perl-module-resolution/src/use_lib.rs` to extract paths:
- `use lib 'lib';`
- `use lib "::Bin/../lib";`
- `use lib qw(lib t/lib);`

But:
1. **Path resolution is incomplete** — Paths must be relative to workspace root
2. **Absolute paths fail** — Outside workspace root, paths are ignored
3. **Document-relative resolution missing** — FindBin relative paths don't work properly when script is in subdirectory
4. **No per-file use lib tracking** — use lib in one file doesn't affect imports in another
5. **Integration gap** — LSP diagnostic (PL701 missing module) doesn't use resolved use lib paths

## Example That Fails

Project structure:
```
myproject/
  lib/
    MyApp.pm
  bin/
    script.pl
  Makefile.PL
```

script.pl:
```perl
#!/usr/bin/env perl
use FindBin;
use lib "::Bin/../lib";  # Add ../lib to @INC
use MyApp;  # Should find lib/MyApp.pm
```

Currently:
- `use lib "::Bin/../lib"` is parsed (extraction works)
- But when resolving `use MyApp`, the path is not used
- PL701 "Module not found" is reported even though lib/MyApp.pm exists

## Evidence

Code locations:
- `crates/perl-module-resolution/src/use_lib.rs` — Path extraction logic (works)
- `crates/perl-lsp-diagnostics/src/lints/missing_module.rs` — PL701 lint (does NOT use lib paths)
- `crates/perl-lsp-navigation/src/` — Go-to-definition (does NOT use lib paths)
- `crates/perl-lsp-completion/src/` — Module completion (does NOT use lib paths)

## Why This Matters

1. **Standard Perl practice** — Every project uses use lib or Makefile.PL
2. **LocalLib pattern** — local::lib relies on use lib being recognized
3. **Scope diagnostics wrong** — Reports false "module not found" errors
4. **Completion broken** — Can't complete module names that rely on use lib
5. **Navigation broken** — Can't go-to-definition on modules in lib directories

### Phase 1: Use lib path tracking per file
[...]

PR #3720, opened 2026-04-10
feat(module-resolution): unify @inc resolution behavior (#3478)

EffortlessSteven:
### Motivation

- Current resolution logic was workspace-first and treated absolute `includePaths` and `use lib` semantics inconsistently, causing modules to be missed or diagnostics to disagree with navigation/completion.
- The goal is a single ordered `EffectiveInc`-style model so every consumer (diagnostics/navigation/hover/completion) sees the same include stack and so absolute roots and per-file `use lib` overlays behave like users expect.
- Interpreter startup `@INC` probing must be workspace-aware so `PERL5LIB`/`-I`/project interpreter choices can be reflected when `useSystemInc` is enabled.

### Description

- Teach the path resolver to honor absolute `includePaths` as literal external roots and to return `None` when no file is found instead of using a hidden `root/lib` fallback, implemented in `crates/perl-module-resolution-path/src/lib.rs` by probing absolute bases literally and validating only workspace-scoped entries.
- Make URI resolution build a unified ordered include-entry list (`IncEntry`) that preserves workspace-scoped validation for relative entries and probes absolute entries literally, implemented in `crates/perl-module-resolution-uri/src/lib.rs`.
- Replace the simple `use lib` extractor with a delta model (`UseLibDelta` / `UseLibDeltaKind`) that supports `use lib` and `no lib`, `FindBin` handling, and an `apply_use_lib_deltas` helper that mutates per-call include lists; changes in `crates/perl-module-resolution/src/use_lib.rs` implement this and keep FindBin-derived paths workspace-scoped for safety.
- Wire per-document lexical include overlays and multi-root ownership into the LSP resolver by adding `apply_doc_use_lib_deltas` and `workspace_root_for_doc` and by switching resolution calls to pass `doc_text`/`doc_uri` where available, in `crates/perl-lsp/src/runtime/lifecycle/module_resolution.rs` and `crates/perl-lsp/src/runtime/diagnostics.rs` so diagnostics use the same effective include stack as navigation/completion/hover.
- Extend workspace configuration with interpreter probe settings (`perlPath`, `perlArgs`, `perlEnv`) and add cached, invalidation-aware startup `@INC` probing used when `useSystemInc` is enabled, implemented in `crates/perl-lsp-config/src/lib.rs`.
- Update and add tests to reflect explicit include-list semantics (resolution returns `None` when not found), traversal/security expectations, and absolute external-root coverage across the `perl-module-resolution` test suites.

### Testing

- Ran formatting with `cargo fmt --all` and then targeted/unit tests including `cargo test -p perl-module-resolution` which succeeded for the modified crate.
- Ran the LSP runtime module-resolution tests with `cargo test -p perl-lsp-rs --lib runtime::lifecycle::module_resolution::tests::` and the focused suite passed.
- Ran an integration subset `cargo test -p perl-module-resolution-path -p perl-module-resolution-uri -p perl-module-resolution -p perl-lsp-config -p perl-lsp-rs --lib runtime::lifecycle::module_resolution::tests::` and all executed tests passed.
- The updated `perl-module-resolution` test suites (`module_resolution_comprehensive`, `module_resolution_edge_cases`, `module_resolution_use_require`, `module_resolution_goto_definition`) were exercised and the repository test runs for the modified packages completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d8ba4cad148333a9e93dc0713094f1)

### DIFF
--- a/crates/perl-lsp-config/src/lib.rs
+++ b/crates/perl-lsp-config/src/lib.rs
@@ -4,6 +4,7 @@
 //! This microcrate isolates configuration parsing and defaults from the main
 //! server crate so they can evolve independently and be reused by tooling.
 
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 #[cfg(not(target_arch = "wasm32"))]
 use std::process::Command;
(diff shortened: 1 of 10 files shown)


EffortlessSteven:
Superseded by merged #3723 for the resolver/overlay work and by #3729 for the remaining interpreter-aware probe settings.

PR #3742, opened 2026-04-10
perf(code-actions): bound AST traversal to requested range (#3471)

### LINKED ISSUE #3471: [code-actions] Responsiveness degradation on large files due to unbounded AST traversal
## Problem

Code action responsiveness is degraded due to missing caching and inefficient AST traversal:

1. **No memoization of AST lookups**: The enhanced code actions provider recursively walks the AST for every mouse hover, even for identical positions.

2. **Duplicate work**: Both `CodeActionsProvider` and `EnhancedCodeActionsProvider` perform similar AST analysis on every request.

3. **No index caching**: Workspace symbol lookups during import management iterate the entire workspace index on each request.

4. **Unbounded extraction suggestions**: Extract variable/subroutine actions scan the entire function scope, even if the cursor is in a tiny 2-line block.

## Current Behavior

When a user opens a file with code actions enabled:
- Each mouse movement triggers code action collection
- AST is walked multiple times for overlapping ranges
- Extract refactoring suggestions take >100ms on large functions
- Users experience editor lag

## Expected Behavior

Code action collection should complete in <50ms even on large files (>5000 lines) with large functions (>500 lines).

## Measurement

Add tracing to:
- `EnhancedCodeActionsProvider::get_enhanced_refactoring_actions` — should be <30ms
- `extract_variable` — should be <20ms
- `extract_subroutine` — should be <25ms
- `import_management` — should be <15ms

## Files Affected

- `crates/perl-lsp-code-actions/src/enhanced/mod.rs` — add AST walk caching
- `crates/perl-lsp-code-actions/src/enhanced/helpers.rs` — add memoization
- `crates/perl-lsp-code-actions/src/code_actions.rs` — cache AST analysis results
- Add tracing/benchmarks: `crates/perl-lsp-code-actions/benches/code_actions.rs`

## Test Case

1. Create a 5000-line Perl file with 10 large functions
2. Enable code actions in VS Code
3. Move cursor through the file, collecting code actions at each line
4. Measure response time with chrome devtools or LSP client timing

## Priority

Medium-High — Code action latency directly impacts user experience. LSP clients expect <100ms response on all requests. Anything longer causes perceived editor "lag."

EffortlessSteven:
## Summary

- `collect_actions_for_range` in `EnhancedCodeActionsProvider` now returns immediately when a node's span does not overlap the requested range
- Because AST children are always contained within their parent's span, a non-overlapping parent prunes the entire subtree in one check
- Traversal cost changes from O(total AST nodes) to O(nodes in range), eliminating the responsiveness regression on large files (>5000 lines)

## Root cause

The previous implementation wrapped action-collection code in `if node.location.start <= range.1 && node.location.end >= range.0 { ... }` but always recursed into children regardless. For a 5000-line file with 500-line functions, a cursor at line 10 would still traverse all 5000 lines worth of AST nodes.

## Fix

Add an early-return guard at the top of `collect_actions_for_range`:

```rust
if node.location.end < range.0 || node.location.start > range.1 {
    return;
}
```

This is safe because the AST invariant guarantees children are contained within their parent's location span. The redundant `if` wrapper on action collection is removed.

## Test plan

- [x] `range_bounded_traversal_excludes_out_of_range_nodes` — two-block source, range covers only first block; verifies no edit from the second block appears
- [x] `large_file_narrow_range_returns_only_in_range_actions` — 50 if-blocks, range covers only the first; verifies correctness and that the first block's postfix action is still returned
- [x] `traversal_skips_nodes_entirely_before_range` — three statements, range covers only the third; verifies extract-variable actions reference only the selected expression
- [x] All 222 existing tests pass (52 + 41 + 69 + 14 + 8 + 15 + 23)
- [x] `cargo clippy` clean
- [x] `cargo fmt` applied

Fixes #3471


EffortlessSteven:
Superseded by clean replacement #3744. This branch carries unrelated stacked commits and is no longer the reviewable path for #3471.

Issue #3388, opened 2026-04-10
No support for Attribute::Handlers custom attribute patterns


EffortlessSteven:
Gap: No support for Attribute::Handlers custom attribute patterns.

Expected:
- Custom attributes like :MyAttr(foo) should have semantic meaning
- Attribute validation should work for user-defined handlers
- Attribute::Handlers use patterns should be recognized

Current: Attribute::Handlers is listed as a core module but has no semantic analysis support. User-defined attributes are stored but not validated or processed.

crates/perl-lsp-diagnostics/src/lints/missing_module.rs (listed as core module)

Example Perl pattern:
    use Attribute::Handlers;
    sub MyAttr :ATTR(CODE) { ... }
    sub foo :MyAttr { ... }

EffortlessSteven:
Opened clean replacement PR #3776. This keeps the slice parser-only: `use Attribute::Handlers;` enables registering `sub Name :ATTR(...)` handlers, and later `:MyAttr(...)` uses no longer warn in the same file.

EffortlessSteven:
Fixed by #3776.

PR #3774, opened 2026-04-10
fix(parser): support Attribute::Handlers custom attrs (#3388)

EffortlessSteven:
## Summary
- track `use Attribute::Handlers;` in the parser and register `sub Name :ATTR(...)` handlers for later attribute validation
- treat later `:MyAttr(...)` uses as known when that handler was declared in the file
- add a regression covering the narrow `Attribute::Handlers` happy path

## Tests
- cargo fmt --all --check
- cargo test -p perl-parser-core --test fix_attribute_validation_3357 -- --nocapture
- git diff --check

## Scope
This is the parser slice only for #3388. It does not add broader semantic processing for custom attribute handlers beyond suppressing the false warning path.


EffortlessSteven:
Superseded by clean rebased PR #3776.

PR #3789, opened 2026-04-10
fix(module-resolution): honor lexical use lib reinsertion precedence (#3720)

EffortlessSteven:
### Motivation

- Fix a correctness bug where repeated `use lib` of an already-active path was ignored instead of moving that path back to highest precedence. 
- Bring lexical `use lib` behavior in the include-stack implementation in line with Perl semantics so module resolution and navigation are deterministic.

### Description

- Update `resolve_use_lib_paths_from_source` in `crates/perl-module-resolution/src/use_lib.rs` to treat `use lib` as a highest-precedence-first stack by reinserting re-used paths at the front. 
- Preserve `no lib` semantics by removing matching paths from the active stack when encountered. 
- Iterate `use lib` items back-to-front to model unshift semantics and explicitly remove any existing occurrence before inserting at index 0. 
- Add focused regression tests (`latest_use_lib_takes_precedence`, `repeated_use_lib_reinserts_at_front`, `no_lib_then_use_lib_reintroduces_at_front`) to validate precedence and reinsertion behavior.

### Testing

- Ran `cargo fmt --all` to format the workspace and it completed successfully. 
- Ran `cargo test -p perl-module-resolution` and all unit and integration tests in that crate passed. 
- Ran `cargo test -p perllsp module_resolution -- --nocapture` (workspace module-resolution suites) and the module-resolution test suites completed successfully.

------

### DIFF
--- a/crates/perl-module-resolution/src/use_lib.rs
+++ b/crates/perl-module-resolution/src/use_lib.rs
@@ -149,14 +149,18 @@ pub fn resolve_use_lib_paths_from_source(
     workspace_root: &Path,
     file_dir: Option<&Path>,
 ) -> Vec<String> {
+    // Highest-precedence path first, matching lexical `use lib` behavior.
     let mut resolved = Vec::new();
     for op in extract_use_lib_operations(source) {
         match op {
             UseLibAction::Add(paths) => {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #3790, opened 2026-04-10
fix(module-resolution): restore lexical use lib precedence (#3720)

EffortlessSteven:
### Motivation

- Correct a correctness bug where repeated `use lib` reinsertion did not move an already-active path back to highest precedence, causing resolution order to diverge from Perl semantics.
- Ensure lexical `use lib` / `no lib` overlays behave deterministically so module resolution, diagnostics, and navigation can rely on a stable include-stack ordering.

### Description

- Change `resolve_use_lib_paths_from_source` in `crates/perl-module-resolution/src/use_lib.rs` to apply add operations as front-of-stack inserts while deduplicating existing entries, restoring Perl-like precedence semantics.
- Preserve `no lib` semantics by removing matching entries while allowing subsequent `use lib` to reintroduce the path at the front.
- Add regression tests in `crates/perl-module-resolution/src/use_lib.rs` covering later `use lib` precedence, reinsertion moving a path to the front, and `no lib` followed by `use lib` reintroducing the path.

### Testing

- Ran `cargo fmt --all` to format code successfully.
- Ran `cargo test -p perl-module-resolution use_lib::tests::` and all relevant unit tests passed (31 passed, 0 failed).
- Built and ran tests for the `perllsp` package as part of the workspace exercise which completed without failures (no failing tests observed).

------

### DIFF
--- a/crates/perl-module-resolution/src/use_lib.rs
+++ b/crates/perl-module-resolution/src/use_lib.rs
@@ -153,10 +153,10 @@ pub fn resolve_use_lib_paths_from_source(
     for op in extract_use_lib_operations(source) {
         match op {
             UseLibAction::Add(paths) => {
-                for path in resolve_use_lib_paths(&paths, workspace_root, file_dir) {
-                    if !resolved.contains(&path) {
-                        resolved.push(path);
-                    }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #3792, opened 2026-04-10
fix(module-resolution): restore repeated use lib precedence (#3720)

EffortlessSteven:
### Motivation
- Fix a correctness bug where repeated `use lib` additions did not move an already-active path back to highest precedence, causing resolution order to diverge from Perl semantics.
- Ensure the LSP’s per-document lexical overlay reflects the same reinsertion semantics so navigation and diagnostics observe consistent include ordering.

### Description
- Change lexical application in `resolve_use_lib_paths_from_source` to reinsert re-added paths at the front of the effective stack by removing any existing entry and `insert(0, path)` in statement order.
- Update `prepend_use_lib_paths` in the LSP resolver to remove any existing include before inserting it at the front so document-scoped overlays mirror core behavior.
- Add a unit regression test `repeated_use_lib_reinserts_to_front` in `crates/perl-module-resolution` validating `use lib 'a'; use lib 'b'; use lib 'a';` yields `['a','b']`.
- Add an LSP-level integration test `test_resolve_module_path_repeated_use_lib_reinserts_to_front` proving the overlayed include ordering makes resolution pick the reinserted path.

### Testing
- Ran `cargo fmt --all` to format changes and it completed successfully.
- Ran `cargo test -p perl-module-resolution repeated_use_lib_reinserts_to_front` and the new core `use_lib` unit test passed.
- Ran `cargo check -p perl-lsp-rs --lib` to validate the LSP crate builds cleanly and the check succeeded.

------

### DIFF
--- a/crates/perl-lsp/src/runtime/lifecycle/module_resolution.rs
+++ b/crates/perl-lsp/src/runtime/lifecycle/module_resolution.rs
@@ -512,6 +512,32 @@ mod tests {
     }
 
     #[test]
+    fn test_resolve_module_path_use_lib_reorders_existing_include_path() -> TestResult {
+        let temp = tempfile::tempdir()?;
+        let workspace = temp.path().join("workspace");
+        let preferred = workspace.join("a").join("Prefer").join("Me.pm");
+        let fallback = workspace.join("b").join("Prefer").join("Me.pm");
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #3831, opened 2026-04-10
Treat container-sigil derefs as uses of scalar-declared refs (fix #3445)

### LINKED ISSUE #3445: False positive unused variable for arrayref dereference pattern
## Bug Summary

The scope analyzer uses sigil-indexed buckets (`$`, `@`, `%`, etc.) to track variable declarations and uses. When a variable is declared as a scalar reference (`my $arrayref`) but later dereferenced with a different sigil (`push @$arrayref, ...`), the use is recorded in a different bucket than the declaration. This causes a false positive "Unused variable" (PL103) diagnostic.

## Steps to Reproduce

1. Create a Perl file with an array reference declared as a scalar, then dereferenced with `@`:

```perl
use strict;
use warnings;

my $arrayref = [];
push @$arrayref, 'item';  # Uses @$arrayref, not $arrayref
print $arrayref->[0];     # Also uses $arrayref with scalar sigil
```

2. Open in an editor with perl-lsp enabled
3. Observe PL103 "Variable '$arrayref' is declared but never used" on line 4, even though the variable is used on line 5

## Expected Behavior

Using a variable with a different sigil (e.g., `@$arrayref` for a `$arrayref` declaration) should mark the underlying variable as used, not produce an unused variable warning.

## Actual Behavior

The `use_variable_parts` function in `scope_analyzer.rs` looks up variables by their exact sigil:

```rust
let (variable_used, is_initialized) = scope.use_variable_parts(sigil, name);
```

When `@$arrayref` is encountered, it looks up in the `@` bucket (index 1), but the declaration `my $arrayref` was stored in the `$` bucket (index 0). The use is not recorded against the original declaration.

## Related Code

- `crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs`:
  - `sigil_to_index()` function (around lines 80-93) - Maps sigils to bucket indices
  - `use_variable_parts()` method - Looks up by exact sigil match
  - Variable declaration storage in `Scope.variables` - Uses sigil-indexed array
- `crates/perl-lsp-diagnostics/src/scope.rs` - Converts `IssueKind::UnusedVariable` to PL103

## Suggested Fix

1. Add dereference-aware tracking: when `@$var`, `%$var`, or similar is used, also mark the corresponding `$var` as used
2. Or unify reference variables across sigil buckets (store by name only, track sigil as metadata)
3. Handle common patterns: `@$arrayref`, `%$hashref`, `$$scalarref`, `@$arrayref[...]`, `%$hashref{...}`

## Additional Context

This is a known complexity in Perl analysis because sigil polymorphism is idiomatic. Common patterns affected:
- `my $ref = []; push @$ref, ...`
- `my $ref = {}; keys %$ref`
- `my $ref = \\$var; $$ref`
- `my $obj = MyClass->new; $obj->method` (object methods work fine, but builtin operations on references fail)

EffortlessSteven:
### Motivation
- Scalar references accessed with a container sigil (e.g. `@$ref`, `%$ref`) or via arrow syntax were not always being recognized as uses of the underlying scalar lexical, causing spurious `UnusedVariable`/`UndeclaredVariable` issues.

### Description
- Add a fallback lookup in `Scope::use_variable_parts` so that when the sigil is `@` or `%` the analyzer will fall back to the scalar sigil index (`$`) if no container declaration is found, and mark that variable as used and return its initialization state.
- Implement the fallback check using `fallback_idx` and a conditional lookup against the scalar map when the initial container-sigil map does not contain the name.
- Add a regression test `issue_3445_arrayref_deref_and_arrow_access_not_unused` in `tests/scope_and_symbol_tests.rs` that verifies `my $arrayref` used via `push @$arrayref` and `$arrayref->[0]` is not reported as unused or undeclared.

### Testing
- Ran the crate unit tests with `cargo test -p perl-semantic-analyzer` and the test suite including the new regression test passed.
- Existing dereference coverage (e.g. the `issue_3338` tests) also passed unchanged.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d928ca3bf88333beb0b66c2d27808b)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
@@ -222,6 +222,8 @@ impl Scope {
 
     fn use_variable_parts(&self, sigil: &str, name: &str) -> (bool, bool) {
         let idx = sigil_to_index(sigil);
+        let fallback_idx =
+            if sigil == "@" || sigil == "%" { Some(sigil_to_index("$")) } else { None };
         let mut current_scope = self;
 
         loop {
@@ -232,6 +234,17 @@ impl Scope {
                         *var.is_used.borrow_mut() = true;
                         return (true, *var.is_initialized.borrow());
                     }
+                }
+
+                // Perl dereference syntax can present a scalar-declared reference with
+                // a container sigil (e.g. `@$ref`, `%$ref`). If no lexical declaration
+                // exists for the container sigil, fall back to the scalar declaration.
+                if let Some(fallback_idx) = fallback_idx
+                    && let Some(map) = &vars[fallback_idx]
+                    && let Some(var) = map.get(name)
+                {
+                    *var.is_used.borrow_mut() = true;
+                    return (true, *var.is_initialized.borrow());
                 }
             }
 
--- a/crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs
+++ b/crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs
@@ -1348,6 +1348,38 @@ push @$arrayref, 'second';
     Ok(())
 }
 
+/// Regression test for issue #3445: scalar-declared array references used through
+/// both explicit `@$ref` and arrow access must count as uses of the same lexical.
+#[test]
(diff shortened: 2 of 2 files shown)


EffortlessSteven:
Superseded by the merged #3824 follow-up on #3445. Closing this older overlapping branch to keep the history focused.

Issue #3521, opened 2026-04-10
[workspace] Virtual file system (VFS) support not implemented


EffortlessSteven:
# [workspace] Virtual file system (VFS) support not implemented

## Problem

perl-lsp does not support virtual file systems, limiting integration with remote development environments, web-based editors, and non-standard file systems. The server assumes all files are local and accessible via standard filesystem APIs.

## Background

LSP clients may use virtual file systems for:
- **Remote development**: SSH, containers, WSL (files accessed via remote protocol)
- **Web-based editors**: Files stored in cloud/backend, not local disk
- **Special schemes**: `untitled:`, `vscode-userdata:`, `git:`, etc.

Current perl-lsp behavior:
- Converts URIs to local file paths via `Url::to_file_path()`
- Uses `std::fs` for file operations
- Assumes `file://` scheme for all documents

## Impact

1. **Remote development**: VS Code Remote-SSH, Remote-Containers may not work properly
2. **Web editors**: GitHub Codespaces, Gitpod integration limited
3. **Untitled files**: New files not yet saved to disk aren't handled
4. **Special URIs**: Breaks on non-file schemes

## Evidence

From `crates/perl-lsp/src/runtime/lifecycle/workspace.rs`:
```rust
let root_path = Url::parse(root_uri).ok().and_then(|u| u.to_file_path().ok());
```

Only proceeds if URI can be converted to a file path. Virtual URIs fail here silently.

## LSP Virtual File System Support

LSP allows custom URI schemes. Servers should:
1. Accept any URI scheme from client
2. Use `workspace/fileOperations` or content from `textDocument/didOpen` instead of reading disk
3. Not assume `file://` scheme

### Phase 1: URI abstraction
- Create a `VirtualPath` or `FileId` type that abstracts over URI schemes
- Don't convert URIs to paths until actually needed for filesystem access
- Keep original URI for all LSP communication

### Phase 2: Content-based indexing
- Index content from `textDocument/didOpen` notifications, not just disk
- Support untitled/unsaved files
- Cache virtual file content

### Phase 3: Remote file support
- For non-file URIs, rely on client for file content
- Use `workspace/fileOperations` for directory scanning if supported
- Gracefully degrade features that require local files (e.g., module execution)

## Code Locations

- `crates/perl-lsp/src/runtime/lifecycle/workspace.rs` — URI-to-path conversion
- `crates/perl-workspace-discovery/src/lib.rs` — filesystem discovery
[...]

PR #4057, opened 2026-04-11
feat(workspace): tolerate non-file workspace URIs (#3521)

EffortlessSteven:
### Motivation

- The server currently assumes all workspace/document URIs are `file://` and converts them with `Url::to_file_path()`, which fails silently for remote/virtual schemes and prevents VFS workflows from integrating correctly.
- Handle non-file URI schemes gracefully so remote/devcontainer/untitled URIs do not crash features that only need URI identity and so filesystem-dependent features degrade cleanly.

### Description

- Use the shared URI helper `perl_uri::uri_to_fs_path` across runtime modules instead of `Url::to_file_path()` to centralize file-URI handling and preserve non-file URIs when conversion fails.
- During initialize, populate `WorkspaceFolderState.path` only when the workspace folder URI is file-backed and keep the original `uri` for LSP communication, avoiding forced conversions on non-file schemes.
- Make module-resolution and folder-matching helpers use the shared `source_path_from_uri`/`uri_to_fs_path` flow so `resolve_module_*` and `workspace_root` logic degrade when no filesystem path exists for a URI.
- Skip non-filesystem workspace folders during the workspace indexing scan with a debug trace instead of treating conversion failure as an error path.
- Add the `perl-uri` workspace dependency and a unit test `set_root_uri_non_file_scheme_keeps_root_path_none` that verifies `root_path` remains `None` for non-file URIs.

### Testing

- Ran formatter: `cargo fmt --all` which completed successfully.
- Built/check: `cargo check -p perl-lsp-rs` which completed successfully.
- Unit test: `cargo test -p perl-lsp-rs --lib set_root_uri_non_file_scheme_keeps_root_path_none -- --nocapture` which passed (`1 passed`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d9fba366d48333ba53fba4b78161bb)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2272,6 +2272,7 @@ dependencies = [
  "perl-subprocess-runtime",
  "perl-symbol-cursor",
  "perl-tdd-support",
+ "perl-uri",
  "perl-workspace-discovery",
  "perl-workspace-folder",
  "perl-workspace-ignore",
--- a/crates/perl-lsp/Cargo.toml
+++ b/crates/perl-lsp/Cargo.toml
@@ -89,6 +89,7 @@ perl-module-import = { workspace = true }
 perl-pod = { workspace = true }
 perl-module-resolution = { workspace = true }
 perl-source-file = { workspace = true }
+perl-uri = { workspace = true }
 perl-workspace-discovery = { workspace = true }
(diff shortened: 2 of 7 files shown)


EffortlessSteven:
Superseded by merged #4059.

Issue #3515, opened 2026-04-10
[workspace] workspace/configuration handler not implemented


EffortlessSteven:
# [workspace] workspace/configuration handler not implemented

## Problem

perl-lsp does not implement the `workspace/configuration` request (LSP method), which means it cannot fetch configuration dynamically from the client. This limits the ability to:
1. Support per-workspace-folder configuration in multi-root workspaces
2. React to configuration changes without full server restart
3. Use LSP's configuration scoping (global vs workspace vs workspace-folder)

## Background

LSP defines `workspace/configuration` for servers to request configuration from clients:
- Client sends `workspace/configuration` request with configuration items
- Server uses the response to adjust behavior

perl-lsp currently:
- Loads `.perl-lsp.toml` at startup
- Handles `didChangeConfiguration` for some settings
- Does NOT request configuration via `workspace/configuration`

## Impact

1. **Multi-root workspaces**: Each folder could have different settings, but server can't request per-folder config
2. **Dynamic settings**: Server can't query "what's the current value of setting X?" — must rely on initial values
3. **Configuration precedence**: LSP defines precedence (client settings > workspace settings), but without `workspace/configuration`, server can't resolve this

### Per-folder includePaths in multi-root workspace
```json
// VS Code settings.json
{
  "folders": [
    { "path": "backend", "name": "API" },
    { "path": "frontend", "name": "UI" }
  ],
  "settings": {
    "perl.workspace.includePaths": ["lib"],  // Different per folder!
  }
}
```

Server should request configuration scoped to each workspace folder.

### Dynamic feature enablement
Server could query `perl-lsp.features.enabled` to toggle features without restart.

## Implementation Notes

1. Add `workspace/configuration` request capability to server capabilities
2. Implement handler that requests configuration when needed
3. Cache configuration with TTL or invalidation on `didChangeConfiguration`
4. Use for multi-root: request per-folder configuration at startup

## Code Locations

- `crates/perl-lsp-protocol/src/capabilities.rs` — server capabilities
- `crates/perl-lsp/src/runtime/workspace.rs` — configuration handling
- `crates/perl-lsp-config/src/lib.rs` — configuration merging logic

## Acceptance Criteria

- [ ] Server declares `workspace.configuration` capability
- [ ] Handler can request configuration items from client
- [ ] Per-folder configuration works in multi-root workspaces
- [ ] Configuration cache invalidates on `didChangeConfiguration`
[...]

PR #4092, opened 2026-04-11
feat(workspace): request workspace/configuration when supported (#3515)

EffortlessSteven:
### Motivation
- Enable the server to fetch per-folder client configuration via the LSP reverse `workspace/configuration` request so multi-root per-folder settings and dynamic client-scoped overrides work correctly.
- Today the server only loads `.perl-lsp.toml` at startup and handles `didChangeConfiguration` without actively requesting client-scoped values, which prevents correct precedence and per-folder overrides.

### Description
- Add a tracked client capability flag `workspace_configuration_support` to `ClientCapabilities` and parse it during `initialize` so the server can gate reverse configuration requests.
- Implement `request_workspace_configuration_for_all_folders()` in the client-requests runtime layer that builds a `ConfigurationItem` for each workspace folder (section `perl`) and sends a `workspace/configuration` server→client request when the client advertises support, and call it after project config load and on `workspace/didChangeConfiguration` to refresh overlays.
- Feature is implemented as a graceful no-op when the client does not support `workspace.configuration` and preserves existing `.perl-lsp.toml`-derived behavior as the fallback (client overlays are intended to override TOML when present).

### Testing
- Ran `cargo fmt --all` to ensure formatting (succeeded).
- Ran targeted unit tests with `cargo test -p perl-lsp-rs --lib workspace_configuration` which exercises `initialize_parses_workspace_configuration_capability` and `initialize_defaults_workspace_configuration_capability_to_false`, and both tests passed.
- Verified the workspace crate compiles and the targeted tests ran successfully during local test runs (no test failures observed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69da1f08a2308333b48a2a582749b6dc)

### DIFF
--- a/crates/perl-lsp/src/runtime/client_requests.rs
+++ b/crates/perl-lsp/src/runtime/client_requests.rs
@@ -3,6 +3,7 @@
 //! Each method checks the relevant client capability before sending.
 
 use super::*;
+use lsp_types::{ConfigurationItem, Uri};
 
 #[allow(dead_code)]
 impl LspServer {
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
Closing in favor of #4093.

This lane sends `workspace/configuration`, but it stops short of the acceptance criteria on #3515 because it does not thread the client response back into per-folder effective config. The issue is not just about emitting the reverse request; it is about using the returned scoped settings so multi-root folders can actually diverge and refresh correctly.

#4093 is the winning implementation in this pack because it is the only one that wires request -> response -> folder config update.

Issue #3486, opened 2026-04-10
DAP: No end-to-end integration test for complete debug workflow


EffortlessSteven:
## User Scenario

The DAP implementation has many unit tests but no comprehensive integration test that validates a complete user debugging workflow from start to finish with a real Perl debugger process.

## Current Behavior

1. Unit tests mock the debugger or test in isolation
2. Full debug workflow never tested against real `perl -d` process:
   - Script launch with arguments
   - Breakpoint setting and hitting
   - Step-in/step-over/step-out
   - Variable inspection
   - Stack trace retrieval
3. Regressions in real debugger interaction not caught
4. Protocol compliance not validated end-to-end

## Expected Behavior

1. Integration test creates real Perl debugger process
2. Test script triggers breakpoint and inspects:
   - Breakpoint confirmed hit
   - Stack frames populated
   - Local variables accessible
   - Stepping commands work
   - Continuation resumes execution
3. Full session lifecycle tested: launch → break → inspect → continue → terminate

## Impact

- Silent failures when debugger interaction breaks
- Regression testing incomplete
- Feature additions (e.g., watchpoints) can't be validated in context
- v0.13.0 public alpha unprepared for real-world use

## Files Affected

- `crates/perl-dap/tests/` (new file: dap_end_to_end_tests.rs)
- `crates/perl-dap/tests/dap_comprehensive_test.rs` (enhance existing test)
- Test fixtures needed: scripts/debug_e2e_*.pl

## Acceptance Criteria

1. E2E test launches real perl -d with stdio transport
2. Script with 3+ breakpoints on different lines
3. Verifies: setBreakpoints response, stopped event, stackTrace accuracy
4. Tests variable inspection at each breakpoint
5. Tests continue/step commands change execution flow
6. Validates protocol message sequences

## Labels

`dap`, `testing`, `needs-spec`

EffortlessSteven:
## Scout Investigation Summary

### /scout-dedup
✓ #3486 is OPEN and active — no duplicate found. Related #4069 (DAP scorecard) and #3487 (truncation testing) noted.

### /scout-locate
**Existing test infrastructure:**
- `crates/perl-dap/tests/dap_smoke_e2e.rs` (154 LOC) — real Perl session test; covers launch with stopOnEntry + initialized event + disconnect, but NO breakpoints/stepping/inspection
[...]

Issue #4069, opened 2026-04-11
metrics(dap): design a debugger success scorecard — launch, attach, variables, evaluate, truncation, end-to-end


EffortlessSteven:
Part of the metric-stack umbrella #4062.

## Current state

DAP is tested at the protocol level (DAP crate unit tests) and via manual editor smoke. There's no published "does a real debug session actually work?" number.

## Proposed metrics

- **Launch success rate** — a fresh DAP session starts and stops on the expected breakpoint
- **Attach success rate** — attach to a running Perl process and break on a signal
- **Missing-dependency error classification rate** — when a debuggee fails to start, is the error message actionable? (Not just "exit code 1".)
- **Variables pane correctness** — for a stop at a known position, does the variables pane reflect the expected $scalar / @array / %hash contents?
- **Evaluate correctness** — `evaluate` expression at a stopped frame returns the expected value
- **Truncation / pagination correctness on deep data** — e.g. deeply nested hashes, large arrays, cyclic structures
- **End-to-end debug workflow pass rate** — a realistic multi-step session: set breakpoint → run → stop → step over → step into → inspect → continue → exit
- **Median time to first stopped event** — latency, not just correctness

## Why DAP needs its own scorecard

The verification protocol treats real-user behavior as a separate tier, and DAP is especially sensitive to that. A DAP session that launches but hangs in the variables pane is **"working"** by any protocol-level metric and **broken** by any user-level metric. The only way to catch it is to automate the full session.

## Open questions

- **Fixture debuggees.** Small Perl scripts with known stop points, known variable values at those points, known evaluate results. Location: `test_corpus/dap/`?
- **Reference editor.** VS Code first (primary surface), or editor-agnostic via raw DAP protocol?
- **Process model coverage.** `perl -d`, Devel::Camelcadedb alternative, XS/threading edge cases — how much to cover?

## Acceptance criteria

- [ ] DAP fixture debuggees in `test_corpus/dap/`
- [ ] Automated DAP session driver (headless, protocol-level)
- [ ] Scorecard surfaced at `docs/project/status/dap.md` (new file)
- [ ] Root-README metric row for "debugger confidence"

## Relation to live work

- `perl-dap` crate — primary implementation surface
- #420 and related DAP compliance tracked work

EffortlessSteven:
## Scout Findings: DAP Metrics Dashboard Current State

### Executive Summary

[...]

PR #4156, opened 2026-04-11
test(dap): add E2E workflow session harness and 3 integration tests (#3486)

EffortlessSteven:
## Summary

- Adds `DapWorkflowSession` helper struct in `crates/perl-dap/tests/common/mod.rs` — a higher-level wrapper around `DebugAdapter` + event channel for driving complete `perl -d` debug sessions in tests
- Adds 3 end-to-end integration tests in `crates/perl-dap/tests/dap_e2e_workflow_tests.rs` covering: single breakpoint hit + variable inspection, multi-breakpoint sequence, and step-over execution
- Three production fixes required to make the tests pass (see details below)

## Production fixes

**`context_re` Windows path support** (`src/debug_adapter/mod.rs`): Extended the file-path character class to allow `:` when immediately followed by a non-digit character. This matches Windows drive letter prefixes (`C:\path`) without consuming the `:42` line-number separator.

**Multiple `DB<N>` prompt stripping** (`src/debug_adapter/parsing.rs`): Changed `if let` to `while let` in `normalize_debugger_output_line`. A single `read_line` from `perl -d` stderr can contain multiple consecutive prompt tokens (e.g. `  DB<1>   DB<2> main::(/path:5):`) when the debugger processes several stdin commands between stops. Only the last prompt was being stripped, leaving a stray `DB<N>` prefix that prevented `context_re` from matching.

**`RunToBreakpoint` double-continue bug** (`src/debug_adapter/process.rs`): Removed the auto-continue `c\n` write from the output reader's `RunToBreakpoint` branch. The `c` sent by `configurationDone` is already driving `perl -d` toward the first user breakpoint. The output reader's implicit-stop detection was sending a *second* `c`, which queued up and ran past the eventual breakpoint, leaving `n` (from a subsequent `stepOver`) consumed in the post-termination state. Now the output reader just suppresses the `stopped` event for that implicit stop (keeping `state=Running`) without sending an extra `c`.

## Key design note

Test 3 (`step_over`) uses `BP_LINE_2` (line 5) rather than `BP_LINE_1` (line 4) because the Perl debugger does not re-trigger a breakpoint on the line where execution is already paused. With a breakpoint on the *initial stop line* (line 4), `c` runs *past* it; with a breakpoint on line 5, `c` correctly runs *to* it and stops, leaving stdin empty for the subsequent `n`.

## Test plan

- [x] `cargo test -p perl-dap --test dap_e2e_workflow_tests` — all 3 tests pass
- [x] `cargo test -p perl-dap` — no new failures (pre-existing `test_cpan_module_installation_fallback` failure is unrelated)
- [x] `cargo fmt -p perl-dap` — no formatting changes needed after fmt
- [x] `cargo clippy -p perl-dap` — no new warnings


EffortlessSteven:
**Deep review complete.** Verified the double-continue fix and pushed 4 improvements (commit ceb7f93e).

**Double-continue fix verified correct**: The production change in `process.rs` removes the `stdin.write_all(b"c\n")` from the output reader's `RunToBreakpoint` branch. Before: `configurationDone` AND the output reader both sent `c\n` to `perl -d`, queuing two continues. After: only `configurationDone` sends `c\n`; the output reader suppresses the implicit first-line stop by setting `state=Running` and `should_auto_continue=true` without writing to stdin. This is correct.

**Edge cases verified clean**:
- `stopOnEntry: true` path is unaffected — it sends `l\n` not `c\n`, and `RunToBreakpoint` mode is never set
- Exception/warning stops during `RunToBreakpoint` correctly bypass the suppression (exception_match check takes priority)
- `ResumeMode` uses `matches!()` macro patterns throughout — no exhaustive match arms that need updating for the new variant
- No race window beyond the pre-existing one (which predates this PR)

**Fix-forwards pushed** (commit ceb7f93e):

1. **Vacuous line assertions in test 2** — the `stopped` event never carries a `line` field in this server; the `if let Some(line)` guards were permanently dead. Replaced with `stackTrace` calls that return the actual frame line.

2. **Multi-breakpoint test was broken** — the new real assertions exposed that BP_LINE_1 (line 4) is the implicit first stop line and cannot be re-hit by the initial `c`. Fixed by using BP_LINE_2 and new BP_LINE_3 (line 6) as the two breakpoints. Added clear documentation of the constraint.

3. **`stack_trace()` helper extended** to return `(frame_id, source_path, frame_line)` instead of 2-tuple.

4. **Orphaned fixture removed** — `tests/fixtures/workflow_e2e.pl` was never referenced by any test. Deleted.

Tests after fix-forward: 3/3 E2E pass, 114/114 regression pass.

Logic: CORRECT | Tests: IMPROVED | Regression risk: LOW

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-4156",
  "current_version": {
    "sha": "ceb7f93ed2b7d5be8451cc032593a7bcabb3dd5f",
    "updated_at": "2026-04-11T19:01:43Z"
  },
  "label_bindings": [
    {
      "label": "merge-ready",
      "bound_at_version": "ceb7f93ed2b7d5be8451cc032593a7bcabb3dd5f",
      "bound_at_timestamp": "2026-04-11T14:58:46Z",
      "bound_by_agent": "pr-ready",
      "valid": true
    },
    {
      "label": "reviewed-deep",
      "bound_at_version": "ceb7f93ed2b7d5be8451cc032593a7bcabb3dd5f",
      "bound_at_timestamp": "2026-04-11T19:01:43Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Deep Review (post-merge) — PR #4156

**Summary: logic correct, two coverage gaps fixed forward.**

### 1. context_re Windows path fix — VERIFIED

Traced the regex against all requested cases via Python-translated pattern (possessive quantifiers stripped for testing):

- `C:\Users\name\perl script.pl:42` — matches correctly (colon followed by `\` is non-digit/space)
- `C:/Users/file.pl:7` — matches (Git Bash forward-slash path)
- `\server\share\file.pl:42` — matches (UNC, backslash is in `[^:)\s]`)
- `:42` line separator — correctly stops path capture (digit after colon is excluded by `[^:)\d\s]`)
- Paths with spaces — correctly rejected (spaces excluded from character class)

The `C:5` degenerate case (no backslash) matches with `file=C, line=5` — technically wrong but `C:5` is not a valid Perl path and cannot appear in practice.

### 2. Multiple DB<N> prompt stripping — VERIFIED

Traced the `while let` loop manually. Termination is guaranteed: each iteration slices `normalized` to everything after the closing `>`, shrinking the string. The loop exits when `find("DB<")` returns `None`. The `DB<100>` case (high prompt number) works correctly — `find('>')` scans to the first `>` within the prompt token. Unclosed `DB<` (no matching `>`) exits safely — the second `&&`-arm returns `None`, breaking the loop.

Concern flagged: no quadratic behaviour — K prompts × L remaining chars, K≤3 in practice.

### 3. RunToBreakpoint double-continue fix — VERIFIED

State machine traced: `configurationDone` sets `state=Running, mode=RunToBreakpoint` and writes exactly one `c\n`. The output reader — on seeing a non-breakpoint context line in this mode — sets `should_auto_continue=true` without writing any `c\n`, then skips the stopped event via the `continue` at process.rs:735. The queued `c` from `configurationDone` is the only one perl ever receives.

**No-breakpoints scenario**: `register_breakpoint_hit` returns `matched=false`, output reader suppresses the stop, perl's queued `c` runs the script to completion. No hang.

**state desync after `next`**: when user sends `stepOver` after stopping, `handle_next` sets `mode=Next`. `RunToBreakpoint` is cleared by the `last_resume_mode = Unknown` at line 710 on actual stop. No state leak.

### 4. Test quality — GOOD

- No vacuous assertions: line-number assertions use `stack_trace()` which round-trips through the real DAP stackTrace request
- BP_LINE_1/BP_LINE_2/BP_LINE_3 constraint documented clearly
- `drain_until_event` timeout loop correctly handles interleaved output events
- `DapWorkflowSession` is stateless between test instances (each test creates its own)

**Timing note**: tests rely on perl -d startup latency covering the window between launch and `configurationDone`. This is pre-existing behaviour in the DAP test suite and not a regression from this PR.

### Coverage gaps fixed forward (pushed to branch)

Two production fixes (context_re, normalize) had no direct unit tests — only E2E validation. Added 14 unit tests across parsing.rs and mod.rs:

- **6 tests** for `normalize_debugger_output_line`: single prompt, multiple consecutive, high prompt number (>99), no-prompt passthrough, unclosed `DB<`, three in sequence
- **8 tests** for `context_re`: Unix path, Windows backslash, Windows forward-slash, UNC, named function, named function + Windows path, spaces rejection, colon-digit as separator

All 128 lib tests pass after additions. Commit: `63de18c4`.

**Verdict: approved. Logic is correct, regression risk low.**

EffortlessSteven:
## Deep Review Wrapup

**Summary:** Reviewer-deep completed post-merge deep correctness review of PR #4156. Approved with 14 new unit tests pushed to branch, `reviewed-deep` label set, label receipt written.

### Ending state

- PR reviewed: https://github.com/EffortlessMetrics/perl-lsp/pull/4156 (already merged)
- Changes made: 14 new unit tests added across 2 files
- Test artifacts touched:
  - `crates/perl-dap/src/debug_adapter/parsing.rs` — 6 tests for `normalize_debugger_output_line`
  - `crates/perl-dap/src/debug_adapter/mod.rs` — 8 tests for `context_re`
- All 128 lib tests pass
- Status: approved — `reviewed-deep` label set, receipt written
- Commits pushed to branch: `63de18c4`, `71c5ae05` (format fix)

### What was verified

**context_re**: Traced regex against 8 cases including Windows drive letter (`C:\`), UNC (`\server\`), Git Bash forward-slash (`C:/`), path-with-spaces (rejected), and colon-digit separator. All behave correctly.

**normalize_debugger_output_line**: Traced while-let loop manually for 1-prompt, 3-prompt, `DB<100>`, no-prompt, and unclosed `DB<` inputs. No infinite loops, no OOB access. Termination guaranteed by shrinking slice.

**RunToBreakpoint state machine**: Verified the single `c\n` from `configurationDone` is the only one sent. No-breakpoints scenario runs to completion cleanly. State correctly resets to Unknown on actual stop (line 710). stepOver after breakpoint hit works because RunToBreakpoint is cleared before the next command is issued.

**Test quality**: No vacuous assertions. Line numbers come from real `stackTrace` round-trips. `DapWorkflowSession` is fully isolated per test. `drain_until_event` correctly handles interleaved output events.

### Retrospective

The most surprising thing: the two production fixes (context_re, normalize) had zero direct unit test coverage — only E2E validation. The `while let` loop fix is the more subtle one; it works correctly but the termination argument requires tracing through the slice-shrinking behavior, not just reading the code. A comment noting "each iteration shrinks normalized by at least (prompt_end+1) bytes, so termination is guaranteed" would help future reviewers.

The worktree contamination from other agents (pre-existing untracked files with fmt drift in `perl-parser-core` and `perl-workspace-index`) blocked the push and required manual `rustfmt` of those files before the ci-gate would clear. The bypass policy explicitly covers this case but the friction is real — formatters run across the full workspace, not just the staged files.

What would have helped: a way to scope the ci-gate fmt check to only the files I changed, or a pre-push hook that skips untracked files from other branches.

- **Logical next**: The branch `review-4147` has 2 commits ahead of origin with the 14 new unit tests. These belong on master, not as a standalone PR — a follow-up ops pass or wisdom agent could rebase them onto master or incorporate them if the review branch is retired.
- **Related issues**: #3486 (the issue this PR closes), #4069 (scorecard builder that will consume `DapWorkflowSession`)
- **Gotcha**: `DapWorkflowSession::stack_trace()` returns `(frame_id, source_path, frame_line)` — the API was extended from the original 2-tuple to 3-tuple in the builder's second commit. Any future callers should use the 3-element destructure.
- **Trap**: The `RunToBreakpoint` mode stays active until an actual breakpoint is hit. If a script has a breakpoint on a line where `perl -d` can't stop (e.g., a `BEGIN` block or compile-time code), the debugger may run to completion without triggering the breakpoint, and the client will wait for a `stopped` event that never arrives. This is a perl debugger limitation, not an adapter bug, but worth noting.
- **Confidence for next**: high

PR #4305, opened 2026-04-12
feat(diagnostics): fast/slow diagnostic split for snappier feedback (#4279)

EffortlessSteven:
## Summary

- Add `publish_parse_errors_fast()` in `diagnostics.rs` that immediately sends only parse-error diagnostics (code PL001) after parsing completes on `didChange`
- Insert fast-path call in `text_sync.rs` at both `didChange` dispatch sites, before the existing debounced slow path
- The debounced `publish_diagnostics` (250ms) follows and replaces the fast notification with the full diagnostic set including scope-analysis and perlcritic

## Latency improvement

Before: parse errors visible after 250ms debounce + 200ms semantic analysis = ~450ms  
After: parse errors visible in ~10ms (parse time only); full set follows at 250ms

## Architecture

Fast path only fires for:
- Push-diagnostic clients (pull-diagnostic clients are unaffected)
- Documents with at least one parse error

LSP `publishDiagnostics` is replace-mode — the slow path's second notification replaces the fast one atomically from the client's perspective.

## Test plan

- `test_fast_path_contains_only_parse_errors`: verifies first notification after `didChange` contains only PL001 codes
- `test_two_phase_diagnostic_delivery_on_change`: verifies two notifications arrive and the second contains non-parse-error codes

## Knowledge artifacts

- The worktree (branch `9c75d601`) has a pre-existing formatting drift in `perl-ci-hygiene/src/main.rs` unrelated to this change — the `--no-verify` push was necessary per bypass policy
- Edit tool uses Windows absolute paths which map to the main checkout, not the worktree; always use the worktree-prefixed path when editing files

### DIFF
--- a/crates/perl-lsp/tests/lsp_fast_slow_diagnostics_test.rs
+++ b/crates/perl-lsp/tests/lsp_fast_slow_diagnostics_test.rs
@@ -1,0 +1,122 @@
+//! Fast/slow diagnostic split tests (Issue #4279)
+//!
+//! Verifies that parse errors are published immediately (fast path) on didChange,
+//! while the full diagnostic set follows after the debounce (slow path).
+//! The fast path publishes ONLY parse-error-coded diagnostics (PL001).
+
+mod support;
+
+use serde_json::json;
+use support::lsp_harness::LspHarness;
+
+/// Fast path: the first publishDiagnostics notification after didChange must
+/// contain only parse-error diagnostics (code "PL001"), not scope-analysis
+/// or quality-check diagnostics.
+///
+/// In production this fires before the 250ms debounce; in tests both fire
+/// synchronously so we check that the FIRST notification is parse-errors-only.
+#[test]
+fn test_fast_path_contains_only_parse_errors() {
+    let mut harness = LspHarness::new();
+    harness
+        .initialize(Some(json!({
+            "textDocument": {
+                // No pull-diagnostics capability: force push path
+            }
+        })))
+        .expect("initialize should succeed");
+
+    let uri = "file:///fast_slow_test.pl";
+
+    // Open a clean document first
+    harness.open(uri, "my $x = 1;\n").expect("open should succeed");
+
+    // Drain any existing notifications (from didOpen)
+    let _ = harness.drain_notifications(Some("textDocument/publishDiagnostics"), 400);
+
+    // Now change to a document with a parse error
+    harness.change_full(uri, 2, "my $broken = ;\n").expect("change should succeed");
+
+    // Wait for notifications to arrive
+    let all_notifications =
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder completed fast/slow diagnostic split (quick win 3 of #4279). 1 PR, 2 tests, 97 net new lines.

### Ending state

- PR: https://github.com/EffortlessMetrics/perl-lsp/pull/4305
- Tests: `test_fast_path_contains_only_parse_errors`, `test_two_phase_diagnostic_delivery_on_change` — both pass
- Files changed:
  - `crates/perl-lsp/src/runtime/diagnostics.rs` — new `publish_parse_errors_fast()` method
  - `crates/perl-lsp/src/runtime/text_sync.rs` — fast-path calls on both `didChange` dispatch sites
  - `crates/perl-lsp/tests/lsp_fast_slow_diagnostics_test.rs` — new test file

### Latency improvement

Before: parse errors visible after 250ms debounce + ~200ms semantic = ~450ms  
After: parse errors visible in ~10ms (parse time only); full set follows at 250ms  
Estimate: **~440ms faster for the user's first parse-error feedback**

### Retrospective

**What was harder than expected:** The Edit tool uses Windows absolute paths (`H:\Code\Rust\...`) which map to the MAIN CHECKOUT, not the worktree. All edits accidentally went to the main checkout first. Took two full edit cycles to realize the issue and fix it. Always use the full worktree path (`H:\Code\Rust\perl-lsp\.claude\worktrees\agent-a94abf51\...`) when editing files in a worktree context.

**What surprised me:** The worktree branch (`9c75d601`) is BEHIND master by 7 commits but also has `build_parse_error_hint` already wired in — which means the codebase on this branch is already richer than on older worktrees. The `cargo check` results correctly reflected the worktree's state once I stopped accidentally compiling the main checkout.

**What I'd do differently:** Check from the start which path the Edit tool is actually writing to (verify with `git diff` IMMEDIATELY after each edit rather than after multiple edits).

### Breadcrumbs for reviewer

- The two `publish_parse_errors_fast` calls in `text_sync.rs` are mirrored: one inside the `#[cfg(feature = "workspace")]` background-indexing path, one in the fallback path. Both must call the fast path before the debounce.
- The spec note about "coordinate with the generation-aware guard builder" — that builder touches `DocumentState::generation` tracking. My changes only touch the diagnostic publish call sites. No conflict expected.
- Pre-existing formatting drift in `crates/perl-ci-hygiene/src/main.rs` in this worktree caused the pre-push hook to fail; used `--no-verify` per bypass policy (environmental issue out-of-band of this change).

### Coordination note

The other #4279 builder handles the generation-aware stale check (a different code path — `DocumentState::generation` in text_sync.rs change detection). My changes touch only the diagnostic publish call sites. Safe to merge independently.

PR #4387, opened 2026-04-14
feat(completion): add hash key completion after $hash{ (#4264)

### LINKED ISSUE #4264: Completion: no suggestions for hash key completion after $hash{
## Problem

After \`$hash{\`, completion should suggest known hash keys from the context, but currently returns empty. This is a completeness gap, not just ranking.

**Evidence:**
- No code path in \`crates/perl-lsp-completion/src/completion.rs\` or submodules handles hash key completion
- Current completion context detection (around line 1030) has no case for \`{key\` pattern
- Hash variables are completed, but not their keys

**Example:**
\`\`\`perl
my %config = (host => 'localhost', port => 5432);
$config{ho<cursor>}  # Should suggest: host
\`\`\`

**Current:** No completions offered  
**Expected:** \`host\` and \`port\` suggested (or all defined keys)

## Root Cause

\`analyze_context\` in \`crates/perl-lsp-completion/src/completion.rs:969\` does not detect \`{\` trigger or parse hash variable + key context. The completion provider has no module to suggest hash keys based on known hash variables.

## Options

1. **Key completion from symbol table** — Detect \`$var{\` pattern in \`analyze_context\`, extract variable name, look up symbol table for that variable. If found with hash init value, suggest keys from the initializer. **Tradeoff:** requires AST analysis of hash initialization. **Effort:** MEDIUM.

2. **Defer to future** — File as known limitation, low priority. **Tradeoff:** users get no help. **Effort:** NONE.

3. **Minimal stub** — Add trigger for \`{\` but return all variables in scope (not smart, but better than nothing). **Tradeoff:** noisy. **Effort:** EASY.

## Recommendation

Option 2 (defer) — genuine feature, not a quick ranking fix. Requires semantic analysis of hash construction. File as enhancement, mark for future roadmap.

## Scope

- **Crate(s):** perl-lsp-completion
- **Files:** crates/perl-lsp-completion/src/completion.rs, possibly new module for hash analysis
- **Effort:** MEDIUM (4-6h)
- **Impact:** Adds new completion context type

---
_Scout finding: completion feature gap. Defer to future sprint._

EffortlessSteven:
## Summary

- Detect `$varname{key_prefix<cursor>` context via backward source-text scan in a new `detect_hash_key_context` pre-filter
- Extract known keys from `%varname = (...)` fat-comma literals and `$varname{key} = ` assignment patterns using `collect_hash_keys_from_source`
- Return `CompletionItemKind::Property` items filtered by typed prefix via `add_hash_key_completions`

## Changes

- `crates/perl-lsp-completion/src/completion.rs` — three new `fn`s + dispatch branch + 5 tests

**Insertion point:** After `is_has_options_key_context` check, before `object_pad_constructor_package`, as specified by plan-reviewer.

**Dispatch guards:** `!context.in_comment && !context.in_string && !context.in_regex` before calling `detect_hash_key_context`.

**Trigger characters unchanged:** `{` was NOT added to server capabilities. Feature works on manual invocation (Ctrl+Space) to avoid noise in hash ref constructors, blocks, and conditionals.

## Evidence

- **Commits**: 1
- **Tests**: 440 passed (5 new hash key completion tests + 435 existing), 0 failed
- **Lint**: clean (0 clippy warnings)
- **Files**: 1 changed, +350/-0 lines

## Test Plan

1. Open a Perl file with `my %config = (host => 'localhost', port => 5432);`
2. Type `$config{ho` and invoke completion manually (Ctrl+Space)
3. Verify `host` appears and `port` is filtered out
4. Type `$config{` and invoke — verify both `host` and `port` appear
5. Verify `$ref->{ho` does NOT produce hash key completions (hashref deref guard)
6. Verify completion inside `# $config{ho` comment is empty

## Unknowns

- Multi-line hash declarations where `%hash = (` and the keys are on subsequent lines are supported (backward scan handles newlines)
- Hash keys that are complex expressions (not bare words or quoted strings) are not extracted — only word-like keys are suggested
- `$hashref->{key}` (hashref deref) is explicitly excluded; a follow-up could add that via type inference
- Keys defined via `push @{$hash{slice}}` or other indirect patterns are not extracted

Closes #4264

### DIFF
--- a/crates/perl-lsp-completion/src/completion.rs
+++ b/crates/perl-lsp-completion/src/completion.rs
@@ -1564,8 +1564,14 @@ impl CompletionProvider {
             return None;
         }
 
-        // Require `$` sigil immediately before the variable name
+        // Require `$` sigil immediately before the variable name.
+        // Also reject `$$var{` (double-sigil deref) by ensuring the char before `$`
+        // is not itself a `$` — that would indicate `$$var{key}` which is a scalar-ref
+        // dereference, not a plain hash subscript.
         if var_name_start == 0 || before_brace.as_bytes()[var_name_start - 1] != b'$' {
+            return None;
+        }
+        if var_name_start >= 2 && before_brace.as_bytes()[var_name_start - 2] == b'$' {
             return None;
         }
 
--- a/crates/perl-lsp-completion/src/completion.rs
+++ b/crates/perl-lsp-completion/src/completion.rs
@@ -779,6 +779,18 @@ impl CompletionProvider {
             self.add_has_type_completions(&mut completions, &context);
         } else if self.is_has_options_key_context(source, position) {
             self.add_has_option_completions(&mut completions, &context);
+        } else if !context.in_comment
+            && !context.in_string
+            && !context.in_regex
+            && let Some((varname, key_prefix)) = Self::detect_hash_key_context(source, position)
+        {
+            Self::add_hash_key_completions(
+                &mut completions,
+                &context,
+                source,
+                &varname,
+                &key_prefix,
+            );
         } else if let Some(package_name) = self.object_pad_constructor_package(source, position) {
             self.add_object_pad_constructor_completions(&mut completions, &context, &package_name);
         } else if (context.trigger_character == Some('>') || context.trigger_character == Some('-'))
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder implemented hash key completion after `$hash{` in perl-lsp-completion. 1 PR created, 5 tests added, 440 tests passing.

**Ending state:**
- PR created: https://github.com/EffortlessMetrics/perl-lsp/pull/4387
- Tests added: 5
  - `completion::tests::test_hash_key_completion_basic`
  - `completion::tests::test_hash_key_completion_empty_prefix`
  - `completion::tests::test_hash_key_completion_does_not_fire_for_hashref_deref`
  - `completion::tests::test_hash_key_completion_in_comment_no_suggestions`
  - `completion::tests::test_hash_key_completion_unknown_variable_returns_empty_for_that_hash`
- Files changed: `crates/perl-lsp-completion/src/completion.rs` (+350 lines)
- Tests passing: 440 passed, 0 failed (9 suites)
- Next step: ready for review

**Retrospective:**

The plan-reviewer's spec was exceptionally complete — the exact insertion point (after `is_has_options_key_context`, before `object_pad_constructor_package`), the function signatures, and all 8 edge cases were specified precisely, which made implementation straightforward. The only non-trivial decision was how to handle the `between` validation (text between `{` and the typed prefix), which protects against false positives like `$hash{ $expr + ` — the approach of checking for non-alphanumeric/non-whitespace chars proved correct.

The plan-reviewer's insight that `context.prefix` is a bare word at `$config{ho` (not `$config{ho`) was the critical structural insight that made this implementable correctly as a pre-filter. Without that, a naive `prefix.starts_with('$')` check would have missed the entire hash key context.

One surprise: the pre-push hook runs a full `cargo test -p perl-lsp-completion` before allowing push, which added ~15s but confirmed the implementation was solid before the PR went up.

- Logical next: reviewer agent on PR #4387
- Related: `$ref->{key}` hashref deref key completion could be a follow-up (requires type inference to know which hash the ref points to)
- Gotchas: `detect_hash_key_context` must remain a pre-filter (not integrated into the sigil-prefix dispatch) because `context.prefix` is a bare word, not `$varname{prefix`, at cursor time
- Gotchas: `collect_hash_keys_from_source` only extracts word-like keys (alphanumeric + `_`); complex expression keys (e.g., `$hash{$computed_key}`) are deliberately excluded
- Traps: do NOT add `{` to trigger characters in `crates/perl-lsp-protocol/src/capabilities.rs` — this breaks the `lsp_capability_snapshot_test` and causes noisy completions in all brace contexts
- Confidence for next: high — implementation matches spec exactly, all tests pass, no regressions

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-4387",
  "current_version": {
    "sha": "32c652a2ec98eac53b2e6a3a2cdbfc5425e89cdf",
    "updated_at": "2026-04-14T06:35:05Z"
  },
  "label_bindings": [
    {
      "label": "in-review",
      "bound_at_version": "85e61c1c3d1f1da7f200fbbe6734196d953a2f38",
      "bound_at_timestamp": "2026-04-13T00:00:00Z",
      "bound_by_agent": "reviewer",
      "valid": false
    },
    {
      "label": "reviewed-deep",
      "bound_at_version": "32c652a2ec98eac53b2e6a3a2cdbfc5425e89cdf",
      "bound_at_timestamp": "2026-04-14T06:35:24Z",
      "valid": true
    },
    {
      "label": "merge-ready",
      "bound_at_version": "32c652a2ec98eac53b2e6a3a2cdbfc5425e89cdf",
      "bound_at_timestamp": "2026-04-14T06:35:24Z",
      "bound_by_agent": "pr-ready",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
**Deep review complete. Logic verified. Two issues found and fixed forward on the branch.**

**Fix 1 — `$$varname` double-sigil false positive (logic bug)**

`detect_hash_key_context` was checking that the char immediately before the variable name is `$`, but not checking whether THAT `$` was itself preceded by another `$`. For `$$data{ke`, the scan extracted `var_name = "data"` with `$` at `var_name_start - 1` — but that `$` was the second `$` of `$$`. If `%data` was also defined in scope, `collect_hash_keys_from_source` would find its keys and produce false completions. Fixed by adding a guard: if `before_brace.as_bytes()[var_name_start - 2] == b'$'`, return `None`.

Added test: `test_hash_key_completion_double_sigil_same_name_no_false_positive` — the existing test used `$$ref{ke` with no `%ref` in scope, so it passed vacuously (varname mismatch hid the bug).

**Fix 2 — 3 edge case test assertions were wrong (test bugs)**

Three tests in `extended_unit_tests.rs` asserted that keys would appear even when the prefix would filter them out:
- `test_hash_key_completion_quoted_keys`: prefix `fir` cannot match `second_key`
- `test_hash_key_completion_multiline_definition`: prefix `ho` cannot match `port`
- `test_hash_key_completion_mixed_definitions`: prefix `c` cannot match `shade`

All three intended to verify key EXTRACTION (not filtering), but inadvertently used a typed prefix. Fixed by changing cursor position to empty prefix. Filtering behavior is separately covered by `test_hash_key_completion_prefix_filtering`.

**Logic analysis summary:**
- Backward `{` scan with depth tracking: correct
- `->` hashref guard: correct; `$ref -> {` naturally returns empty var_name after `>`
- `%varname` substring scan: no false positives — guarded by `=` and `(` checks
- `extract_fat_comma_keys`: values containing `=>` handled correctly via depth-tracked `,` terminator
- `$varname{key}` at cursor: correctly skipped (no closing `}` present)
- Prefix filtering: consistent with all other providers in the codebase
- `between` underflow: impossible by construction

**Regression risk: LOW.** New code behind `else if` guard. All 456 tests pass (22 hash key tests, up from 18 passing before).

Research verification: SKIPPED — no external Perl language, LSP protocol, or crate API claims.

EffortlessSteven:
## Deep Reviewer Wrapup

**Summary:** reviewer-deep completed deep correctness review of PR #4387 (hash key completion after `$hash{`). 2 bugs fixed, 1 new test added, 456 tests passing.

### Ending State

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/4387
- Changes made: 2 logic/test fixes pushed in commit `92683c1d3`, formatted in `32c652a2e`
- Files touched:
  - `crates/perl-lsp-completion/src/completion.rs` — added double-sigil guard in `detect_hash_key_context`
  - `crates/perl-lsp-completion/tests/extended_unit_tests.rs` — fixed 3 bad assertions, added 1 new test
- Tests passing: 456 passed, 0 failed (22 hash key tests, up from 18 passing before)
- Labels: `reviewed-deep` set, `merge-ready` set, `in-review` removed
- Status: approved, PR exited draft

### Bugs Found and Fixed

**Bug 1 — `$$varname` double-sigil false positive in `detect_hash_key_context`:**
When the cursor is at `$$data{ke`, the function correctly finds `$` before `data`, but that `$` is the second sigil of `$$data`. If `%data` happens to exist in the source, `collect_hash_keys_from_source` would find it and produce false completions. Fix: check that `before_brace.as_bytes()[var_name_start - 2] != b'$'` before returning `Some`.

The existing test (`test_hash_key_completion_double_sigil_dereference`) used `$$ref{ke` with no `%ref` in scope — it passed vacuously via varname mismatch, not by detecting the double-sigil correctly. Added `test_hash_key_completion_double_sigil_same_name_no_false_positive` which uses `$$data{ke` with `%data` defined, proving the fix is actually exercised.

**Bug 2 — 3 edge case tests had wrong assertions:**
`test_hash_key_completion_quoted_keys`, `test_hash_key_completion_multiline_definition`, and `test_hash_key_completion_mixed_definitions` all asserted that keys would appear even when prefix filtering would exclude them (`second_key` with prefix `fir`, `port` with prefix `ho`, `shade` with prefix `c`). These tests intended to verify extraction, not filtering. Fixed by using empty prefix in all three.

### What Was CORRECT (verified, no changes needed)

- Backward `{` scan with depth tracking: correctly finds the last unmatched `{`
- `->` hashref guard: covers `$ref->{`; `$ref -> {` naturally returns None via empty var_name after `>`
- `%varname` substring scan: substring matches are safe because the `=` and `(` checks gate them
- `extract_fat_comma_keys` with values containing `=>`: correctly uses depth-tracked `,` as terminator
- `$varname{key}` at cursor position: correctly skipped (no closing `}` present)
- `between` underflow: impossible by construction (key_prefix.len ≤ position - brace_pos - 1)
- Prefix filtering: consistent with all other completion providers in the codebase
- Dispatch ordering: correctly placed after `is_has_options_key_context` and before object-pad/method branches

### Retrospective

The hardest part was distinguishing test bugs from implementation bugs. Three failing tests initially looked like extraction failures, but tracing the prefix filter revealed the assertions were simply wrong about which keys would survive filtering. The key insight: when a test checks that key X appears with prefix Y where X doesn't start with Y, the test cannot pass unless filtering is broken or disabled — so a passing assertion would prove a bug, not correct behavior.

The double-sigil false positive was subtle because the existing test passed for the wrong reason (different varname). Vacuous test detection requires asking: "would this test still pass if the feature code were commented out?" For `$$ref{ke` with no `%ref` in scope, the answer is yes — the hash key lookup would return empty regardless. That's the litmus test from the analyze skill that caught it.

For future deep reviewers: source-text scanning functions like `collect_hash_keys_from_source` deserve special scrutiny for substring name collisions. The `=` gate here saves the day, but the pattern is fragile and worth a comment explaining why it's safe.

- Logical next: ops merge batch — PR is merge-ready, CI will trigger on current HEAD `32c652a2e`
- Related: issue #4264 can be closed when this merges
- Gotchas: `detect_hash_key_context` is a free function (no `&self`) — it works on raw source bytes, not the symbol table. Any enhancement that needs type inference would need a different architecture.
- Follow-up worth filing: hash key completion for `$hash{'quoted_key'` and `$hash{"interpolated_$key"` patterns currently returns None (the `between` check blocks them). A follow-up could handle the quoted case by stripping the leading quote before computing `key_prefix`.
- Confidence for next: high — core logic is sound, tests are accurate, CI should green.

PR #4399, opened 2026-04-15
fix(hover): associate inline POD inside sub bodies with subroutine hover (#3407)

### LINKED ISSUE #3407: POD inside subroutine bodies not associated with subroutine hover
## Summary

POD appearing INSIDE a subroutine body (rare but valid Perl) is not associated with that subroutine for hover documentation.

## Location

`crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs:57-107`

## Current Behavior

The `extract_documentation()` function only looks backwards from a position, so inline POD inside sub bodies is missed.

## Example

```perl
sub process_data {
    =pod
    Internal documentation for this sub
    =cut
    my $data = shift;
    ...
}
```

The POD "Internal documentation for this sub" is not shown when hovering over `process_data`.

## Root Cause

```rust
pub(super) fn extract_documentation(&self, start: usize) -> Option<String> {
    let before = &self.source[..start];  // Only looks backwards!
    // ...
}
```

## Suggested Fix

For subroutine nodes, search both:
1. Before the sub definition (current behavior)
2. Inside the subroutine body for inline POD

Could be implemented by:
```rust
fn extract_sub_documentation(&self, sub_node: &Node) -> Option<String> {
    // 1. Try before the sub
    self.extract_documentation(sub_node.location.start)
        // 2. Try inside the body for inline POD
        .or_else(|| self.find_pod_in_sub_body(sub_node))
}
```

## Related

- Parser handles inline POD without errors (tested in `pod_inline_parsing_tests.rs`)
- The POD is preserved in trivia but not associated with symbols

EffortlessSteven:
## Summary

Closes #3407.

POD blocks that live **inside** a subroutine body (a rare but legitimate Perl idiom — the "literate programming" style where narrative documentation sits alongside the code it describes) were silently dropped from hover. The existing `extract_documentation` helper only searches the source text *before* a node's start offset, so it could never see `=pod … =cut` appearing after the opening brace.

This PR teaches the semantic analyzer to fall back to an **in-body** POD scan for subroutine-like nodes when no preceding POD or comment block is found. Preceding POD continues to win — it's the idiomatic documentation location — so no existing hover output changes.

## Example

Before this change, hovering `process_data` returned no documentation. After it, the hover now includes "Internal documentation for this sub":

```perl
sub process_data {
=pod

Internal documentation for this sub.

=cut

    my $data = shift;
    return $data * 2;
}
```

## What changed

`crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs`

- **New** `extract_sub_documentation(node_start, body_start, body_end)` — thin wrapper that first calls the existing `extract_documentation(node_start)` and, only if that returns `None`, falls back to an in-body POD scan via the new helper. Preceding POD → inline POD is the precedence rule.
- **New** `find_pod_in_range(start, end)` — scans a byte range for the first `(?ms)^(=[a-zA-Z0-9].*?\n=cut\b)` POD block. The regex requires both the opening directive and `=cut` to be anchored at column 0 (per `perlpod`), so indented text that happens to contain an `=` cannot produce a false positive. The compiled regex is cached in a `OnceLock` to match the pattern used by `extract_documentation`.
- Wired the new helper into three hover sites that previously called `extract_documentation(node.location.start)` directly:
  - `NodeKind::Subroutine` named sub (line 141)
  - `NodeKind::Subroutine` anonymous sub / closure (line 176)
  - `NodeKind::Method` (line 208)

No other call sites changed — variable, package, and list-variable hovers still use the preceding-text search, which is correct for them.

## Why this design

The issue template suggested a separate `extract_sub_documentation` that composes preceding + in-body lookups. I followed that structure because:

1. **Precedence is explicit and testable** — preceding POD wins, in-body POD is strictly a fallback. This matches the test for `test_subroutine_prefers_preceding_pod_over_inline_pod`.
2. **No regression surface for non-sub nodes** — only callers that pass a body range opt into the in-body scan. Variable/package hovers are untouched.
3. **Body range is already available** — `NodeKind::Subroutine` and `NodeKind::Method` both carry a `body: Box<Node>`, so we reuse its `location.start..location.end` rather than re-walking the AST.
4. **Column-0 anchoring** — real Perl POD *must* start at column 0. Anchoring the regex at line boundaries (`(?m)^=…`) avoids matching things like `my $x = "=pod"` or indented fake-POD inside string literals.

## Tests

Added 5 new tests under `crates/perl-semantic-analyzer/src/analysis/semantic/mod.rs`:

| Test | What it proves |
|------|---------------|
| `test_subroutine_with_inline_pod_in_body_hover` | `=pod … =cut` inside the body surfaces in hover documentation. |
| `test_subroutine_with_head_inline_pod_in_body` | `=head2 … =cut` (not just `=pod`) is recognised — any POD directive works. |
| `test_subroutine_prefers_preceding_pod_over_inline_pod` | Precedence rule: preceding POD takes priority, inline POD is never shown when a preceding block exists. |
| `test_method_with_inline_pod_in_body_hover` | Regression guard for the `NodeKind::Method` path used by `class … { method … }` blocks. |
| `test_subroutine_without_any_pod_has_no_documentation` | Negative test: a plain sub still yields `documentation: None` so downstream consumers can distinguish "no doc" from "empty doc". |

All new tests were verified red-before-green (confirmed failing prior to the implementation change), then green after. The full `perl-semantic-analyzer` suite (~700 tests across lib + integration targets) passes.

## Verification

```
$ cargo test -p perl-semantic-analyzer --lib
test result: ok. 121 passed; 0 failed; 0 ignored

$ cargo test -p perl-semantic-analyzer      # lib + all integration targets
test result: ok. (all green)

$ cargo test -p perl-lsp-providers --lib
test result: ok. 4 passed; 0 failed

$ cargo test -p perl-parser --lib
test result: ok. 12 passed; 0 failed

$ cargo clippy -p perl-semantic-analyzer --lib
Finished (clean)

$ cargo fmt -p perl-semantic-analyzer -- --check
(clean)
```

*Note:* `cargo clippy -p perl-semantic-analyzer --lib --tests` surfaces two `expect_used` errors in `tests/frameworks_web.rs:343,345` — these are pre-existing on `master` and unrelated to this change (the file is untouched by this PR).

## Risk assessment

- **Scope:** strictly additive to the hover documentation path; no change to semantic tokens, symbol extraction, scope analysis, or any other consumer.
- **Non-sub hovers:** untouched. Variable/package/import hovers still use `extract_documentation` directly.
- **Perf:** one extra regex run *per sub hover*, and only when no preceding POD exists. The regex is cached in a `OnceLock` like the existing one. Body scans are bounded by the sub body's byte range.
- **False positives:** prevented by requiring both `=<alpha>` and `=cut` to start at column 0.
- **Backward compat:** every sub that previously returned a documentation string still returns the same string (preceding POD still wins). This can only *add* documentation where there was none; it cannot change existing hover output.

## Follow-ups (out of scope, not required for merge)

- Consider applying the same fallback to comment-style docs inside sub bodies (unusual, but mechanically similar).
- Consider a `#[ignore]`-removal follow-up for the pre-existing `frameworks_web.rs` clippy violations.

Fixes #3407.

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/semantic/mod.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/semantic/mod.rs
@@ -1557,6 +1557,191 @@ sub process {
     }
 
     #[test]
+    fn test_subroutine_with_inline_pod_in_body_hover() -> Result<(), Box<dyn std::error::Error>> {
+        // Inline POD inside a subroutine body (rare but valid Perl) should be
+        // associated with the subroutine for hover documentation. See issue #3407.
+        //
+        // Per perlpod, POD directives must start at column 0 — the block below
+        // temporarily drops out of the sub body for documentation and resumes
+        // with `=cut`, which is a legitimate Perl idiom.
+        let code = r#"sub process_data {
+=pod
+
+Internal documentation for this sub.
+
+=cut
+
+    my $data = shift;
+    return $data * 2;
+}
+"#;
+        let mut parser = Parser::new(code);
+        let ast = parser.parse()?;
+        let analyzer = SemanticAnalyzer::analyze_with_source(&ast, code);
+
+        let sub_symbols =
+            analyzer.symbol_table().find_symbol("process_data", 0, SymbolKind::Subroutine);
+        assert!(!sub_symbols.is_empty(), "Should find sub process_data");
+
+        let hover = analyzer.hover_at(sub_symbols[0].location);
+        assert!(hover.is_some(), "Should have hover for sub process_data");
+
+        let hover = hover.unwrap();
+        assert!(
+            hover.signature.contains("sub process_data"),
+            "Hover should show sub signature, got: {}",
+            hover.signature
+        );
+        let doc = hover
+            .documentation
+            .as_ref()
+            .ok_or("Sub hover should include documentation from inline POD in body")?;
+        assert!(
+            doc.contains("Internal documentation"),
+            "Sub hover should contain inline POD body text, got: {}",
+            doc
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_subroutine_prefers_preceding_pod_over_inline_pod()
+    -> Result<(), Box<dyn std::error::Error>> {
+        // When POD precedes the sub AND inline POD exists in the body,
+        // the preceding POD takes precedence (it is the idiomatic doc location).
+        let code = r#"=head2 outer
+
+Preceding doc for process_data.
+
+=cut
+
+sub process_data {
+=pod
+
+Inline doc inside body.
+
+=cut
+
+    my $data = shift;
+    return $data;
+}
+"#;
+        let mut parser = Parser::new(code);
+        let ast = parser.parse()?;
+        let analyzer = SemanticAnalyzer::analyze_with_source(&ast, code);
+
+        let sub_symbols =
+            analyzer.symbol_table().find_symbol("process_data", 0, SymbolKind::Subroutine);
+        assert!(!sub_symbols.is_empty(), "Should find sub process_data");
+
+        let hover = analyzer.hover_at(sub_symbols[0].location).ok_or("hover missing")?;
+        let doc = hover.documentation.as_ref().ok_or("documentation missing")?;
+        assert!(doc.contains("Preceding doc"), "Preceding POD should take precedence, got: {doc}");
+        assert!(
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as duplicate of #4403. Ensemble pattern — kept the best of the cluster per `feedback_codex_ensemble_pattern.md`.

Issue #4066, opened 2026-04-11
metrics(editor): design an editor-intelligence scorecard for hover, completion, goto-definition, rename, symbols


EffortlessSteven:
Part of the metric-stack umbrella #4062.

## Current state

The feature catalog says every editor feature is implemented (102/102). What we don't measure is whether each feature **returns the right thing** in realistic workflows.

### Hover
- Non-null rate (how often hover returns something)
- Correctness rate (does the returned content match the symbol?)
- Declaration-context accuracy (package / sub / variable distinguished correctly)
- POD / external-doc link coverage

### Completion
- Top-1 relevance (expected suggestion was first)
- Top-5 relevance (expected suggestion was in top 5)
- Empty-when-should-not-be-empty rate
- Noise rate (irrelevant keywords in typed semantic contexts)
- Context-specialized pass rate:
  - method completion after `$obj->`
  - import completion
  - constructor-key completion
  - Moose / Moo / Object::Pad specific
- Latency p50 / p95

### Goto-definition
- Exact-hit rate
- Multi-definition disambiguation correctness
- Imported-symbol resolution rate
- Inherited-method resolution rate (related to #3482)
- Cross-file / cross-workspace success rate

### Document symbols
- Returns non-empty for every parsable file
- Symbols match expected package + sub + variable structure

### Rename
- Cross-file rename correctness
- Preserve comments / whitespace invariant
- No-partial-rename invariant (all-or-nothing)

## Open questions

- **Gold corpus.** Shared with scorecard **B** (diagnostics) so each fixture carries both expected diagnostics and expected hover/completion/goto/rename answers.
- **Automation harness.** Proposal: headless LSP driver that replays `textDocument/*` requests against fixture files and asserts responses. Could reuse parts of the existing LSP test infrastructure.
- **Relevance labeling.** "Top-1 relevance" needs someone to label the expected answer for each fixture position; label once, re-use forever.

## Acceptance criteria

- [ ] Headless LSP driver for replayable `textDocument/*` requests
- [ ] Gold corpus shared with #B
- [ ] Scorecard surfaced at `docs/project/status/editor.md` (new file)
- [ ] Root-README metric row links to it

## Relation to live work

- #3482 — inherited method resolution (feeds goto-definition + completion rows)
- #3472, #3475, #3476 — symbol visibility (feeds completion + hover rows)

EffortlessSteven:
[...]

PR #4401, opened 2026-04-15
fix(hover): surface inline POD inside sub bodies + stop POD leak between decls (#3407)

### LINKED ISSUE #3407: POD inside subroutine bodies not associated with subroutine hover

EffortlessSteven:
Closes #3407.

## Summary

Two related hover-documentation fixes in `perl-semantic-analyzer`:

1. **Inline POD inside subroutine bodies is now surfaced as hover docs.** Perl permits `=pod...=cut` blocks anywhere a statement is valid, including inside a `sub { ... }` body. The analyzer previously only searched *backwards* from a subroutine's start position, so inline POD was silently dropped.
2. **The backward POD search no longer leaks a POD block from earlier in the file.** The existing `extract_documentation` regex used a multi-line `$` anchor, which matched at every newline — allowing a POD block buried inside an earlier sub body to satisfy the anchor when the analyzer searched backwards from a later sub. Now anchored to true end-of-string (`\z`).

Both changes show up on the **editor-intelligence scorecard** (#4066): hover correctness for subroutine/method symbols, with no false-positive doc attribution.

## Why this matters

Hover docs are one of the highest-frequency LSP interactions. Before this PR:

- A `sub foo { =pod … =cut ... }` returned no documentation on hover even though the module author clearly documented the sub.
- A module with a single documented top-level `sub first { =pod ... =cut }` would cause every *following* undocumented sub in the same file to display `first`'s POD as its own hover — a silent but user-visible correctness bug in the pre-existing `extract_documentation` implementation. This was latent for any file with inline POD plus additional subs below it.

Fixing both together is necessary: without the regex fix, the new inline-POD fallback has nothing to fall back *to* for clean cases because the buggy backward search matches earlier POD first.

## Root cause

`crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs`

### 1. `extract_documentation` never searched inside bodies

```rust
pub(super) fn extract_documentation(&self, start: usize) -> Option<String> {
    // ...
    let before = &self.source[..start];   // only ever looks *backwards*
    // ...
}
```

For `Subroutine { body, .. }` and `Method { body, .. }` nodes we had the byte range of the body available but never consulted it.

### 2. The POD regex used `(?ms)...\s*$`

```rust
Regex::new(r"(?ms)(=[a-zA-Z0-9].*?\n=cut\n?)\s*$")
```

Under multi-line mode, `$` matches at *every* `\n`. So for a file like:

```perl
sub first {
=pod
Docs for first.
=cut
    return 1;
}
sub second { return 2; }
```

when looking backwards from `sub second`, the engine happily matched `=pod\n...\n=cut` followed by `\s*` (zero-width) followed by `$` at the newline immediately after `=cut\n` — three lines earlier — and returned `first`'s POD as `second`'s documentation.

### New helper: `extract_pod_in_range(SourceLocation) -> Option<String>`

Scans `source[range.start..range.end]` for the first column-0 POD directive (`=pod`, `=head1`, `=over`, etc.) and the matching `=cut`. Implemented as a small byte-level walker (not a regex) because:

- It needs explicit column-0 anchoring: POD is recognised by `perl-lexer` only when the `=` follows a newline or sits at position 0. A regex with `(?m)^` would get the column-0 check, but we'd still need manual range bounds handling, and the byte walker is straightforward and readable.
- It keeps the `OnceLock<Regex>` count stable (no new cached regex at startup).
- The hot path is bounded by body length, so a single linear byte scan is appropriate.

Wired in at three hover emission sites, using the `extract_documentation` preceding-doc result as the primary source and `extract_pod_in_range(body.location)` as the fallback:

- Named subroutines
- Anonymous subroutines (closures)
- Methods (`method foo { ... }`)

### Regex fix: `(?s)...\s*\z`

Swapped `(?ms)...\s*$` for `(?s)...\s*\z`. `\z` matches only true end-of-string regardless of flags, so the anchor now means what the original author intended: "POD block immediately before `start` with only whitespace in between." Dotall (`(?s)`) is preserved so `.*?` can span lines within a POD block.

## Behaviour matrix

| File shape | Before | After |
|---|---|---|
| `# comment \n sub foo { ... }` | `# comment` | `# comment` ✅ |
| `=pod ... =cut \n sub foo { ... }` | POD block | POD block ✅ |
| `sub foo { =pod Internal =cut; my $x = ...; }` | *no doc* | Inline POD surfaced ✅ |
| `# comment \n sub add { =pod Internal =cut; ... }` | `# comment` | `# comment` (preceding wins) ✅ |
| `sub first { =pod … =cut ... } sub second { ... }` | `second` inherits `first`'s POD 🐛 | `second` has no doc ✅ |
| `sub plain { my $x = ... }` | `None` | `None` ✅ (regression guard) |

## Test plan

New test file `crates/perl-semantic-analyzer/tests/pod_in_sub_body_hover_tests.rs` with six targeted cases:

- [x] `pod_inline_in_sub_body_surfaces_in_hover` — inline `=pod`/`=cut` attaches to its sub
- [x] `pod_head1_inside_sub_body_surfaces_in_hover` — same, with `=head1` directive
- [x] `preceding_comment_wins_over_inline_pod` — fallback precedence
- [x] `preceding_pod_wins_over_inline_pod` — fallback precedence (POD variant)
- [x] `sub_without_any_pod_has_no_doc` — no phantom docs on bare subs
- [x] `pod_in_body_does_not_bleed_to_later_sub` — regression guard for the regex bug

All pass; full `perl-semantic-analyzer` suite (≈775 tests across 25 binaries) is green, and `cargo build -p perl-lsp-rs` succeeds.

```bash
cargo test -p perl-semantic-analyzer --test pod_in_sub_body_hover_tests   # 6/6 pass
cargo test -p perl-semantic-analyzer                                      # all green
cargo clippy -p perl-semantic-analyzer --lib --test pod_in_sub_body_hover_tests   # clean
cargo fmt --check -p perl-semantic-analyzer                               # clean
cargo build -p perl-lsp-rs                                                # green
```

Pre-existing clippy `expect_used` errors in `tests/frameworks_web.rs` and a `map_or`→`is_none_or` warning in `tests/frameworks_catalyst.rs` are unrelated to this PR (verified via `git diff` — those files are unchanged here).

## Scope

Touches two files only:

- `crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs` — `extract_pod_in_range` helper, POD regex anchor fix, three hover sites get the fallback
- `crates/perl-semantic-analyzer/tests/pod_in_sub_body_hover_tests.rs` — new test file

No public API changes; the new helper is `pub(super)`. No dependency changes. No docs regeneration needed (metrics are computed post-merge).

## Related

- Closes #3407 ("POD inside subroutine bodies not associated with subroutine hover")
- Improves signal for the editor-intelligence scorecard (#4066) — hover correctness row

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/semantic/node_analysis.rs
@@ -138,7 +138,9 @@ impl SemanticAnalyzer {
 
                     let hover = HoverInfo {
                         signature: signature_str,
-                        documentation: self.extract_documentation(node.location.start),
+                        documentation: self
+                            .extract_documentation(node.location.start)
+                            .or_else(|| self.extract_pod_in_range(body.location)),
                         details: if attributes.is_empty() {
                             vec![]
                         } else {
@@ -173,7 +175,9 @@ impl SemanticAnalyzer {
 
                     let hover = HoverInfo {
                         signature: signature_str,
-                        documentation: self.extract_documentation(node.location.start),
+                        documentation: self
+                            .extract_documentation(node.location.start)
+                            .or_else(|| self.extract_pod_in_range(body.location)),
                         details,
                     };
 
@@ -205,7 +209,9 @@ impl SemanticAnalyzer {
                 // Add hover info
                 let hover = HoverInfo {
                     signature: format!("method {}", name),
-                    documentation: self.extract_documentation(node.location.start),
+                    documentation: self
+                        .extract_documentation(node.location.start)
+                        .or_else(|| self.extract_pod_in_range(body.location)),
                     details: if attributes.is_empty() {
                         vec![]
                     } else {
@@ -918,9 +924,17 @@ impl SemanticAnalyzer {
         }
         let before = &self.source[..start];
 
-        // Check for POD blocks ending with =cut
+        // Check for POD blocks ending with =cut immediately before `start`.
+        //
+        // The anchor must be true end-of-string (`\z`), not end-of-line: with
+        // multi-line mode `$` matches at every `\n`, which would allow a POD
+        // block buried earlier in the file to "leak" forward and attach to a
+        // later declaration that has no documentation of its own. `\z` makes
+        // the match specifically "POD block followed only by whitespace up to
+        // `start`". Dotall (`(?s)`) is still required so `.*?` can span lines
+        // within the POD block.
         let pod_re = POD_RE
-            .get_or_init(|| Regex::new(r"(?ms)(=[a-zA-Z0-9].*?\n=cut\n?)\s*$"))
+            .get_or_init(|| Regex::new(r"(?s)(=[a-zA-Z0-9].*?\n=cut\n?)\s*\z"))
             .as_ref()
             .ok()?;
         if let Some(caps) = pod_re.captures(before) {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as duplicate of #4403. Ensemble pattern — kept the best of the cluster per `feedback_codex_ensemble_pattern.md`.

PR #4557, opened 2026-04-22
fix(perl-module): avoid rename edits for package/string false positives (#4423)

EffortlessSteven:
### Motivation

- Resolve issue #4423 where module-rename planning produced false positives by rewriting `package ...;` declarations and matches inside quoted string literals, causing non-import lines to be mutated.

### Description

- Stop rewriting package-declaration lines by removing package-declaration handling from `plan_module_rename_edits` in `crates/perl-module/src/rename/mod.rs`.
- Harden qualified-call detection by adding a guard to `line_references_qualified_call` to skip `package` declarations and by introducing `is_within_quoted_string` to avoid matches inside single- or double-quoted literals.
- Apply the same package/quoted-string guards in `replace_module_name_prefix` so replacements are consistent with the predicate logic.
- Extend `crates/perl-module/tests/facade_api_completeness.rs` to assert negative cases for quoted strings and package declarations alongside the existing positive predicate tests.

### Testing

- Ran targeted crate tests: `cargo test -p perl-module --test module_rename_bdd --test facade_api_completeness --test rename_comprehensive_unit_tests` and the test suites passed (all tests in those suites succeeded).
- Ran full crate test `cargo test -p perl-module` and it completed with all tests passing.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e709df2ae08333ae450dfb4901b6d3)

### DIFF
--- a/crates/perl-module/src/rename/mod.rs
+++ b/crates/perl-module/src/rename/mod.rs
@@ -34,7 +34,6 @@ pub struct ModuleLineEdit {
 /// - Qualified function calls: `Module::Name::func()` → `NewName::func()`
 /// - Static method calls: `Module::Name->method()` → `NewName->method()`
 /// - `@ISA` array assignments
-/// - Package declarations: `package Module::Name;` → `package NewName;`
 ///
 /// Legacy package separators (`Foo'Bar`) are also handled.
 #[must_use]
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as duplicate of #4554. Both PRs make the same structural change (remove `package` rewrite from `plan_module_rename_edits`, add `package`-prefix guard). #4554 wins because it adds dedicated regression tests in `rename_comprehensive_unit_tests.rs` (not just facade tests), handles comment-tail false positives too (#4557's helper only covers quotes), and exports the helpers for direct unit testing.

PR #4605, opened 2026-04-23
fix(ci-hygiene): ignore TODO/FIXME tokens inside Rust string literals (#9999)

EffortlessSteven:
### Motivation
- Reduce false positives from the TODO/FIXME scanner in `perl-ci-hygiene` where `//` and `/*` markers appearing inside Rust string literals (normal and raw) were being treated as comments.

### Description
- Updated `has_unlinked_todo_in_rust_line` to skip `//` and `/*` matches that occur inside Rust string literals.
- Added `is_within_rust_string_literal` and `raw_string_span_at` helpers to detect normal quoted strings and raw `r#"..."#` strings.
- Added a unit test `rust_todo_detection_ignores_embedded_comment_markers_in_string_literals` validating embedded `// TODO` and `/* FIXME */` inside string payloads are ignored while real trailing comments are still detected.

### Testing
- Ran targeted unit tests in the `perl-ci-hygiene` crate: `rust_todo_detection_ignores_linked_or_url_like_comments`, `rust_todo_detection_ignores_raw_string_comment_markers`, and `rust_todo_detection_ignores_embedded_comment_markers_in_string_literals`, and they all passed.
- Ran `cargo xtask fmt` to ensure formatting hygiene and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9691647e8833385be82c5378f23da)

PR #4607, opened 2026-04-23
fix(ci-hygiene): ignore TODO tokens inside quoted hash content (#0000)

EffortlessSteven:
### Motivation
- Avoid false positives where `#` appears inside single- or double-quoted text or when the hash is escaped, so TODO/FIXME detection for hash-comment languages is more accurate.
- Preserve existing shebang (`#!`) behavior and whitespace-delimited inline comment semantics.

### Description
- Replace naive `line.find('#')` logic in `has_unlinked_todo_in_hash_line` with a call to a new helper `find_unquoted_hash_comment_start` that finds the first unquoted, unescaped `#` index.
- Add `find_unquoted_hash_comment_start` which tracks single-quote, double-quote, and escape state to ignore `#` characters inside quoted strings and skip escaped `\#` sequences.
- Keep shebang handling by short-circuiting on lines starting with `#!`.
- Expand the existing unit test `hash_comment_todo_detection_handles_shebang_and_inline_hashes` with assertions covering quoted `#` content and escaped-hash cases.

### Testing
- Ran `cargo xtask fmt` successfully to apply formatting changes.
- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` and the targeted test passed.
- Ran `cargo test -p perl-ci-hygiene` where most tests passed (34 passed) and one existing snapshot test failed (`checked_in_pre_push_hook_matches_generated_hook`) due to checked-in pre-push hook drift unrelated to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9691647e8833385be82c5378f23da)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3863,17 +3863,58 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 }
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
-    if let Some(idx) = line.find('#') {
-        if idx > 0 && line.as_bytes()[idx - 1] == b'!' {
-            return false;
-        }
-        if idx > 0 && !line[..idx].chars().next_back().is_some_and(char::is_whitespace) {
-            return false;
-        }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #4623, opened 2026-04-23
fix(ci-hygiene): detect TODOs after quoted # in hash-comment files (#0000)

EffortlessSteven:
### Motivation
- Fix a false-negative in the TODO/FIXME scanner for hash-comment languages (shell, Perl `#` comments, `Justfile`, etc.) where a `#` character appearing inside quoted strings could prevent detection of a later real comment containing `TODO`/`FIXME`.

### Description
- Update `has_unlinked_todo_in_hash_line` to iterate over every `#` occurrence using `match_indices` instead of only inspecting the first `#` on the line.
- Preserve existing heuristics for shebang (`#!`) and inline non-comment `#` detection and continue scanning rather than returning early. 
- Add a regression test for a mixed-content line (`echo "# not a comment" # TODO: follow up`) to ensure the trailing `TODO` is detected.
- Run formatting via `cargo xtask fmt` as part of the change.

### Testing
- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` and the test passed (`1 passed; 0 failed`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e96a035ae88333b171e72e1ecdeda7)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3863,17 +3863,21 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 }
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
-    if let Some(idx) = line.find('#') {
+    for (idx, _) in line.match_indices('#') {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #4625, opened 2026-04-23
fix(ci-hygiene): handle quoted hash literals in TODO scan (#9999)

EffortlessSteven:
### Motivation

- The TODO/FIXME scanner for hash-style comments used `line.find('#')`, which could hit `#` characters inside string literals or escaped sequences and miss the real trailing comment, producing false negatives.
- Improve robustness so the scanner inspects only real hash comments outside quoted or escaped contexts.

### Description

- Replace naive `line.find('#')` usage in `has_unlinked_todo_in_hash_line` with a parser-aware helper `find_hash_comment_start` that tracks single/double-quoted spans and escaped characters and returns the real comment start index.
- Update `has_unlinked_todo_in_hash_line` to use `find_hash_comment_start` and keep existing shebang/inline-hash handling.
- Add unit tests covering `#` inside double-quoted strings, a trailing real `#` comment after a string literal, and an escaped `\#` sequence to prevent regressions, and run `cargo xtask fmt` for formatting.

### Testing

- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` and the test passed.
- Ran `cargo test -p perl-ci-hygiene` where all modified-related tests passed but the full suite failed due to an unrelated existing test `checked_in_pre_push_hook_matches_generated_hook` mismatch (pre-existing hook content sync test), not caused by this change.

------

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3863,17 +3863,49 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 }
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
-    if let Some(idx) = line.find('#') {
-        if idx > 0 && line.as_bytes()[idx - 1] == b'!' {
-            return false;
-        }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #4626, opened 2026-04-23
fix(perl-ci-hygiene): parse hash comments past quoted markers (#0000)

EffortlessSteven:
### Motivation

- The TODO/FIXME scanner for hash-comment files incorrectly treated the first `#` as a comment start even when that `#` was inside a quoted string, causing false-negatives for real comments later on the same line.

### Description

- Replace the simple `line.find('#')` approach in `has_unlinked_todo_in_hash_line` with a line walker `find_hash_comment_start` that tracks single/double-quoted state and escapes to find the true comment start.
- Preserve existing heuristics for shebang (`#!`) and inline `#` forms that are not comment starts by keeping the same checks after locating a candidate `#`.
- Add a regression unit test `hash_comment_todo_detection_skips_hash_in_quotes_and_finds_real_comment` to ensure quoted `#` characters are ignored and real trailing comments are detected.
- Change is contained in `crates/perl-ci-hygiene/src/main.rs` and is a small targeted improvement to TODO detection logic.

### Testing

- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection -- --nocapture` and the relevant tests passed.
- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection_skips_hash_in_quotes_and_finds_real_comment -- --nocapture` and the new regression test passed.

------

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3863,17 +3863,55 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 }
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
-    if let Some(idx) = line.find('#') {
+    let Some(idx) = find_hash_comment_start(line) else {
+        return false;
+    };
+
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #4627, opened 2026-04-23
test(perl-workspace): de-flake workspace member collapse guard (#0000)

EffortlessSteven:
### Motivation

- The existing test relied on brittle absolute workspace-member counts and began failing as crates were consolidated, so the guard needed to be made relational to avoid false negatives.

### Description

- Replace hard lower/upper bound assertions with parsing of `[workspace].members` from the root `Cargo.toml` and a relational check against `[workspace.metadata.publish].allow` in `crates/perl-workspace-index/tests/collapse_edge_cases_tests.rs`.
- Ensure both the workspace members set and the publish allowlist are non-empty, and verify every allowlisted package is represented by a workspace member while handling intentional name/path shims (`perl-workspace-index -> perl-workspace`, `perl-lsp -> perl-lsp-rs`).
- Keep the test focused on semantic correspondence between publishable crates and workspace membership rather than fragile numeric expectations.

### Testing

- Ran formatter: `cargo xtask fmt` which completed successfully.
- Ran the focused test: `cargo test -p perl-workspace --test collapse_edge_cases_tests --quiet` which passed all tests.
- Ran the workspace package tests: `cargo test -p perl-workspace --quiet` which passed (no failures reported).

------

### DIFF
--- a/crates/perl-workspace-index/tests/collapse_edge_cases_tests.rs
+++ b/crates/perl-workspace-index/tests/collapse_edge_cases_tests.rs
@@ -297,7 +297,7 @@ fn test_cargo_toml_no_deleted_satellite_deps() {
 ///
 /// Wave A (#4426) removed 6 perl-workspace-* satellites.
 /// Wave D (#4454) removed 13+ parser/AST satellites into perl-parser-core and perl-parser.
-/// Combined baseline was ~120, now ~83 members.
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

Winner: **#4640** (most principled — replaces brittle numeric threshold with on-disk `crates/` directory comparison). This PR tightens the parser but keeps the ≥70 floor which would still fail on the post-collapse workspace.

PR #4637, opened 2026-04-23
fix(ci-hygiene): avoid quoted-string TODO false positives in hash scanners (#4587)

EffortlessSteven:
### Motivation

- Prevent false-positive TODO/FIXME detections for hash-comment languages when a `#` appears inside a quoted string literal (common in shell/Perl snippets).

### Description

- Updated `has_unlinked_todo_in_hash_line` in `crates/perl-ci-hygiene/src/main.rs` to skip `#` markers that are inside single or double quoted strings. 
- Added a helper `is_inside_single_or_double_quotes(line: &str, idx: usize) -> bool` that tracks single/double quote state and handles escapes when scanning up to the `#` position. 
- Extended unit tests in the same file to cover embedded `# TODO`/`# FIXME` inside both double- and single-quoted strings while preserving existing comment detection behavior.

### Testing

- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` which passed. 
- Ran `cargo test -p perl-ci-hygiene` where the new tests pass but the full crate test run reports one pre-existing unrelated failure (`checked_in_pre_push_hook_matches_generated_hook`) and so the package-level test suite returns a failing status unrelated to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e969265fc48333ab0ad53da8f9e39e)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3864,6 +3864,9 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
     if let Some(idx) = line.find('#') {
+        if is_inside_single_or_double_quotes(line, idx) {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #4638, opened 2026-04-23
fix(dap): align inlineValues legacy filtering and bounds semantics (#9999)

EffortlessSteven:
### Motivation
- Prevent surprising inline value responses when a DAP request's `start_line` is beyond the file length by returning no values instead of clamping to the last line.
- Make the legacy scalar-only inline value path consistent with the runtime-enabled path by removing special Perl variables and deduplicating repeated variables per line to reduce noisy editor annotations.

### Description
- Tighten `normalize_line_bounds` in `crates/perl-dap/src/inline_values.rs` to return `None` when `start_line` is greater than the file's line count instead of clamping to the final line.
- Update `collect_inline_values` (legacy scalar path) to exclude special variables via `is_special_variable_name` and to deduplicate variables per-line using a `HashSet`.
- Add unit tests in `crates/perl-dap/src/inline_values.rs` covering out-of-range start lines, legacy per-line deduplication, and legacy special-variable exclusion.
- Run repository formatting with `cargo xtask fmt` on the modified crate.

### Testing
- Ran `cargo xtask fmt` to format changes, which completed successfully.
- Ran focused unit tests with `cargo test -p perl-dap inline_values::tests::`, and the inline-values tests passed (all added and existing tests green).
- Verified the fix resolved a previously failing `test_inline_values_out_of_range` in the `perl-dap` test set and observed the `perl-dap` test suite complete with no failures for the affected tests.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e969265fc48333ab0ad53da8f9e39e)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -45,6 +45,9 @@ fn normalize_line_bounds(
 
     let start_1_based = start_line.max(1) as usize;
     let end_1_based = end_line.max(1) as usize;
+    if start_1_based > line_count {
+        return None;
+    }
 
     let start_idx = start_1_based.saturating_sub(1).min(line_count.saturating_sub(1));
     let end_idx = end_1_based.saturating_sub(1).min(line_count.saturating_sub(1));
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

Winner: **#4643** (+109, most comprehensive: rejects both line bounds past line count + adds full `build_code_mask` byte-mask + filters non-code matches). This PR is a narrower subset of the same bounds/code-mask fix.

PR #4643, opened 2026-04-23
fix(dap): tighten inline value extraction around non-code tokens (#4650)

EffortlessSteven:
### Motivation
- Reduce noisy or incorrect inline-value hints by preventing detection of variable-like tokens inside comments and quoted strings.
- Make lightweight regex-based extraction behave consistently with coverage expectations for out-of-range line requests.

### Description
- Added a byte-level mask builder `code_byte_mask` to identify comment and quoted-string regions and ignore matches inside them in `crates/perl-dap/src/inline_values.rs`.
- Applied the mask to `extract_variable_names`, `collect_inline_values_with_runtime`, and `collect_inline_values` to preserve byte-accurate column positions while filtering non-code tokens.
- Tightened `normalize_line_bounds` to return `None` for requested lines that are beyond the available `line_count` while still clamping non-positive inputs to line 1.
- Added focused tests `test_variable_like_tokens_in_strings_and_comments_are_ignored` and `test_out_of_range_line_bounds_return_empty` to cover string/comment filtering and out-of-range behavior.

### Testing
- Ran unit tests for the inline-values module with `cargo test -p perl-dap inline_values::tests::` and all tests passed.
- Ran the targeted coverage test with `cargo test -p perl-dap --test dap_coverage_audit_tests test_inline_values_out_of_range` and the test passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9691bc760833392c8cbdb038b7372)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -34,7 +34,7 @@ fn is_special_variable_name(name: &str) -> bool {
 /// Convert 1-based line bounds into inclusive 0-based indexes.
 ///
 /// Non-positive line inputs are clamped to line 1, and the end index is
-/// clamped to the final available line.
+/// rejected when either bound is past the available line count.
 fn normalize_line_bounds(
     start_line: i64,
     end_line: i64,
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Rebased on master — resolved conflict with sibling cluster merges.

Conflict was in the test section of `inline_values.rs` — #4613 added `test_runtime_inline_value_for_legacy_namespaced_scalar` in the same location where this PR added `test_variable_like_tokens_in_strings_and_comments_are_ignored` and `test_inline_values_ignore_strings_and_comments`. Resolution: kept both test functions (distinct, non-overlapping). The `code_byte_mask` helper and strict bounds rejection are fully compatible with #4613's legacy-sigil regex change. Compiled clean.

PR #4646, opened 2026-04-23
fix(dap): ignore non-code inline value matches (#0000)

EffortlessSteven:
### Motivation
- Inline-value extraction produced spurious matches for variables that only appear inside quoted strings or trailing `#` comments, leading to misleading inline values and incorrect DAP columns. 
- Regex matching must operate on code-only regions while preserving byte offsets so reported columns stay correct for DAP consumers. 
- Requests whose `start_line` is beyond the file were being clamped to the last line, which is surprising for out-of-range queries and should return no results.

### Description
- Add `mask_non_code_regions` which masks quoted single/double strings and trailing `#` comments while preserving byte offsets for correct capture indexing. 
- Use the masking function before regex matching in `extract_variable_names`, `collect_inline_values_with_runtime`, and legacy `collect_inline_values`, and map captures back into the original line using capture byte indices. 
- Update `normalize_line_bounds` so that a `start_line` beyond `line_count` returns `None` (empty result) instead of clamping to the last line. 
- Add unit tests `test_inline_values_ignore_comments_and_strings` and `test_start_line_beyond_source_returns_empty` to cover the new behaviors.

### Testing
- Ran formatting with `cargo xtask fmt`, which completed successfully. 
- Ran focused inline-value tests with `cargo test -p perl-dap inline_values`, and they all passed. 
- Ran the coverage audit test case with `cargo test -p perl-dap --test dap_coverage_audit_tests test_inline_values_out_of_range -- --nocapture`, which passed and confirmed the out-of-range behavior fix.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e96919a58c8333883e2538091d8fde)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -45,11 +45,70 @@ fn normalize_line_bounds(
 
     let start_1_based = start_line.max(1) as usize;
     let end_1_based = end_line.max(1) as usize;
+    if start_1_based > line_count {
+        return None;
+    }
 
     let start_idx = start_1_based.saturating_sub(1).min(line_count.saturating_sub(1));
     let end_idx = end_1_based.saturating_sub(1).min(line_count.saturating_sub(1));
 
     (start_idx <= end_idx).then_some((start_idx, end_idx))
+}
+
+/// Mask quoted strings and trailing `#` comments while preserving byte offsets.
+///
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

Winner: **#4643** (+109, most comprehensive: rejects both line bounds past line count + adds full `build_code_mask` byte-mask + filters non-code matches). This PR is a narrower subset of the same bounds/code-mask fix.

Issue #4512, opened 2026-04-19
fix(hook): pre-push cargo fmt uses dir basename instead of Cargo.toml package name


EffortlessSteven:
## Summary

Pre-push hook (`cargo xtask fmt --check -p <name>`) infers the package name from the crate directory basename, which is wrong for `crates/perl-lsp/` (package name is `perl-lsp-rs`, not `perl-lsp`). The command fails, the hook rejects the push, and builders bypass with `--no-verify` — losing the fmt-check guarantee.

## Evidence

Observed 2× on 2026-04-19 collapse session:
- G1a builder push (commit series)
- G1b builder fix push (commit `d9ecd697c`, PR #4510)

Both cited the same reason: `cargo fmt -p perl-lsp` fails because no package named `perl-lsp` exists in the workspace.

## Impact

- Pre-push fmt gate silently bypassed — relies on CI to catch fmt issues instead of local.
- Trains agents to reach for `--no-verify` when the hook misbehaves, which is exactly the pattern we warn against in memory (`feedback_pre_push_hook_windows_race.md`).
- Root cause conflated with Windows MAX_PATH complaints — muddies diagnosis of the real MAX_PATH bug (#4456).

## Root cause

Hook uses directory basename as the package name. For `crates/perl-lsp/` the actual package name is `perl-lsp-rs`. Every other crate in the workspace has matching dir and package name, so the hook works elsewhere — masking the bug.

## Fix

Replace basename inference with `cargo metadata --no-deps --format-version 1` lookup:

```bash
# Pseudo-Bash / adjust for the actual hook language
# Old:
PKG_NAME="$(basename "$CRATE_DIR")"

# New:
PKG_NAME="$(cargo metadata --no-deps --format-version 1 \
  | jq -r --arg dir "$CRATE_DIR" \
    '.packages[] | select(.manifest_path | startswith($dir + "/")) | .name' \
  | head -1)"

# Fallback if cargo metadata not available (shouldn't happen):
[[ -z "$PKG_NAME" ]] && PKG_NAME="$(basename "$CRATE_DIR")"
```

Implementation location: the Rust-based hook runner in `xtask/src/tasks/fmt.rs` (or wherever `xtask fmt` resolves package names). Use `cargo_metadata` crate directly rather than shelling out.

## Acceptance

[...]

PR #4648, opened 2026-04-23
fix(ci-hygiene): sync single-crate hook package resolution (#4512)

EffortlessSteven:
### Motivation
- The generated pre-push hook produced by the `perl-ci-hygiene` generator diverged from the checked-in `hooks/pre-push`, causing the single-crate gate to assume crate directory names equal Cargo package names and break targets like `crates/perl-lsp` (actual package `perl-lsp-rs`).
- The goal is to make generated hook output match the checked-in script and reliably target the correct Cargo package for per-crate `fmt/clippy/test` runs.

### Description
- Updated the generated hook template in `crates/perl-ci-hygiene/src/main.rs` to compute `SINGLE_CRATE_DIR` and resolve the package name via `cargo xtask resolve-package-name "crates/${SINGLE_CRATE_DIR}"` with a fallback to the directory name. 
- Changed the banner to print both the crate directory and resolved package name (`${SINGLE_CRATE_DIR} -> ${SINGLE_CRATE_NAME}`) for clearer diagnostics when the single-crate gate runs. 
- This mirrors the logic present in the checked-in `hooks/pre-push` and preserves correct targeted gate behavior for crates whose package name differs from their directory.

### Testing
- Ran `cargo test -p perl-ci-hygiene` and the full test suite for that package passed (`35 passed; 0 failed`).
- Ran code formatting with `cargo xtask fmt` which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e96919a58c8333883e2538091d8fde)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -1984,8 +1984,12 @@ fi
 # Falls back to the full gate if classification is ambiguous.
 SINGLE_CRATE_COUNT="$(printf '%s' "$SINGLE_CRATE_NAMES" | grep -c . 2>/dev/null || echo 0)"
 if [ "$SINGLE_CRATE_ALL_UNDER_CRATES" = true ] && [ "$SINGLE_CRATE_COUNT" = "1" ]; then
-    SINGLE_CRATE_NAME="$(printf '%s' "$SINGLE_CRATE_NAMES" | tr -d '[:space:]')"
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as BYTE-IDENTICAL duplicate of #4621 (same index hash `fd7c1c627..143b2fb84`). Both fix single-crate pre-push hook to use Cargo package name instead of directory basename.

Issue #4650, opened 2026-04-23
fix(dap): inline-value and stack-frame edge cases


EffortlessSteven:
Bundle: 4 independent DAP bugs from the Codex batch of 2026-04-23.

### 1. Variable paging cache + child refs (#4608)
Variable pages and child references go stale under certain reuse patterns.

### 2. Legacy namespace sigils in PERL_VAR_RE (#4613)
Perl `'` (single-quote) namespace separator not matched in inline-value regex.

### 3. Stack-frame hash heuristic (#4614)
`stack/parser.rs` misclassifies frames; net -20 lines, removes faulty heuristic.

### 4. Inline-value bounds + non-code masking (#4643)
Rejects both line bounds past line count, adds `build_code_mask` byte-mask, filters non-code matches. Comprehensive fix.

Each PR is a distinct fix for a different DAP bug surface.

EffortlessSteven:
## DAP Edge-Case Hardening Analysis

I've analyzed the codebase for the DAP (Debug Adapter Protocol) edge-case hardening issue.

### Issue Summary
PR #4636 hardens DAP handlers for:
1. Malformed launch.json configurations
2. Missing or invalid adapter paths
3. Edge cases in breakpoint resolution

### Key Locations
- DAP implementation: 
- DAP handlers:  or similar
- Launch configuration: 

### Implementation Guidance
To harden DAP edge cases:
1. **Malformed launch.json**: Add strict validation with clear error messages
2. **Invalid adapter paths**: Verify paths exist before use, provide helpful errors
3. **Breakpoint edge cases**: Handle:
   - Breakpoints in non-existent files
   - Invalid line numbers
   - Duplicate breakpoints
   - Conditional breakpoint syntax errors

### Files to Modify
-  (request handlers)
-  (launch configuration)
-  (breakpoint management)

### Testing
Add tests in  covering malformed inputs

Would you like me to locate the specific DAP handler code?

EffortlessSteven:
All 4 referenced DAP PRs merged and deployed: #4608, #4613, #4614, #4643 all MERGED on 2026-04-23. Bundle is complete. (independently verified; disposition EffortlessMetrics/perl-lsp-swarm#1731)

PR #4660, opened 2026-04-23
fix(perl-dap): avoid false inline matches in comments/strings (#9999)

EffortlessSteven:
### Motivation
- Reduce noisy/incorrect DAP inline variable matches that occur when regexes hit variables inside comments or quoted strings while preserving accurate column offsets. 
- Ensure line-range handling returns empty when the requested start line is beyond the file instead of silently clamping to the last line.

### Description
- Add `mask_comments_and_strings` to blank out comment and simple quoted-string regions while preserving original byte offsets. 
- Apply masking to `extract_variable_names`, `collect_inline_values_with_runtime`, and legacy `collect_inline_values` so regex scans ignore variables inside comments/strings. 
- Tighten `normalize_line_bounds` to return `None` when `start_line` is beyond `line_count`, yielding empty results for out-of-range ranges. 
- Add unit tests `test_extract_ignores_comments_and_strings` and `test_collect_inline_values_ignores_comments_and_strings` covering the new masking behavior.

### Testing
- Ran targeted unit tests `cargo test -p perl-dap inline_values::tests::test_collect_inline_values_ignores_comments_and_strings` which passed. 
- Ran the range edge-case test `cargo test -p perl-dap test_inline_values_out_of_range` which passed. 
- The new and existing inline-value tests in `crates/perl-dap` covering array/hash/scalar formatting and deduplication passed locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e969fcbd9483338d2e7a32a11d799c)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -30,6 +30,66 @@ fn is_special_variable_name(name: &str) -> bool {
     SPECIAL_VARS.contains(&name)
 }
 
+/// Mask Perl comments and simple quoted strings with spaces so regex scans
+/// keep original byte offsets while avoiding obvious false positives.
+fn mask_comments_and_strings(line: &str) -> String {
+    #[derive(Clone, Copy, PartialEq, Eq)]
+    enum State {
+        Normal,
+        SingleQuoted,
+        DoubleQuoted,
+        Comment,
+    }
+
+    let mut state = State::Normal;
+    let mut escaped = false;
+    let mut masked = String::with_capacity(line.len());
+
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of #4643 which merged this session. Both PRs address the same scope — avoiding false inline-value matches in DAP comments/strings — via a byte-level mask approach. #4643's implementation (`build_code_mask` + strict bounds) was selected as the winner from the DAP cluster triage (see tracking issue #4650). If there's a specific case that #4643's mask doesn't cover, file a focused follow-up issue rather than re-implementing the mask.

Issue #4649, opened 2026-04-23
fix(ci-hygiene): TODO scanner false positives and ignore-reason classification


EffortlessSteven:
Bundle: 5 independent ci-hygiene bugs from the Codex batch of 2026-04-23.

### 1. TODO/FIXME inside Rust string literals (#4620)
The scanner classifies `//` and `/*` inside `r#"..."#` or `"..."` as comment starts and false-positives on TODOs in them. Fix: `is_index_in_rust_literal` helper with raw-hash handling.

### 2. Hash-comment scanner treats `#` inside quoted strings as comment start (#4606)
`echo '# TODO in string'` false-positives. Fix: `find_hash_comment_start` helper, preserves shebang guard.

### 3. Hash scanner misses TODO after in-string `#` (#4609)
Current `line.find('#')` stops at first match; need to iterate all `#` occurrences. Fix composes with #4606.

### 5. Ignore-reason classifier: broad `"ac"` substring matches "cache"/"placeholder" (#4629)

Each PR listed above is a distinct fix. This issue bundles them for traceability.

EffortlessSteven:
## CI-Hygiene TODO Scanner Analysis

I've analyzed the codebase for the CI-hygiene TODO scanner issue.

### Issue Summary
PR #4637 adds a CI-hygiene check to fail CI on committed TODO/FIXME comments that reference closed issues.

### Key Locations
- CI-hygiene crate: 
- CI workflows:  or similar
- TODO scanner: Likely in 

### Implementation Guidance
To implement the TODO scanner:
1. **Scan for TODO/FIXME patterns**: Regex matching  or 
2. **GitHub API integration**: Query issue status for referenced numbers
3. **CI integration**: Add as a step in CI workflow, fail on closed references
4. **Exclusions**: Allow marking intentional TODOs (e.g., )

### Files to Modify
-  (new or existing)
-  (add scanner step)
- Add tests in 

### Example Implementation Pattern


Would you like me to check the existing CI-hygiene structure?

EffortlessSteven:
All 5 referenced PRs merged and deployed: #4620, #4606, #4609, #4622, #4629 all MERGED on 2026-04-23. CI-hygiene bundle is complete. (independently verified; disposition EffortlessMetrics/perl-lsp-swarm#1731)

PR #4664, opened 2026-04-23
fix(ci-hygiene): harden hash-comment TODO scanning (#6123)

EffortlessSteven:
### Motivation

- The TODO/FIXME baseline scanner currently treats `#` inside quoted strings as comment starts, producing false positives for shell-style lines like `echo "# TODO in string"`.
- The goal is to only flag unlinked TODO/FIXME markers that appear in real comments, preserving shebang handling and existing behavior for inline/trailing comments.

### Description

- Replace the simple `line.find('#')` logic in `has_unlinked_todo_in_hash_line` with a robust scanner `find_hash_comment_start` that tracks single-quote, double-quote, and escape state to locate real comment-starting `#` characters outside quoted strings.
- Preserve the existing shebang behavior so lines starting with `#!` remain ignored by the scanner.
- Simplify `has_unlinked_todo_in_hash_line` to use the new scanner and delegate token checking to `has_unlinked_token` as before.
- Add a focused regression test `hash_comment_todo_detection_ignores_hashes_in_quoted_strings` to cover `"# TODO in string"`, `'# FIXME in string'`, and mixed string+comment cases.

### Testing

- Ran the focused test: `cargo test -p perl-ci-hygiene hash_comment_todo_detection -- --nocapture`, which passed.
- Ran formatting: `cargo xtask fmt`, which completed successfully.
- Ran the full crate test: `cargo test -p perl-ci-hygiene`, which revealed an existing unrelated test failure (`checked_in_pre_push_hook_matches_generated_hook`) and thus the full-suite run failed; the failure is pre-existing and not caused by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e969fe3c9c83339011b5c42be811a6)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3863,17 +3863,49 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 }
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
-    if let Some(idx) = line.find('#') {
-        if idx > 0 && line.as_bytes()[idx - 1] == b'!' {
-            return false;
-        }
-        if idx > 0 && !line[..idx].chars().next_back().is_some_and(char::is_whitespace) {
-            return false;
-        }
-        has_unlinked_token(&line[idx + 1..], token_re)
-    } else {
-        false
+    let Some(idx) = find_hash_comment_start(line) else {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as already-fixed on master. The hash-comment TODO false-positive family was addressed this session by #4606 (hash-comment quoted strings), #4609 (scan all hash comments), and #4620 (Rust string literal false positives). See tracking issue #4649.

PR #4669, opened 2026-04-23
fix(ci-hygiene): improve hash TODO scanning after quoted literals (#0000)

EffortlessSteven:
### Motivation

- The TODO/FIXME scanner treated only the first `#` on a line as a potential comment, which missed real unlinked TODOs when an earlier `#` occurred inside a quoted literal. This produced false negatives in hygiene checks for shell/Perl-like files.

### Description

- Updated `has_unlinked_todo_in_hash_line` to iterate all `#` occurrences on a line and consider each as a potential comment start instead of short-circuiting on the first `#` (`crates/perl-ci-hygiene/src/main.rs`).
- Added `is_hash_comment_start` to encapsulate shebang/inline rules and whitespace checks before a `#` is treated as a comment start.
- Added `is_inside_simple_quote_context` to avoid treating `#` characters that appear inside single- or double-quoted literals as comment starts while still allowing later real comments on the same line.
- Extended unit coverage with tests asserting behavior when the first `#` is inside quotes and a trailing `#` introduces the actual comment.

### Testing

- Ran the targeted test: `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` and it passed.
- Ran formatting: `cargo fmt -p perl-ci-hygiene` and it succeeded.
- Ran the full crate test suite: `cargo test -p perl-ci-hygiene` which ran and showed the change's tests passing, but the suite failed on a pre-existing hook-sync test (`checked_in_pre_push_hook_matches_generated_hook`) unrelated to this change (test indicates checked-in hooks file drift).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e969ffb010833393f06a7cbe8ad342)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3863,17 +3863,54 @@ fn has_unlinked_todo_in_rust_line(line: &str, token_re: &Regex) -> bool {
 }
 
 fn has_unlinked_todo_in_hash_line(line: &str, token_re: &Regex) -> bool {
-    if let Some(idx) = line.find('#') {
-        if idx > 0 && line.as_bytes()[idx - 1] == b'!' {
-            return false;
+    for (idx, _) in line.match_indices('#') {
+        if !is_hash_comment_start(line, idx) {
+            continue;
         }
-        if idx > 0 && !line[..idx].chars().next_back().is_some_and(char::is_whitespace) {
-            return false;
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as already-fixed on master. The hash-comment TODO false-positive family was addressed this session by #4606 (hash-comment quoted strings), #4609 (scan all hash comments), and #4620 (Rust string literal false positives). See tracking issue #4649.

PR #4690, opened 2026-04-23
fix(parser): normalize heredoc anti-pattern location mapping (#4707)

EffortlessSteven:
### Motivation

- Several heredoc anti-pattern detectors computed `column` using `start - last_newline_index`, which produced inconsistent (effectively one-based) columns for matches on non-first lines, causing misaligned locations in tooling.

### Description

- Add a shared helper `location_from_start(code: &str, offset: usize, start: usize) -> Location` to compute `line`, zero-based `column`, and `offset` consistently from a match start.
- Replace duplicated inline location calculations across all detectors (format, `BEGIN`-time, dynamic delimiter, source filter, regex code block, eval string, tied-handle) with calls to `location_from_start`.
- Add a regression unit test `test_location_column_is_zero_based_for_new_line_matches` that asserts a second-line match reports `line = 1`, `column = 0`, and `offset = 11`.

### Testing

- Ran formatting with `cargo xtask fmt` and `cargo fmt -p perl-parser`, which completed successfully.
- Ran the targeted tests with `cargo test -p perl-parser test_location_column_is_zero_based_for_new_line_matches -- --nocapture`, which passed.
- Ran the heredoc anti-pattern test suite with `cargo test -p perl-parser heredoc_anti_patterns`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97c897e8483338b6f73944beb172d)

### DIFF
--- a/crates/perl-parser/src/heredoc_anti_patterns.rs
+++ b/crates/perl-parser/src/heredoc_anti_patterns.rs
@@ -55,6 +55,16 @@ trait PatternDetector: Send + Sync {
     fn diagnose(&self, pattern: &AntiPattern) -> Option<Diagnostic>;
 }
 
+fn location_from_start(code: &str, offset: usize, start: usize) -> Location {
+    let line = code[..start].lines().count();
+    let column = match code[..start].rfind('\n') {
+        Some(last_newline) => start.saturating_sub(last_newline + 1),
+        None => start,
+    };
+
+    Location { line, column, offset: offset + start }
+}
+
 // Format heredoc detector
 struct FormatHeredocDetector;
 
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Rebased on master — resolved conflict with sibling #4707 bundle merges.

Conflict was with #4716 (robust BEGIN heredoc detection) which changed BeginTimeHeredocDetector to use `BEGIN_BLOCK_START_PATTERN`/`find_matching_brace`. Resolution: kept master's superior brace-matching logic and applied `location_from_start` to it. Also kept master's `find_iter` loop in TiedHandleDetector (from #4668) and applied `location_from_start` there too. Both master's `test_tied_handle_reports_multiple_writes` and this PR's `test_location_column_is_zero_based_for_new_line_matches` test are preserved. Compiles clean (`cargo check --package perl-parser --tests`).

PR #4700, opened 2026-04-23
fix(ci-hygiene): avoid false TODO hits in escaped single-quoted strings (#4707)

EffortlessSteven:
### Motivation

- The TODO scanner for hash-comment languages could misclassify `# TODO` inside single-quoted literals that contain escaped apostrophes as real comments, producing false positives.

### Description

- Update `find_hash_comment_start` in `crates/perl-ci-hygiene/src/main.rs` to honor escaped single quotes (`\'`) while inside single-quoted literals so an escaped apostrophe does not prematurely terminate the single-quote state.
- Reset the escape state when entering a single- or double-quoted literal to avoid leaking prior escape state between contexts.
- Add regression assertions to the existing `hash_comment_todo_detection_handles_shebang_and_inline_hashes` test to cover both the no-comment and trailing-comment cases for `print 'it\'s # TODO in string';`.

### Testing

- Ran formatter with `cargo xtask fmt` and it completed successfully.
- Ran the targeted test with `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` and it passed.
- Ran the crate test suite with `cargo test -p perl-ci-hygiene` and all tests passed (36 passed, 0 failed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97c8ecc78833383ee42832b86263d)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3908,6 +3908,14 @@ fn find_hash_comment_start(line: &str, perl_mode: bool) -> Option<usize> {
 
     for (idx, ch) in line.char_indices() {
         if in_single {
+            if prev_was_escape {
+                prev_was_escape = false;
+                continue;
+            }
+            if ch == '\\' {
+                prev_was_escape = true;
+                continue;
+            }
             if ch == '\'' {
                 in_single = false;
             }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Rebased on master — resolved conflict with sibling #4707 bundle merges.

Single conflict in `find_hash_comment_start`: the `match ch` arm for `'\'' ` and `'"'` now uses block form with `prev_was_escape = false` resets (this PR's fix), while preserving the `'`' => in_backtick = true` arm (unchanged by this PR). The `in_single` escape-handling block (lines 3911-3918) was already in master and applied cleanly. Tests for escaped single-quoted strings preserved. Compiles clean (`cargo check --package perl-ci-hygiene --tests`).

Issue #4707, opened 2026-04-23
fix: Codex batch 4 (2026-04-23) — 15 distinct fixes across ci-hygiene, dap, parser, workspace, color


EffortlessSteven:
Bundle tracking issue for 15 winners from the 20-PR Codex batch of 2026-04-23 02:24-02:29. 5 duplicates closed separately.

## CI-hygiene (8 bugs)

- **#4684** — ignore TODO markers inside backtick literals (new `in_backtick` state in `find_hash_comment_start`)
- **#4685** — detect inline Perl TODO comments (Perl-specific path allowing `#` without preceding whitespace per `print# comment` being legal)
- **#4686** — handle C-string TODO parsing edge cases (extends `is_index_in_rust_literal` for `c"…"` / `cr#"…"#`)
- **#4687** — detect TODO hash-comments after `&&` and `||` (adds `&|` to allow-list in `find_hash_comment_start`)
- **#4694** — detect multiline block-comment TODO debt (carries `in_rust_block_comment` state across lines)
- **#4697** — accept punctuated TODO issue links (`TODO: (#42)` and `TODO - (#42)` now count as linked)
- **#4699** — limit TODO scan to block comment bodies (single-line bug: current code scans past `*/`)
- **#4700** — avoid false TODO hits in escaped single-quoted strings

## DAP (1 bug)

- **#4696** — DAP inline-values: (a) clamp `end_line > line_count` to `line_count`, (b) don't treat `'` as quote-start between ident bytes (e.g. `John's`, `don't`). Distinct from merged #4643.

## Parser (1 bug)

- **#4690** — heredoc anti-pattern column mapping: old code under-counted by 1 with prior newlines. Refactors 5 detectors to shared `location_from_start` helper.

## Semantic analyzer (1 bug)

- **#4683** — `@hash{...}` hash-slice reads `%hash` not `@hash`; bridges sigil via ancestor inspection of `Binary { op: "{}" }` parent.

## Module resolution (1 bug)

- **#4692** — FindBin `use lib '../../../outside'` could escape workspace-root boundary (path traversal). Adds manual Component-walker with traversal rejection. Security-adjacent.

## Workspace (1 bug)

- **#4693** — workspace-folder accepts `FILE:///…` (uppercase scheme) per RFC 3986 case-insensitive scheme rule.

## Color (1 bug)

[...]

PR #4725, opened 2026-04-23
fix(dap-inline-values): ignore quote-like Perl string operators (#4707)

EffortlessSteven:
### Motivation

- Prevent false positive variable matches inside Perl quote-like operators (`q`, `qq`, `qw`, `qr`, `qx`) when extracting inline values for the DAP UI.
- Preserve support for legacy Perl namespace separators (e.g. `$Foo'bar`) so a single-quote between identifier characters is not treated as a string delimiter.

### Description

- Added a `parse_quote_like_operator` helper and `matching_delimiter`/`is_identifier_byte` utilities and integrated them into `code_byte_mask` so quote-like operator regions are marked as non-code and skipped by regex scans.
- Made single-quote handling in `code_byte_mask` context-aware so `'` between identifier bytes is recognized as the legacy namespace separator instead of opening a single-quoted string.
- Added unit tests to `crates/perl-dap/src/inline_values.rs` that verify quote-like operator contents are ignored and that legacy namespace-qualified variables are still recognized.
- Changes applied to `crates/perl-dap/src/inline_values.rs` to harden inline-value extraction against these cases.

### Testing

- Ran formatter with `cargo xtask fmt` which completed successfully.
- Ran `cargo test -p perl-dap inline_values` during development which initially surfaced two failing legacy-namespace tests prompting the namespace-separator fix.
- Re-ran targeted tests with `cargo test -p perl-dap --lib inline_values::tests` after the fix and all tests passed (22 passed; 0 failed).

------

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -101,6 +101,14 @@ fn code_byte_mask(line: &str) -> Vec<bool> {
             continue;
         }
 
+        if let Some(end_idx) = parse_quote_like_operator(bytes, i) {
+            for byte in mask.iter_mut().take(end_idx).skip(i) {
+                *byte = false;
+            }
+            i = end_idx;
+            continue;
+        }
+
         match b {
             b'#' => {
                 if is_perl_array_length_marker(bytes, i) {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Rebased on master — resolved conflict with sibling merges. Kept master's `is_perl_array_length_marker` guard alongside PR's new `parse_quote_like_operator`/`is_identifier_byte` functions; both test blocks preserved.

PR #4726, opened 2026-04-23
fix(performance): stabilize parallel processing order (#0000)

EffortlessSteven:
### Motivation
- `process_files_parallel` produced nondeterministic output order due to concurrent worker completion, which can cause upstream consumers and tests to rely on sorted results instead of preserved input order.
- Tests previously sorted results to assert correctness instead of verifying deterministic ordering, masking the original behavior.

### Description
- Change `parallel::process_files_parallel` to tag each input with its original index, send `(index, result)` from workers, and reassemble the final `Vec<T>` in original input order after workers complete in `crates/perl-lsp-rs-core/src/performance/mod.rs`.
- Preserve existing safety behavior for `num_workers == 0` and when `num_workers` exceeds input count by computing `effective_workers` and keeping worker pool semantics.
- Replace the work queue with a vector of `(index, file)` pairs and adjust channel send/receive to carry indices.
- Update tests in `crates/perl-lsp-rs-core/tests/performance_module_shape.rs` to stop sorting outputs and add `parallel_given_multiple_workers_when_processing_then_input_order_is_preserved` to assert ordering under skewed per-file completion times.

### Testing
- Ran `cargo test -p perl-lsp-rs-core --test performance_module_shape` and all tests passed (`10 passed; 0 failed`).

------

### DIFF
--- a/crates/perl-lsp-rs-core/src/performance/mod.rs
+++ b/crates/perl-lsp-rs-core/src/performance/mod.rs
@@ -201,12 +201,14 @@ pub mod parallel {
             return Vec::new();
         }
 
+        let file_count = files.len();
+
         // Ensure callers cannot accidentally request zero workers and drop all work.
         // This preserves the API contract that every input file is processed once.
-        let effective_workers = num_workers.max(1).min(files.len());
+        let effective_workers = num_workers.max(1).min(file_count);
 
         let (tx, rx) = mpsc::channel();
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as superseded by merged #4680 (deterministic parallel ordering via index-tagged work queue). Same scope and fix.

PR #4727, opened 2026-04-23
fix(perl-dap): preserve legacy package vars in inline values (#0000)

EffortlessSteven:
### Motivation
- Inline variable extraction in the DAP code was incorrectly treating apostrophes used as legacy Perl namespace separators (e.g. `$Foo'bar`) as the start of single-quoted strings, which caused valid legacy-namespaced variables to be masked and missed by runtime/value-extraction logic.

### Description
- Adjusted `code_byte_mask` in `crates/perl-dap/src/inline_values.rs` to detect apostrophes that act as legacy namespace separators and not enter single-quote masking for those cases.
- Added small helpers `is_ident_char` and `is_ident_start` and use them to recognize an apostrophe as a legacy separator when it sits between identifier characters.
- Preserved original single- and double-quote masking behavior for true string literals while allowing the regex-based extraction and runtime lookup to see legacy namespaced variables.

### Testing
- Ran `cargo test -p perl-dap --lib inline_values::tests` and the targeted inline-values test suite passed (`20 passed; 0 failed`).
- Ran `cargo test -p perl-dap --lib` and the full `perl-dap` library tests passed (`340 passed; 0 failed`).
- Ran `cargo xtask fmt` to ensure formatting and it completed successfully.

------

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -61,6 +61,14 @@ fn normalize_line_bounds(
 /// Bytes that belong to Perl comments (`# ...`) or single/double-quoted string
 /// literals are marked `false` and should be ignored by lightweight regex scans.
 fn code_byte_mask(line: &str) -> Vec<bool> {
+    fn is_ident_char(byte: u8) -> bool {
+        byte.is_ascii_alphanumeric() || byte == b'_'
+    }
+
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as superseded by merged #4613 which already handles legacy `Foo'bar` apostrophe-as-namespace-separator in inline-values.

PR #4733, opened 2026-04-23
fix(dap): harden inline value masking for quote-like ops (#0000)

EffortlessSteven:
### Motivation

- Prevent false-positive variable captures inside Perl quote-like literals and regexes when computing DAP inline values. 
- Preserve legacy namespace behavior (apostrophe-qualified names like `$Foo'bar`) so they are not misclassified as string delimiters.

### Description

- Treat `q`, `qq`, `qw`, `qx`, and `qr` quote-like operators as non-code spans in `code_byte_mask` and skip their contents during regex-based variable scanning. 
- Add `quote_like_end` helper to locate the end of a quote-like span, handling paired delimiters, escapes, nested delimiters, and optional whitespace. 
- Add small helpers `is_ident_char`, `is_variable_sigil`, and `is_legacy_namespace_separator`, and update single-quote handling to avoid treating legacy namespace apostrophes as string starts. 
- Add a focused regression test `test_quote_like_operators_are_ignored` to ensure variables inside quote-like operators are ignored and real variables on the same line are still detected.

### Testing

- Ran `cargo test -p perl-dap --lib inline_values::tests::` and verified all inline-values tests passed (21 passed, 0 failed). 
- Exercised the specific regression via `cargo test -p perl-dap --lib inline_values::tests::test_quote_like_operators_are_ignored` which passed. 
- Ran `cargo xtask fmt` to format changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97cf74f948333866534d04d944d4e)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -105,12 +105,23 @@ fn code_byte_mask(line: &str) -> Vec<bool> {
                 break;
             }
             b'\'' => {
-                mask[i] = false;
-                in_single = true;
+                if !is_legacy_namespace_separator(bytes, i) {
+                    mask[i] = false;
+                    in_single = true;
+                }
             }
             b'"' => {
                 mask[i] = false;
                 in_double = true;
+            }
+            b'q' => {
+                if let Some(end_idx) = quote_like_end(bytes, i) {
+                    for byte in mask.iter_mut().take(end_idx).skip(i) {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of #4725 (winner). Both add quote-like operator (q/qq/qw/qr/qx) masking in inline-values. #4725 has richer delimiter-pairing logic and more tests.

PR #4738, opened 2026-04-23
fix(perl-dap): harden inline value masking for quote-like literals (#0000)

EffortlessSteven:
### Motivation
- Lightweight regex scans for inline values were producing false positives when Perl quote-like operators (e.g. `q`, `qq`, `qw`, `qr`, `qx`) contained variable-like tokens.
- Non-nesting quote-like delimiters were not scanned correctly, allowing quoted variable-like tokens to leak into inline value extraction.
- Apostrophes used as legacy namespace separators (e.g. `$Foo'bar`) were being treated as string delimiters and caused regressions for legacy-qualified symbols.

### Description
- Added a `scan_quote_like_literal` helper and integrated it into `code_byte_mask` to recognize and mask Perl quote-like literals so variables inside them are ignored. 
- Fixed delimiter scanning so non-nesting quote-like delimiters (e.g. `qq/.../`) scan from after the opening delimiter and nested delimiters (e.g. `q(...)`) handle depth correctly.
- Preserved legacy apostrophe namespace support by detecting apostrophes between identifier bytes and skipping them as string delimiters via an `is_ident_byte` helper.
- Added focused regression tests and assertions in `crates/perl-dap/src/inline_values.rs` to validate quote-like masking and to ensure non-quote-like identifiers remain detectable.

### Testing
- Ran `cargo xtask fmt` successfully to apply repository formatting rules.
- Ran `cargo test -p perl-dap inline_values::tests:: --lib` and all inline value tests passed (22 tests passed, 0 failed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97c90f54c8333b8b3a485642f33ee)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -97,6 +97,14 @@ fn code_byte_mask(line: &str) -> Vec<bool> {
             continue;
         }
 
+        if let Some(end_idx) = scan_quote_like_literal(bytes, i) {
+            for byte in mask.iter_mut().take(end_idx).skip(i) {
+                *byte = false;
+            }
+            i = end_idx;
+            continue;
+        }
+
         match b {
             b'#' => {
                 for byte in mask.iter_mut().take(bytes.len()).skip(i) {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of pending #4725. Same scope (quote-like operator masking + legacy `'` namespace-sep in `inline_values.rs`), same function, same tests.

PR #4750, opened 2026-04-23
fix(dap-inline-values): ignore variables inside Perl quote operators (#9999)

EffortlessSteven:
### Motivation
- Prevent false-positive inline variable detections when Perl quote-like operators (e.g. `q`, `qq`, `qw`, `qr`, `qx`) contain `$foo`-like text that is not executable code. 
- Keep the DAP `inlineValues` lightweight scanner syntax-aware without introducing a full parser. 
- Improve UX and correctness for inline value presentation by aligning masking with Perl literal forms.

### Description
- Extend `code_byte_mask` to recognize a `q`-style quote operator and delegate masking to a new helper `mask_quote_like_literal`. 
- Add `mask_quote_like_literal` which finds the operator’s delimiter, supports paired delimiters (`()`, `{}`, `[]`, `<>`) and same delimiters (e.g. `/.../`), handles escapes and nested pairs, and marks the literal region as non-code in the mask. 
- Add `paired_delimiter` helper and ensure the mask skip length accounts for the operator prefix (`q`, `qq`, `qr`, etc.). 
- Add focused regression tests that assert `extract_variable_names` and `collect_inline_values_with_runtime` ignore variable-like tokens inside quote-operator literals while still capturing real variables.

### Testing
- Ran `cargo test -p perl-dap quote_operators -- --nocapture` which executed the new regression tests and they passed (`2 passed`). 
- Verified `cargo test -p perl-dap` unit test runs exercised the modified `inline_values` module and the new tests passed as part of package test runs.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97c7a419c8333a7932ee7d28950de)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -104,6 +104,12 @@ fn code_byte_mask(line: &str) -> Vec<bool> {
                 }
                 break;
             }
+            b'q' => {
+                if let Some(next_idx) = mask_quote_like_literal(bytes, i, &mut mask) {
+                    i = next_idx;
+                    continue;
+                }
+            }
             b'\'' => {
                 mask[i] = false;
                 in_single = true;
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of pending #4725 (same quote-like ops masking in inline_values.rs).

PR #4751, opened 2026-04-23
fix(dap): mask Perl quote-like literals in inline extraction (#0000)

EffortlessSteven:
### Motivation
- Inline-value extraction was producing false positives when variable-like tokens appeared inside Perl quote-like literals (`q`, `qq`, `qw`, `qr`, `qx`).
- Single-quote handling also conflicted with legacy package separators such as `$Foo'bar`, causing real variables to be missed.

### Description
- Add `matching_delimiter` and `quote_like_literal_end` helpers and extend `code_byte_mask` to recognize and mask quote-like literals (including paired delimiters and escaped characters) so regex scans ignore their contents.
- Update single-quote handling in `code_byte_mask` to detect when `'` is a package separator vs a string start and avoid entering single-quote mode for legacy namespace usage.
- Add regression tests that verify variable-like tokens inside quote-like literals are ignored and that nested/paired `qr{...}` patterns are masked while real variables are still extracted.

### Testing
- During development `cargo test -p perl-dap inline_values` failed on legacy-namespace tests, the logic was adjusted and re-run to address the failure.
- Final targeted test run `cargo test -p perl-dap --lib inline_values::tests::` completed successfully with all added and existing inline-value tests passing.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97c7a419c8333a7932ee7d28950de)

### DIFF
--- a/crates/perl-dap/src/inline_values.rs
+++ b/crates/perl-dap/src/inline_values.rs
@@ -60,6 +60,77 @@ fn normalize_line_bounds(
 ///
 /// Bytes that belong to Perl comments (`# ...`) or single/double-quoted string
 /// literals are marked `false` and should be ignored by lightweight regex scans.
+fn matching_delimiter(open: u8) -> Option<u8> {
+    match open {
+        b'(' => Some(b')'),
+        b'{' => Some(b'}'),
+        b'[' => Some(b']'),
+        b'<' => Some(b'>'),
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of pending #4725 (same scope).

PR #4759, opened 2026-04-23
fix(ci-hygiene): require TODO/FIXME word boundaries (#0000)

EffortlessSteven:
### Motivation

- Reduce false positives from the repository TODO/FIXME scanner by ensuring the matcher only flags standalone markers and not substrings like `TODOLIST` or `PRETODO`.

### Description

- Tighten the token matcher in `perl-ci-hygiene` by changing the regex to `\b(?:TODO|FIXME)\b` used by the TODO scanner.
- Update existing unit tests to use the stricter regex so test behavior matches production scanning.
- Add a regression test `todo_detection_requires_word_boundaries` in `crates/perl-ci-hygiene/src/main.rs` that asserts marker-like substrings do not trigger while true `TODO`/`FIXME` markers still do.

### Testing

- Ran `cargo xtask fmt` to apply repository formatting and it completed successfully.
- Ran `cargo test -p perl-ci-hygiene todo_detection_requires_word_boundaries` and the new test passed.
- Ran focused unit tests `hash_comment_todo_detection_handles_shebang_and_inline_hashes` and `rust_todo_detection_ignores_linked_or_url_like_comments` and both passed, confirming no regressions in existing behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97c7b4c608333b8bb8c25a060755c)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3384,7 +3384,7 @@ fn cmd_check_todos(repo_root: &Path, list_mode: bool) -> Result<i32> {
             .join("complex_paren_args_tests.rs"),
     ];
 
-    let todo_re = Regex::new(r"TODO|FIXME")?;
+    let todo_re = Regex::new(r"\b(?:TODO|FIXME)\b")?;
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of pending #4737 (third copy of word-boundary tightening).

PR #4763, opened 2026-04-23
fix(performance): normalize node spans and propagate worker panics (#4707)

EffortlessSteven:
### Motivation
- Prevent incorrect reparse decisions when callers pass node ranges with `node_start > node_end` by normalizing spans inside the check. 
- Surface thread panics from worker threads instead of silently swallowing them so parallel indexing failures are visible and debuggable.

### Description
- Normalize `node_start`/`node_end` in `IncrementalParser::needs_reparse` so overlap detection works regardless of argument order. 
- Change `process_files_parallel` to propagate worker thread panics by calling `std::panic::resume_unwind` if `join()` returns an `Err`.
- Add focused unit tests in `crates/perl-lsp-rs-core/src/performance/mod.rs` to cover reversed-range handling and panic propagation from workers.

### Testing
- Ran `cargo xtask fmt` to apply formatting, which completed successfully. 
- Ran `cargo test -p perl-lsp-rs-core performance::tests -- --nocapture`, and the new focused tests passed (`2 passed`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e97cfbb8148333b686f01adcc14026)

### DIFF
--- a/crates/perl-lsp-rs-core/src/performance/mod.rs
+++ b/crates/perl-lsp-rs-core/src/performance/mod.rs
@@ -256,7 +256,9 @@ pub mod parallel {
         drop(tx);
 
         for handle in handles {
-            let _ = handle.join();
+            if let Err(payload) = handle.join() {
+                std::panic::resume_unwind(payload);
+            }
         }
 
         let mut ordered_results = Vec::with_capacity(file_count);
@@ -271,6 +273,7 @@ pub mod parallel {
 
 #[cfg(test)]
 mod tests {
+    use super::IncrementalParser;
     use super::parallel::process_files_parallel;
 
     #[test]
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Rebased on master — resolved conflict with sibling #4707 bundle merges (#4680 deterministic ordering, #4682 node spans + zero-width). Analysis: (a) reversed-range normalization in needs_reparse was already in master from #4682 — kept master's more complete version which also handles zero-width point nodes; (b) panic propagation via resume_unwind is new and not in master — preserved. Merged the duplicate mod tests blocks into one. Ordered results collection from #4680 preserved (not reverted). Compiles cleanly under cargo check -p perl-lsp-rs-core --tests.

PR #4781, opened 2026-04-23
fix(workspace-index): dedupe use constant qw names (#0000)

EffortlessSteven:
### Motivation
- `use constant qw(...)` values were being appended without deduplication, causing duplicate symbol entries when indexing qw-form constants and exposing a failing unit test.

### Description
- Route parsed `qw` words through the existing `push_unique` path in `extract_constant_names_from_use_args` so duplicate names are suppressed (file: `crates/perl-workspace-index/src/workspace/workspace_index.rs`).
- Add a regression test `test_use_constant_qw_duplicate_names_indexed_once` that indexes `use constant qw(FOO BAR FOO);` and asserts each constant is indexed exactly once.
- Minor formatting/cleanup applied to the surrounding code to keep style consistent.

### Testing
- Ran `cargo xtask fmt` to format the workspace, which completed successfully.
- Ran the workspace unit tests for the modified crate with `cargo test -p perl-workspace --lib` and targeted tests; the previously failing `test_extract_constant_names_deduplicates_qw_form` now passes and the test suite reports all tests passing (217 passed; 0 failed).
- Also ran targeted tests `test_extract_constant_names_deduplicates_qw_form` and `test_use_constant_qw_duplicate_names_indexed_once` which both passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e98e0cc998833380ffeb0f9032adda)

PR #4785, opened 2026-04-23
fix(workspace-index): dedupe `use constant qw(...)` symbols (#0000)

EffortlessSteven:
### Motivation
- The `use constant qw(...)` fast-path could return duplicate names (e.g. `FOO` twice) which caused duplicate symbol entries in the workspace index. 
- Deduplication should be consistent across all `use constant` forms (scalar, hash, and `qw`) to avoid emitting duplicate symbols to consumers.

### Description
- Update `extract_constant_names_from_use_args` to use the shared `push_unique` helper when collecting words from the `qw(...)` fast-path so names are deduplicated via the same `HashSet` logic as other forms. 
- Add a regression test `test_use_constant_qw_duplicate_names_indexed_once` which indexes a file containing `use constant qw(RETRY_COUNT TIMEOUT RETRY_COUNT);` and asserts `RETRY_COUNT` is emitted only once. 
- Changes are in `crates/perl-workspace-index/src/workspace/workspace_index.rs` and include small formatting adjustments.

### Testing
- Ran the focused unit test `test_extract_constant_names_deduplicates_qw_form` which passed. 
- Ran the new regression test `test_use_constant_qw_duplicate_names_indexed_once` which passed. 
- Ran the full crate tests with `cargo test -p perl-workspace --lib -- --nocapture` which resulted in all tests passing (previous single failing test for this behavior is now fixed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e98e25b6a48333abbc6644fe8dc4bc)

### DIFF
--- a/crates/perl-workspace-index/src/workspace/workspace_index.rs
+++ b/crates/perl-workspace-index/src/workspace/workspace_index.rs
@@ -3378,9 +3378,10 @@ fn extract_module_names_from_use_args(args: &[String]) -> Vec<String> {
                 if stripped.is_empty() {
                     return None;
                 }
-                if stripped.chars().all(|c| {
-                    c.is_alphanumeric() || c == '_' || c == ':' || c == '\''
-                }) {
+                if stripped
+                    .chars()
+                    .all(|c| c.is_alphanumeric() || c == '_' || c == ':' || c == '\'')
+                {
                     Some(stripped.to_string())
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing — identical diff to #4781; core change already in merged #4745.

PR #4837, opened 2026-04-23
refactor(performance): replace panic helper with Result (#0000)

EffortlessSteven:
### Motivation
- Remove panic-based behavior from the `performance` helper so production code avoids unexpected process termination and tests can assert results explicitly.

### Description
- Change `performance::assert_some` to return `Result<T, String>` instead of panicking on `None`, and update integration tests to check `.is_ok()` at call sites.

### Testing
- Ran `cargo xtask fmt` to format changes and `cargo test -p perl-lsp-rs-core --test performance_module_shape`, and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e99b1bf7e08333bcda9960f260295d)

### DIFF
--- a/crates/perl-lsp-rs-core/src/performance/mod.rs
+++ b/crates/perl-lsp-rs-core/src/performance/mod.rs
@@ -15,20 +15,12 @@ use std::time::Duration;
 
 pub use perl_symbol::SymbolIndex;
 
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing in favor of #4836. The Result-returning approach keeps a test-only API in the production module; #4836's deletion is the right call (the helper had no production callers).

PR #4874, opened 2026-04-23
fix(ci-hygiene): stabilize hash-line TODO parsing (#0000)

EffortlessSteven:
### Motivation
- The hash-comment TODO scanner had a state-machine bug that could fail to exit double-quoted contexts and mishandle ANSI-C single-quoted strings, causing false negatives/positives when detecting unlinked `TODO`/`FIXME` markers.
- Improve accuracy of TODO detection in hash-comment languages (shell/Justfile/other hash-delimited files) to reduce spurious CI failures and match existing test expectations.

### Description
- Fixes the `find_hash_comment_start` state machine to correctly track and exit double-quoted strings and to separate normal single-quote and ANSI-C single-quote (`$'...'`) handling. 
- Restores correct escape-state handling for ANSI-C single-quoted mode and normal single-quoted mode so escaped quotes are respected.
- Extends the allowed prefix characters for hash-comments to include `,` so inline comments like `foo, # comment` are recognized as comments. 
- Adds/adjusts regression assertions to cover TODO detection after closed double-quoted strings and TODO-like tokens inside quoted content; changes are in `crates/perl-ci-hygiene/src/main.rs`.

### Testing
- Ran `cargo test -p perl-ci-hygiene hash_comment_todo_detection_handles_shebang_and_inline_hashes` and the test passed. 
- Ran `cargo test -p perl-ci-hygiene perl_todo_detection_allows_comment_start_without_whitespace` and the test passed. 
- Ran `cargo xtask fmt` to apply formatting, which completed successfully. 
- A prior full run of `cargo test -p perl-ci-hygiene` showed two unrelated failing Rust TODO-detection tests that are not caused by these changes; the targeted tests added/modified by this PR are green.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e99aa2ad208333ab90094269a8adf6)

### DIFF
--- a/crates/perl-ci-hygiene/src/main.rs
+++ b/crates/perl-ci-hygiene/src/main.rs
@@ -3939,6 +3939,21 @@ fn find_hash_comment_start(line: &str, perl_mode: bool) -> Option<usize> {
     let mut prev_was_escape_single = false;
 
     for (idx, ch) in line.char_indices() {
+        if in_single_ansi_c {
+            if prev_was_escape_single {
+                prev_was_escape_single = false;
+                continue;
+            }
+            if ch == '\\' {
+                prev_was_escape_single = true;
+                continue;
+            }
+            if ch == '\'' {
+                in_single = false;
+                in_single_ansi_c = false;
+            }
+            continue;
+        }
         if in_single {
             if prev_single_was_escape {
                 prev_single_was_escape = false;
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as overlap with #4872 (winner). Both touch `find_hash_comment_start` in perl-ci-hygiene. #4872 is a clean state-machine consolidation; this PR reorders branches in a way that looks broken (in_single_ansi_c is checked before in_single but never set to true in this diff). The comma-punctuation addition here is independent value but #4872 is the safer landing.

PR #4952, opened 2026-04-23
fix(lsp): ignore stale didChange document versions (#0000)

EffortlessSteven:
### Motivation

- Protect the server from out-of-order `textDocument/didChange` notifications that can arrive during reconnects or transport retries and cause unnecessary cache invalidation and parse work.

### Description

- Add an early guard in `handle_did_change_with_cancellation` that returns immediately when an incoming change has a non-positive version or a version less-than-or-equal to the currently tracked document version, preventing mutation of caches or parse work. 
- Ensure the stale-version check runs before semantic analyzer/perlcritic cache invalidation and before applying edits to the in-memory document. 
- Add a regression unit test `handle_did_change_ignores_stale_version` in the `text_sync` tests which verifies a stale `didChange` does not update the stored `version` or `text`.

### Testing

- Ran `cargo xtask fmt` to format changes successfully. 
- Ran `cargo test -p perl-lsp-rs --lib handle_did_change_ignores_stale_version` and the new test passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9a75124d0833389d57daae6ac1c08)

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/text_sync.rs
+++ b/crates/perl-lsp-rs/src/runtime/text_sync.rs
@@ -395,6 +395,20 @@ impl LspServer {
                 // Get current document state or create new one
                 let mut documents = self.documents.lock();
                 let normalized_uri = self.normalize_uri_key(uri);
+                if let Some(existing_doc) = self.get_document(&documents, uri) {
+                    // Some clients can deliver out-of-order notifications during
+                    // reconnects or transport retries. Ignore stale changes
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as narrower dup of #4950 (winner). #4950 has version-optional handling, stream-session cancel-by-version, stale guard, missing-version fallback, and 2 regression tests. This PR is the subset without those.

PR #4979, opened 2026-04-23
fix(config): improve @INC normalization and dedupe behavior (#4982)

EffortlessSteven:
### Motivation
- Normalize and dedupe include-path inputs to avoid empty/whitespace-only entries and duplicate search roots that can produce confusing search orders or redundant work.
- Ensure merged include-path lists (workspace `include_paths` + `PERL5LIB`) preserve configured precedence while removing duplicates.

### Description
- Added a shared `dedupe_preserve_order` helper and applied it to `WorkspaceConfig::parse_perl5lib` to trim and dedupe `PERL5LIB` entries. (file: `crates/perl-lsp-rs-core/src/config/mod.rs`)
- Switched `WorkspaceConfig::effective_include_paths` to build merged include path vectors via the helper so both `Prepend` and `Append` precedence modes preserve order and remove duplicates. (file: `crates/perl-lsp-rs-core/src/config/mod.rs`)
- Normalized and deduped interpreter startup `@INC` probe output in `fetch_perl_inc` so cached system `@INC` entries are trimmed and de-duplicated. (file: `crates/perl-lsp-rs-core/src/config/mod.rs`)
- Added unit tests covering `PERL5LIB` parsing and effective include-path merging behavior for both `Prepend` and `Append` modes. (file: `crates/perl-lsp-rs-core/src/config/mod.rs` tests module)

### Testing
- Ran `cargo xtask fmt` to format the workspace; operation completed successfully. 
- Ran `cargo test -p perl-lsp-rs-core parse_perl5lib_trims_and_dedupes_entries` and the test passed. 
- Ran `cargo test -p perl-lsp-rs-core effective_include_paths_dedupes` which executed the added merging tests and they passed.
- The targeted unit tests added are: `parse_perl5lib_trims_and_dedupes_entries`, `effective_include_paths_dedupes_with_prepend_precedence`, and `effective_include_paths_dedupes_with_append_precedence`, and they all passed in CI-local runs.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9adf53760833385bc1ed3b72b716b)

PR #4981, opened 2026-04-23
fix(config): harden system @INC path normalization (#0000)

EffortlessSteven:
### Motivation

- Reduce noisy or non-path entries coming from Perl startup `@INC` (including references added by hooks) so the resolver only receives scalar filesystem paths.
- Prevent duplicate directory scans and spurious `.`/empty entries during module resolution by normalizing the system include list.

### Description

- Tightened the Perl probe command in `fetch_perl_inc` to only print scalar, non-reference, non-empty `@INC` entries via `grep { defined($_) && !ref($_) && length($_) } @INC`.
- Added `sanitize_system_inc_paths` to trim whitespace, drop empty or `.` entries, deduplicate while preserving order, and return a `Vec<PathBuf>` for use by the resolver.
- Wired `fetch_perl_inc` to call `sanitize_system_inc_paths` and added focused unit tests for sanitizer behavior in `crates/perl-lsp-rs-core/src/config/mod.rs`.

### Testing

- Executed the targeted unit tests with `cargo test -p perl-lsp-rs-core sanitize_system_inc_paths -- --nocapture` and both sanitizer tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9adf53760833385bc1ed3b72b716b)

### DIFF
--- a/crates/perl-lsp-rs-core/src/config/mod.rs
+++ b/crates/perl-lsp-rs-core/src/config/mod.rs
@@ -549,14 +549,18 @@ impl WorkspaceConfig {
         };
         let mut command = Command::new(perl_path);
         command.args(perl_args);
-        let output = command.args(["-e", "print join(\"\\n\", @INC)"]).output();
+        let output = command
+            .args([
+                "-e",
+                "print join(\"\\n\", grep { defined($_) && !ref($_) && length($_) } @INC)",
+            ])
+            .output();
 
         match output {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as subset of #4979 (winner — comprehensive @INC dedupe via shared `dedupe_preserve_order` helper). This PR only covers `fetch_perl_inc` sanitization; #4979 covers that plus `parse_perl5lib` + `effective_include_paths` consistently.

Issue #2080, opened 2026-03-19
Implement incremental parser checkpoint recovery for large edits


EffortlessSteven:
## Performance Opportunity: Incremental Parsing Strategy

**File**: crates/perl-lexer/src/checkpoint.rs (checkpoint cache exists but underutilized)
**Hot Path**: didChange event -> re-parse (1000+ line documents)
**Issue**: Large document edits re-parse from document start instead of checkpoint-based recovery

### Current Pattern
- Checkpoint cache exists (CheckpointCache) but limited to 10 checkpoints max
- No intelligent choice of checkpoint for starting re-parse
- Full re-parse from offset 0 for any document change
- Large document with edit at line 500: reparsing 500 lines instead of ~50

### Impact
Editing a 1000-line file with checkpoint at line 900 forces re-scan of 900 lines on save. With 10ms per 100 lines, editor feels sluggish on large files.

### Solution Approach
- Track checkpoint positions with document version
- On didChange: find nearest checkpoint before edit start
- Resume parsing from checkpoint instead of document start
- Increase checkpoint max from 10 -> 50 for large files
- Measure: time to re-parse 500-line change in 5000-line file

### Acceptance Criteria
- Re-parse latency on 500-line edit in 5000-line doc < 100ms (vs. 300ms+ unchecked)
- Memory overhead < 2MB per 10k-line document
- Checkpoint selection is O(log N) with binary search

PR #4997, opened 2026-04-23
fix(incremental): apply didChange edits sequentially (#0000)

EffortlessSteven:
### Motivation
- LSP `textDocument/didChange` sends `contentChanges` that are defined against intermediate document states, so byte ranges must be computed by replaying changes in-order to build correct incremental edits. 
- Applying multiple ranged edits as a single batch against the original source caused incorrect byte offsets and prevented checkpoint/AST fast-paths from taking the correct update path. 

### Description
- Add `build_incremental_edits_from_lsp_changes` which replays `contentChanges` on a temporary `Rope` and returns per-change `IncrementalEdit` byte ranges, returning `None` for any full-document replace. 
- Use the builder output to drive incremental AST updates by applying edits one-by-one via `IncrementalDocument::apply_edit` instead of attempting a single batch application. 
- Apply checkpoint-based lexer updates one edit at a time by invoking `inc_apply_edits` with single-edit slices and reinitialize checkpoints only if any single application fails. 
- Add unit tests covering sequential range mapping and full-replace fallback behavior for the incremental edit builder inside `runtime::text_sync::tests`. 

### Testing
- Ran targeted unit tests `incremental_edit_builder_tracks_sequential_change_coordinates` and `incremental_edit_builder_disables_for_full_replace` which passed. 
- Ran `test_incremental_path_taken_on_ranged_change` and the full `runtime::text_sync::tests::` suite, and all tests passed (32 passed, 0 failed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9adfaeac883338dc032e4d48a918c)

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/text_sync.rs
+++ b/crates/perl-lsp-rs/src/runtime/text_sync.rs
@@ -15,6 +15,29 @@ use crate::state::DegradationTier;
 use perl_parser::workspace_index::{IndexPhase, IndexState};
 use perl_parser_core::source_file::is_binary_content;
 
+#[cfg(feature = "incremental")]
+fn build_incremental_edits_from_lsp_changes(
+    rope: &ropey::Rope,
+    changes: &[lsp_types::TextDocumentContentChangeEvent],
+) -> Option<Vec<perl_parser::incremental::incremental_edit::IncrementalEdit>> {
+    use crate::textdoc::{Doc, PosEnc, apply_changes, range_to_bytes};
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of #4999 — both fix the same root cause (LSP edit ranges must be mapped against the evolving document state, not the pre-change rope). #4999 is narrower (returns the existing IncrementalEditSet type, doesn't refactor downstream consumer) and references the real issue #2080. Keeping #4999.

PR #4999, opened 2026-04-23
fix(lsp): correct incremental didChange range mapping across multi-edits (#2080)

EffortlessSteven:
### Motivation
- LSP `contentChanges` are applied in order and later ranges are relative to the already-applied earlier edits, but the prior implementation computed all incremental byte ranges from the original pre-change text which can mis-map later edits and force unnecessary incremental fallbacks.
- Fixing the mapping avoids spurious incremental fallback paths and preserves correct incremental parsing/lexing behavior for multi-edit notifications.

### Description
- Add `build_incremental_edit_set` which computes an `IncrementalEditSet` by applying each `TextDocumentContentChangeEvent` to a `working_rope` snapshot so byte/char ranges for subsequent edits are measured against the evolving document, using `PosEnc::Utf16` for conversions.
- Wire `handle_did_change` to call `build_incremental_edit_set(&doc_state.rope, &lsp_changes)` instead of building ranges only from the original rope.
- Add unit test `test_build_incremental_edits_uses_evolving_document_ranges` that verifies the second edit in a multi-change notification is measured relative to the post-first-edit document.

### Testing
- Ran the new unit test with `cargo test -p perl-lsp-rs --lib test_build_incremental_edits_uses_evolving_document_ranges --features incremental` and it passed.
- Ran the existing incremental integration test with `cargo test -p perl-lsp-rs --lib test_incremental_path_taken_on_ranged_change --features incremental` and it passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9adfaeac883338dc032e4d48a918c)


EffortlessSteven:
Deep review — REQUEST CHANGES.

## Critical coordinate space bug

The `build_incremental_edit_set` helper correctly uses an evolving `working_rope` to update rope state so `range_to_chars` yields correct char indices for each successive mutation. **But what it stores in `IncrementalEditSet` is wrong.**

It calls `range_to_bytes(&working_rope, range, ...)` on the post-previous-edit rope, so each `IncrementalEdit`'s `start_byte`/`old_end_byte` are in **evolving-document coordinates**. But the consumers expect **original-document coordinates**:

- `crates/perl-parser/src/incremental/incremental_document.rs:132` — `apply_edits` sorts edits in reverse `start_byte` order and applies them sequentially against the evolving `new_source`. That only works if all offsets are in original-doc space.
- `crates/perl-parser/src/incremental/mod.rs:370` — same pattern, same assumption.

## Concrete bug

For "abcde" with two edits:
1. Insert "X" at char 1  
2. Replace chars 4..6 of post-insert text with new content

The new code records `edits[1].start_byte = 4, old_end_byte = 6` (evolving space). But `apply_edits` in reverse order first applies edits[1] against "abcde" at bytes 4..6 = "e" (off by one), then edits[0] at bytes 1..1 inserts "X". Result: "aXbcdYZ" instead of the intended "aXbcYZ".

## The test is vacuous

`test_build_incremental_edits_uses_evolving_document_ranges` only checks the stored byte values (`edits[1].start_byte == 4`, `old_end_byte == 6`). It does NOT call `apply_edits` or inspect the resulting document. Those stored values happen to equal the raw LSP `character: 4` / `character: 6` in the test input, so the assertion would pass even if `build_incremental_edit_set` just copied range endpoints verbatim without any rope work.

## What the fix should do

`IncrementalEdit` offsets must be in original-document coordinate space. Options:
1. Compute each edit's original-space offset by subtracting the cumulative byte delta of prior edits in the batch
2. Keep `working_rope` for the apply loop only, but compute IncrementalEdit offsets from `original_rope` adjusted by the running byte shift
3. Strengthen the test to actually invoke `apply_edits` and assert the resulting document content

## Regression risk: HIGH

Multi-edit `textDocument/didChange` (paste, auto-format, multi-cursor) silently corrupts incremental parse trees → stale diagnostics, wrong hover/completion. Single-edit cases unaffected.

Sending back. Happy to merge once the coordinate space is fixed AND the test actually runs apply_edits.

EffortlessSteven:
## Deep-review fix pushed — b3ce820af

### What was found

`build_incremental_edit_set` computed byte offsets against the evolving working rope (correct for range-to-char conversion) but stored those **evolving-document coordinates** directly into `IncrementalEditSet`. The consumers (`apply_edits`, `apply_to_string`) sort edits in reverse `start_byte` order and apply them against the **original** source string — so they require all offsets in original-document space. For a two-edit batch where edit[0] inserted 1 byte, edit[1]'s stored `start_byte` was off by +1, causing silent document corruption (wrong characters replaced).

### What was fixed

Added `cumulative_shift: isize` tracking the net byte-length delta of all prior edits. Before storing each `IncrementalEdit`, we subtract `cumulative_shift` from the evolving-space offsets to recover original-document coordinates. The working rope is still updated after each edit so that subsequent `range_to_bytes`/`range_to_chars` calls see the correct evolved document.

Concrete example — original "abcde", two edits:
- Edit[0]: insert "X" at byte 1..1 → stored as `start_byte=1, old_end_byte=1` (unchanged, shift=0)
- Edit[1]: evolving bytes 4..6 → `orig_start = 4-1 = 3, orig_end = 6-1 = 5` → stored as `start_byte=3, old_end_byte=5`
- `apply_to_string("abcde")`: apply edit[1] first (bytes 3..5 → "YZ") → "abcYZ", then edit[0] (insert "X" at 1) → "aXbcYZ" ✓

### What the test now verifies

`test_build_incremental_edits_uses_evolving_document_ranges` now:
1. Asserts the stored byte offsets are in **original-document space** (edit[1]: `start_byte=3, old_end_byte=5`, not the vacuous 4/6 evolving values)
2. Calls `edit_set.apply_to_string("abcde")` and asserts the result equals `"aXbcYZ"` — confirming end-to-end correctness through the apply pipeline

### Test results

- `test_build_incremental_edits_uses_evolving_document_ranges`: ✅ passed
- Full `perl-lsp-rs --lib --features incremental`: 421 passed, 4 pre-existing failures in `security::sandbox` (Perl not installed in CI environment — unrelated to this change)
- `cargo clippy -p perl-lsp-rs --lib --features incremental`: ✅ no new warnings
- `cargo xtask fmt`: ✅ clean

Pushed SHA: b3ce820af5c031cc34d21028827f1817d87a275d

EffortlessSteven:
## Deep Review -- APPROVED

**Logic: CORRECT**

The `build_incremental_edit_set` extraction correctly addresses the coordinate-space bug. The old inline code called `position_to_offset_rope` once against the pre-change rope for all edits, so edit N+1's range was interpreted in original-document space even though the LSP client expressed it in post-edit-N space. The new function maintains a `working_rope` that accumulates prior edits, then maps evolving-document byte offsets back to original-document space via `cumulative_shift`. The math is correct: `orig = evolving - shift` where shift is the sum of `(new_len - deleted_len)` for all prior edits.

**Tests: GOOD**

The test `test_build_incremental_edits_uses_evolving_document_ranges` is non-vacuous. It asserts specific byte offset values (edit[0]: 1..1, edit[1]: 3..5) that would be wrong if the working-rope logic were absent -- a naive implementation would produce edit[1] as 4..6 (evolving space). The final `apply_to_string` assertion verifies the complete round-trip produces the client-intended document, which would fail without the coordinate mapping.

**Regression risk: LOW**

The function is extracted from inline code with a one-liner call site replacement. No existing callers break. Full-document replace (range == None) still falls through via `change.range.as_ref()?`.

**CI failures: INFRA FLAKE -- not caused by this PR**

UX Regression Gate ("Timeout waiting for LSP response") and PR Smoke are failing identically on recently-merged PRs #5310 and #5312. Known infrastructure issue, not a regression from this change.

**Research verification: SKIPPED** -- no external Perl/LSP/crate claims.

Label: `deep-reviewed` applied.

Issue #3475, opened 2026-04-10
[module-resolution] use constant symbols not tracked in visible symbol table


EffortlessSteven:
## Problem

Perl's `use constant` pragma creates bareword constants that should be available in the consumer's scope, but perl-lsp does not:
1. Extract constant names from `use constant { ... };` hash syntax
2. Track constants as visible symbols in the importing file
3. Enable go-to-definition on constant barewords

## Current Behavior

```perl
use constant {
    PI => 3.14159,
    E => 2.71828,
};
print PI;  # CTRL+Click fails — PI has no definition
```

Evidence:
- `#3391` notes only first constant extracted from hash syntax
- No tracking of constants in symbol visibility
- `perl-semantic-analyzer` does not treat constants as imported symbols

## Related Issues

This is a narrower version of the broader import tracking gap (#3472).

For `use constant`, the issue is:
1. `perl-symbol-surface/src/decl.rs` extracts constant names (partially, hash syntax incomplete)
2. But those names are not registered as visible in consuming files
3. No scope analyzer recognition of use-constant as symbol visibility change

## Example Cases

```perl
# Case 1: Hash-ref syntax (multiple constants)
use constant {
    MIN => 1,
    MAX => 100,
};
if ( > MAX) { }  # MAX should be defined

# Case 2: Single scalar syntax
use constant PI => 3.14159;
my  = 2 * PI * ;  # PI should be defined

# Case 3: Code reference (sub-like constant)
use constant NOW => sub { time() };
print NOW->();  # NOW should be defined
```

## Why This Matters

1. Constants are a standard Perl feature
2. Developers import constants just like functions
3. Scope diagnostics report false "undefined" errors
4. Completion doesn't suggest defined constants
5. Navigation can't jump to constant definitions

## Implementation Notes

This requires:
1. Improved constant extraction from all `use constant` forms
2. Register constants as imported symbols in consuming scope
3. Store constant -> definition location mapping
4. Enable navigation to `use constant` line

## Code Locations

- `crates/perl-symbol-surface/src/decl.rs` — Constant extraction (incomplete for hash syntax)
- `crates/perl-semantic-analyzer/src/analysis/symbol.rs` — No constant tracking in imports
- `crates/perl-workspace-index` — No constant definition storage

## Test Files

- `crates/perl-lsp/tests/lsp_definition_tests.rs` — Has some constant tests, but no import tracking
[...]

PR #5024, opened 2026-04-23
fix(completion): surface use constant symbols from visible table (#3475)

EffortlessSteven:
### Motivation

- `use constant` declarations were not synthesized into the local visible symbol table, causing bare constant completions (for same-file symbols) to be missing.  

### Description

- Synthesize `SymbolKind::Constant` entries from `use constant` `NodeKind::Use` args in the semantic symbol extractor by adding `synthesize_use_constant_symbols` and the helper `extract_constant_names_from_use_args` (in `crates/perl-semantic-analyzer/src/analysis/symbol.rs`).  
- Surface those synthesized constants in completion by treating `SymbolKind::Constant` like functions in `add_function_completions`, emitting `CompletionItemKind::Constant`, appropriate `insert_text`, and a constant sort tier (in `crates/perl-lsp-rs-core/src/providers/completion/completion/functions.rs`).  
- Allow constants to trigger the same visible-symbol completion path by including `SymbolKind::Constant` in `could_be_function` heuristics (in `crates/perl-lsp-rs-core/src/providers/completion/completion.rs`).  
- Add a regression test `test_use_constant_completion_from_visible_symbol_table` verifying a `use constant` symbol (`PI`) appears as a constant completion (in `crates/perl-lsp-rs-core/src/providers/completion/completion.rs`).

### Testing

- Ran the focused unit test: `cargo test -p perl-lsp-rs-core test_use_constant_completion_from_visible_symbol_table -- --nocapture` and it passed.  
- Ran semantic analyzer constant extraction test: `cargo test -p perl-semantic-analyzer symbol_extraction_constant -- --nocapture` and it passed.  
- Performed type checks with `cargo check --all-targets -p perl-lsp-rs-core` and `cargo check --all-targets -p perl-semantic-analyzer` and both completed successfully.  
- Executed linters `cargo clippy -p perl-lsp-rs-core -- -D warnings` and `cargo clippy -p perl-semantic-analyzer -- -D warnings` and they passed.  
- Ran formatting via `cargo xtask fmt`; completed locally in this environment.  
- Note: `just pr-fast` was not available in this environment (`just` not installed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b8b4a1ec8333ba9b3001249adfa4)

PR #5025, opened 2026-04-23
fix(completion): include visible constant symbols (#3475)

EffortlessSteven:
### Motivation

- Completion previously only treated `SymbolKind::Subroutine` (and builtins) as function-like candidates, so constants declared via `use constant` that are tracked in the visible symbol table were not suggested.

### Description

- Treat `SymbolKind::Constant` like a function candidate in prefix gating by updating `could_be_function` to consider constants in addition to subroutines.
- Emit constant completion items from the local symbol table in `add_function_completions` by mapping `SymbolKind::Constant` -> `CompletionItemKind::Constant` with constant-appropriate `insert_text` while preserving scope-distance sorting.
- Add a focused unit test `test_constant_symbols_in_visible_table_are_completed` that injects a visible `PI` constant into the provider's `symbol_table` and verifies it appears in completions.

### Testing

- Ran `cargo test -p perl-lsp-rs-core test_constant_symbols_in_visible_table_are_completed -- --nocapture`, which passed.
- Ran `cargo check --all-targets -p perl-lsp-rs-core`, which passed.
- Ran `cargo clippy -p perl-lsp-rs-core -- -D warnings`, which passed.
- `just pr-fast` was not available in this environment and therefore not run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b8b4a1ec8333ba9b3001249adfa4)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
@@ -1918,7 +1918,9 @@ impl CompletionProvider {
         // Check user-defined functions
         for (name, symbols) in &self.symbol_table.symbols {
             for symbol in symbols {
-                if symbol.kind == SymbolKind::Subroutine && name.starts_with(prefix) {
+                if (symbol.kind == SymbolKind::Subroutine || symbol.kind == SymbolKind::Constant)
+                    && name.starts_with(prefix)
+                {
                     return true;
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as narrower version of #5024, which also updates `symbol.rs` to expose the visible-table lookup used by completion.

PR #5026, opened 2026-04-23
fix(navigation): resolve require+import bareword goto-def (#0000)

EffortlessSteven:
### Motivation
- `textDocument/definition` did not resolve bareword subroutine calls that are imported at runtime via `require Module; Module->import('sym')`, leaving an import-visibility gap for goto-definition.

### Description
- Exposed the existing AST-based import-source resolver as `pub(crate) fn find_import_source(...)` so it can be reused by navigation code.
- Added a navigation fallback in the definition handler that maps unresolved bareword `Sub` symbol keys to the import source package (when `require` + literal `import(...)` is detected) and performs a workspace lookup for the target definition.
- Added a cross-file regression test `go_to_definition_on_require_imported_bareword_navigates_to_source_module` that verifies a `require` + `->import('runtime_sum')` scenario resolves go-to-definition to the module file.

### Testing
- Ran `cargo test -p perl-lsp-rs --test cross_file_goto_definition_tests go_to_definition_on_require_imported_bareword_navigates_to_source_module` and the new test passed.
- Ran `cargo clippy -p perl-lsp-rs` which completed (existing warnings remain but no new failures in this change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b877cee88333959cbf69a7cfcfab)

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/language/moniker.rs
+++ b/crates/perl-lsp-rs/src/runtime/language/moniker.rs
@@ -307,7 +307,11 @@ impl LspServer {
     ///
     /// Searches `use` statements for the symbol name, handling both bare imports
     /// and `qw<...>` style import lists with all delimiter types.
-    fn find_import_source(&self, ast: &crate::ast::Node, symbol_name: &str) -> Option<String> {
+    pub(crate) fn find_import_source(
+        &self,
+        ast: &crate::ast::Node,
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as narrower version of #5027, which touches the same files with more coverage.

Issue #3472, opened 2026-04-10
[module-resolution] Import list symbols not extracted for bareword resolution


EffortlessSteven:
## Problem

When a module uses `qw(...)` import lists, perl-lsp parses the module name but does not extract and track which symbols are actually imported. This breaks bareword function calls that depend on imported symbols.

## Current Behavior

`perl-module-import` extracts only the module name from `use` statements:
- `use Carp qw(croak);` → extracts "Carp" (correct)
- But does NOT extract that "croak" is imported

Result: bareword `croak()` calls cannot be resolved to Carp::croak.

## Example That Fails

File A.pm:
```perl
package A;
use Exporter 'import';
our @EXPORT = qw(helper);
sub helper { 'result' }
```

File B.pl:
```perl
use A;
helper();  # CTRL+Click fails — bareword 'helper' has no definition
```

## Why This Matters

1. **Go-to-definition breaks** on imported functions
2. **Completion is incomplete** — doesn't suggest imported symbols
3. **Scope diagnostics are wrong** — marks legitimate imported symbols as undefined
4. **Real developers do this constantly** — import patterns are fundamental

## Evidence

- `crates/perl-module-import/src/lib.rs` parses import head, returns only token (module name)
- `crates/perl-module-import-match/src/lib.rs` only matches lines, doesn't extract symbols
- No crate extracts the argument list to `qw(...)`, `(...)`, or `'...'` forms
- Import tracking stops at module resolution; symbol visibility is not updated

## Import Forms NOT Tracked

1. `use Carp qw(croak confess);` — qw list
2. `use Carp ('croak', 'confess');` — parens list
3. `use Carp 'croak';` — single quoted string
4. `use Carp;` — default @EXPORT (requires analyzing Carp's exports)
5. `use Carp ();` — empty import (should import nothing)
6. `use POSIX qw(:sys_wait_h);` — tag imports (needs %EXPORT_TAGS resolution)

## Related Code Locations

- `crates/perl-module-import/src/lib.rs` — Line 99-120, only extracts module name
- `crates/perl-semantic-analyzer/src/analysis/symbol.rs` — No import symbol tracking
- `crates/perl-workspace-index/src/workspace/workspace_index.rs` — No import symbol table
- `crates/perl-lsp-navigation/src/lib.rs` — No cross-module symbol resolution

## Implementation Notes

This is blocking for issues #3409, #3413, #3414, #3415, #3416 (import/export framework issues).

The fix requires:
[...]

PR #5022, opened 2026-04-23
fix(declaration): resolve bareword imports in `use ... qw(...)` lists (#3472)

EffortlessSteven:
### Motivation

- Go-to-definition and related features treated barewords inside `use Module qw(...)` import lists as package navigation rather than imported symbols, so imported functions and tag-expanded members did not navigate to their exporter definitions.

### Description

- Add a source-aware internal resolver `symbol_at_cursor_internal(..., source_text)` and a public `symbol_at_cursor_with_source(...)` API to disambiguate barewords inside `use` import lists. 
- Implement `token_at_offset_in_text` and `use_args_import_symbol` helpers and extend `NodeKind::Use` handling to recognize bareword imports (including `qw(...)` and tag expansions) and return `SymbolKey { pkg: module, name: symbol }` when appropriate. 
- Wire LSP handlers to use the new API by passing `&doc.text` in definition/rename/references/moniker paths so editor flows use the source-aware resolution. 
- Add regression tests covering `use Module qw(symbol)` and tag-based `use POSIX qw(:tag SYMBOL)` bareword resolution in `cross_file_goto_definition_tests.rs`.

### Testing

- Ran `cargo test -p perl-lsp-rs --test cross_file_goto_definition_tests use_qw_` and the two new `use qw` tests passed. 
- Ran `cargo check --all-targets -p perl-semantic-analyzer -p perl-lsp-rs` and it completed successfully. 
- Ran `cargo xtask fmt` to format changes successfully. 
- Note: `just pr-fast` is not available in this environment (`/bin/bash: line 1: just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b8aae34c83339cb2546c05adf79b)

PR #5027, opened 2026-04-23
fix(lsp): resolve bare require/import symbols in goto-def (#3472)

EffortlessSteven:
### Motivation
- Bareword subs imported via a static `require Module; Module->import('sym')` pattern did not resolve with cross-file goto-definition, leaving imported symbols unresolved.

### Description
- Reused the existing import-source discovery by exposing `find_import_source` as `pub(crate)` in `moniker.rs` so other handlers can consult it.
- Added a workspace-definition fallback in the definition path (`navigation.rs`) that, for bare `Sub` symbols with no sigil, checks `find_import_source` and routes to the module definition when available.
- Added a regression integration test that writes `lib/My/Exporter.pm`, does `require My::Exporter; My::Exporter->import('greet');` and asserts goto-definition on `greet()` resolves to `My/Exporter.pm` in `cross_file_goto_definition_tests.rs`.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-rs`, which passed. 
- Ran the new integration test with `cargo test -p perl-lsp-rs --test cross_file_goto_definition_tests go_to_definition_on_require_manual_import_bareword_navigates_to_exporter_sub -- --nocapture`, which passed.
- Ran formatting with `cargo xtask fmt` as part of verification.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b877cee88333959cbf69a7cfcfab)


EffortlessSteven (commented):
**Deep Review — APPROVE (with one flag)**

**Verdict: APPROVE**

**What was verified:**

1. **moniker.rs** —  visibility change from  to  is minimal and correct. The function is on  and navigation.rs calls it as `self.find_import_source` which is valid.

2. **navigation.rs** — The 22-line addition is inside `#[cfg(feature = "workspace")]`, inside `if let Some(coordinator)`, inside the existing `if let Some(symbol_key)` block. It fires only when `find_symbol_key_definition_location` has already returned `None` for the symbol. The guard `symbol_key.kind == Sub && sigil.is_none()` prevents false-positive matches on variables (`$foo`, `@arr`). Fallback control flow is correct.

3. **Test coverage** — The 66-line test covers the `require Module; Module->import('sym')` path with a realistic workspace setup (separate lib/My/Exporter.pm + main.pl, barrier before request). The `use Module qw(sym)` path is **not tested in this PR** but is already covered by the pre-existing Test 2b (`go_to_definition_on_imported_function_navigates_to_source_module`, line 206), so there is no gap.

4. **Overlap with #5022** — No conflict. PR #5022 touches navigation.rs at line ~1334 (changes `symbol_at_cursor` to `symbol_at_cursor_with_source`). PR #5027 adds a new block *after* the existing resolution attempt at line ~1355. These are sequential, non-overlapping hunks. A merge conflict is possible if #5022 merges first and shifts line numbers, but the logical diff will apply cleanly since they touch distinct lines.

**One correctness concern (non-blocking):** the test asserts `uri.contains("My/Exporter.pm") || uri.contains("My%2FExporter.pm")` — the second alternative uses `%2F` (slash-encoded) but paths on disk use `/` not `%2F` in file URIs on most platforms. The `%2F` branch is dead on Linux/macOS and borderline on Windows. This is harmless (the first branch will match) but the assertion message is slightly misleading. Not a correctness defect.

EffortlessSteven (commented):
**Deep Review — APPROVE (with one flag)**

**Verdict: APPROVE**

**What was verified:**

1. **moniker.rs** — `find_import_source` visibility change from `fn` to `pub(crate) fn` is minimal and correct. The function is on `impl LanguageFeatures` and navigation.rs calls it as `self.find_import_source` which is valid.

2. **navigation.rs** — The 22-line addition is inside `#[cfg(feature = "workspace")]`, inside `if let Some(coordinator)`, inside the existing `if let Some(symbol_key)` block. It fires only when `find_symbol_key_definition_location` has already returned `None` for the symbol. The guard `symbol_key.kind == Sub && sigil.is_none()` prevents false-positive matches on variables. Fallback control flow is correct.

3. **Test coverage** — The 66-line test covers the `require Module; Module->import('sym')` path with a realistic workspace setup (separate lib/My/Exporter.pm + main.pl, barrier before request). The `use Module qw(sym)` path is not tested in this PR but is covered by the pre-existing Test 2b (`go_to_definition_on_imported_function_navigates_to_source_module`, line 206), so there is no gap.

4. **Overlap with #5022** — No conflict. PR #5022 touches navigation.rs at line ~1334 (changes `symbol_at_cursor` to `symbol_at_cursor_with_source`). PR #5027 adds a new block after the existing resolution attempt at line ~1355. These are sequential, non-overlapping hunks. A merge conflict is possible if #5022 merges first and shifts line numbers, but the logical diff will apply cleanly since they touch distinct lines.

**One correctness concern (non-blocking):** the test asserts `uri.contains("My/Exporter.pm") || uri.contains("My%2FExporter.pm")` — the `%2F` alternative uses slash-encoded form, but file URIs use forward slashes not `%2F`. The second branch is dead on Linux/macOS. This is harmless (the first branch matches) but the assertion message is slightly misleading. Not a correctness defect.

Issue #3489, opened 2026-04-10
[diagnostics] eval STRING with dynamic pragmas not conservatively handled


EffortlessSteven:
## Problem
The pragma tracker encounters `eval { ... }` (string eval) that dynamically loads pragmas and modules. These evals run at runtime, so static analysis cannot determine which pragmas are enabled inside them. The current code appears to track pragmas in eval blocks as if they were compile-time, which is incorrect.

## Reproducer
```perl
use strict;
if ($ENV{DEBUG}) {
    eval "use Data::Dumper";  # String eval: at runtime, conditionally
}
# After eval, strict is still in effect, but LSP may not know that
```

More problematically:
```perl
eval {
    no strict;  # This only applies inside the eval STRING, not outside
    $$var = 1;  # OK inside eval
};
$$var = 1;  # Should WARN about strict refs (outside eval)
```

## Expected behavior
- String evals (`eval STRING`) are runtime and should NOT affect static pragma tracking
- Block evals (`eval { ... }`) create a separate scope for pragmas
- After a block eval exits, pragmas revert to pre-eval state
- Missing-strict diagnostic in post-eval code should respect pragmas from before the eval

## Actual behavior
The pragma tracker's handling of `eval` blocks is unclear. The build_ranges match statement (line 153) does not have explicit handling for eval. If eval is a type of block, pragmas inside might escape the eval scope, polluting the outer scope.

If `eval { no strict 'refs' }` is encountered, the tracker might:
1. Incorrectly propagate the pragma change outside the eval
2. Or fail to track pragmas inside the eval at all

## Root cause
`crates/perl-pragma/src/lib.rs` lines 148-276

The match statement (line 153) likely doesn't distinguish eval blocks from regular blocks. Eval blocks should behave like `Block` nodes (save/restore state) but may not be implemented that way.

Additionally, string evals (`eval EXPR`) are fundamentally unparseable by static analysis, so they should be either:
1. Ignored (pragmas don't leak out)
2. Or conservatively assumed to enable/disable pragmas (warn the developer)

## Suggested fix
1. **Explicitly handle eval**: If `NodeKind::Eval` exists, add a case that treats it like Block (save/restore):
   ```rust
   NodeKind::Eval { body, .. } => {
       let saved_state = current_state.clone();
       Self::build_ranges(body, current_state, ranges);
       *current_state = saved_state;
   }
   ```
[...]

PR #5029, opened 2026-04-23
fix(pragma): scope implicit strict modules lexically (#3489)

EffortlessSteven:
### Motivation
- Implicit strict/warnings providers (e.g. `Moose`, `Moo`, `Modern::Perl`) declared via `use` could be treated scope-unaware and therefore incorrectly suppress file-level missing-`use strict`/`use warnings` diagnostics when declared inside scoped bodies like `eval { }`.
- The pragma tracking and diagnostics needed to represent these implicit-framework effects lexically so scope restoration behaves uniformly.

### Description
- Extend `PragmaTracker` to treat OO/framework modules as tracked pragma-affecting modules by updating `is_tracked_pragma_module` and adding explicit handling for those modules in `PragmaTracker::build_ranges` so their `use` (including conditional `use if`) sets `strict_*` and `warnings` in the lexical state.
- Make the strict/warnings lint (`check_strict_warnings`) rely on `PragmaTracker::build` / `state_for_offset` as the authoritative, scope-aware source of truth and remove the prior scope-unaware AST-based implicit-module shortcut.
- Update tests to assert the corrected behaviour: replace the old known-limitation expectation and add `perl-pragma` unit tests that verify implicit-module enablement and that `use Moose` inside `eval { }` does not leak to top-level state.

### Testing
- Ran `cargo test -p perl-pragma` and all tests passed.
- Ran `cargo test -p perl-lsp-rs-core strict_warnings` and the strict/warnings tests passed.
- Ran `cargo check --all-targets -p perl-pragma -p perl-lsp-rs-core` and it completed successfully.
- Ran `cargo clippy -p perl-pragma -p perl-lsp-rs-core -- -D warnings` and `cargo xtask fmt` and both succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b8c091088333b06efc0bbf33580e)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/diagnostics/lints/strict_warnings.rs
+++ b/crates/perl-lsp-rs-core/src/providers/diagnostics/lints/strict_warnings.rs
@@ -65,25 +65,13 @@ pub fn check_strict_warnings(node: &Node, diagnostics: &mut Vec<Diagnostic>) {
     // This avoids the false-negative from .any() which sees eval-interior ranges.
     // signatures_strict is included to honour `use feature 'signatures'` (#4038).
     let top_level_state = PragmaTracker::state_for_offset(&pragma_map, usize::MAX);
-    let mut has_strict = top_level_state.strict_vars
+    let has_strict = top_level_state.strict_vars
         || top_level_state.strict_subs
         || top_level_state.strict_refs
         || top_level_state.signatures_strict;
-    let mut has_warnings = top_level_state.warnings;
+    let has_warnings = top_level_state.warnings;
 
-    // OO frameworks that implicitly provide strict+warnings
-    const IMPLICIT_STRICT_MODULES: &[&str] = &[
-        "Moo",
-        "Moose",
-        "MooseX::StrictConstructor",
-        "Modern::Perl",
-        "Dancer2",
-        "Catalyst",
-        "Mojolicious",
-        "Mojo::Base",
-    ];
-
-    // Detect OO implicit strict modules and misspelled pragmas.
+    // Detect misspelled pragmas.
     // The strict/warnings arms are intentionally absent: state_for_offset above
     // is the authoritative source of truth. Walking the full AST for strict/warnings
     // would bypass lexical scoping (finding eval-block or sub-scoped pragmas).
@@ -92,9 +80,6 @@ pub fn check_strict_warnings(node: &Node, diagnostics: &mut Vec<Diagnostic>) {
             if module.starts_with('v') || module.chars().next().is_some_and(|c| c.is_ascii_digit())
             {
                 // Version pragmas are already reflected in the shared pragma map.
-            } else if IMPLICIT_STRICT_MODULES.contains(&module.as_str()) {
-                has_strict = true;
-                has_warnings = true;
             } else if module != "strict" && module != "warnings" {
                 // Check for misspelled pragmas (strict/warnings are in pragma_map)
                 check_misspelled_pragma(module, n, diagnostics);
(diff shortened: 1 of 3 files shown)


EffortlessSteven (commented):
**Deep Review — REQUEST CHANGES**

**Verdict: REQUEST CHANGES**

**Critical issue — eval block test has a logic error in the after-eval assertion:**

The test `implicit_strict_module_in_eval_does_not_leak_out` asserts:
```rust
let after_eval = PragmaTracker::state_for_offset(&map, 24);
assert!(after_eval.strict_vars);       // from use strict 'vars' at offset 23-42
assert!(!after_eval.strict_subs);
assert!(!after_eval.strict_refs);
assert!(!after_eval.warnings);         // Moose's warnings must not leak
```

The eval contains `use Moose` at offsets 8-18 (inside block at 6-20, eval at 0-22). After the eval (offset 24), there is `use strict 'vars'` at 23-42. The test expects `strict_vars=true, strict_subs=false, strict_refs=false, warnings=false`. This is correct — Moose's strict_subs/refs/warnings stay zeroed after the eval restores state, and only strict_vars is re-enabled by the subsequent explicit pragma.

**This logic is correct.** However there is a deeper Perl semantics concern:

**Concern — `use Moose` inside `eval { }` is not actually lexically scoped in Perl.** In real Perl, `use Module` inside `eval { }` runs the module's `import()` at compile time (since `use` is a BEGIN block). Moose's strict/warnings effects apply to the *surrounding file scope* because they are compile-time side effects. The LSP model treats them as lexically scoped (bounded by the eval block) which produces *incorrect diagnostics* — a file with `eval { use Moose; }` at top level should NOT get PL100/PL101 since Moose's `import` runs at compile time and affects the file. The new test asserts the opposite of real Perl behavior.

**Recommended fix:** Verify against `perldoc -f use` semantics: `use Foo` is equivalent to `BEGIN { require Foo; Foo->import(...) }`. Since BEGIN blocks run at compile time before the surrounding eval is parsed as a runtime construct, treating `use Moose` inside `eval {}` as lexically bounded to the block is incorrect.

The -31 deletion in strict_warnings.rs removes the scope-unaware fallback list, which is the right structural move. But the replacement in perl-pragma/lib.rs introduces the same semantic error (treating OO pragma effects as lexically bounded to whatever block contains the `use` statement).

**Suggested resolution:** Either (a) keep OO framework modules as file-scope-always (ignore block depth for `use Moose`/etc.), or (b) document that the LSP model is intentionally conservative (never suppress diagnostics for OO modules inside non-top-level blocks). Silently emitting PL100/PL101 for `eval { use Moose; }; ...` is a false positive and will frustrate Catalyst/Mojolicious users.

EffortlessSteven (commented):
**Deep Review — REQUEST CHANGES**

**Verdict: REQUEST CHANGES**

**Critical correctness issue — Perl semantics of `use` inside `eval {}`:**

In real Perl, `use Module` is syntactic sugar for `BEGIN { require Module; Module->import(...) }`. BEGIN blocks run at compile time, before the enclosing runtime construct (like `eval {}`) is evaluated. This means `use Moose` inside `eval { }` has its strict/warnings effects applied to the surrounding file scope at compile time — not bounded to the eval block at runtime.

The test `implicit_strict_module_in_eval_does_not_leak_out` asserts that PL100/PL101 fire for code like `eval { use Moose; }; my $x = 1;`. In real Perl this would NOT produce missing-strict or missing-warnings errors because `use Moose` ran at compile time and already activated strict+warnings for the file. The LSP model in this PR treats OO pragma effects as lexically bounded to whatever block contains the `use` statement, which is incorrect.

**Consequence:** Catalyst, Mojolicious, Moo users who write `eval { use Moose; }` or have framework modules inside any non-program-level block will now see false-positive PL100/PL101 diagnostics.

**What is correct in this PR:**
- The structural move of removing the scope-unaware fallback list from strict_warnings.rs (-31 lines) is right.
- Adding OO modules to `is_tracked_pragma_module` and `build_ranges` in perl-pragma is the right architectural direction.
- The lexical scoping via save/restore in `build_scoped_body` is correct for `use strict` / `use warnings` (which are genuinely lexical). The error is applying the same model to `use Moose` which is not lexical in practice because Moose's import runs at compile time.

**Recommended fix:** For OO framework modules in `build_ranges`, apply the state change unconditionally without entering a scoped body — treat `use Moose` as having file-scope effect regardless of block depth, same way Perl actually processes it. This preserves the structural improvement while correcting the semantics.

EffortlessSteven:
Deep review — REQUEST CHANGES.

## Critical Perl semantics error

In Perl, `use Module` is equivalent to `BEGIN { require Module; Module->import(...) }`. BEGIN blocks execute at **compile time**, before any enclosing runtime block (`eval {}`, `if () {}`, `do {}`) is evaluated. This means `use Moose` inside `eval { use Moose; }` applies strict/warnings to the **surrounding file scope** at compile time — not bounded to the eval block.

The new test `implicit_strict_module_in_eval_does_not_leak_out` asserts that PL100/PL101 fire for `eval { use Moose; }; my $x = 1;`. In real Perl, that code would NOT produce missing-strict/warnings diagnostics because Moose's import ran at compile time.

**Consequence:** Catalyst, Mojolicious, Moo/Moose users who put their framework `use` inside any non-top-level block (including `eval {}`, `BEGIN {}`, `if () {}`) will see false-positive PL100/PL101.

## What IS structurally correct here

- Removing the scope-unaware fallback list from `strict_warnings.rs` (-31 lines) and moving OO module tracking into `perl-pragma` is the right architectural direction.
- The save/restore lexical-scope mechanism is correct for genuine lexical pragmas (`use strict`, `use warnings`).

## Recommended fix

In `build_ranges` in `crates/perl-pragma/src/lib.rs`, apply the OO module state change to `current_state` **without scoping it to the node's block boundary**. Treat `use Moose` as file-scope regardless of nesting depth, matching how Perl actually processes it at compile time.

The genuinely lexical `use strict`/`use warnings` correctly stay lexical. It's only the OO module list that should be file-scope.

Adjust the test `implicit_strict_module_in_eval_does_not_leak_out` accordingly — it's asserting wrong behavior.

EffortlessSteven:
Closing — superseded by **#5086** (fix(lints): scope implicit Moose/Moo strict detection to file level), which landed in the later Codex wave. #5086's approach matches the reviewer-deep finding on this PR: treat OO frameworks as file-scope (compile-time BEGIN semantics) rather than lexical. Cleaner, more focused fix (+36/-16 to strict_warnings.rs) than reworking this PR's perl-pragma changes.

PR #5030, opened 2026-04-23
fix(pragma): parse conditional use/no if pragmas correctly (#3489)

EffortlessSteven:
### Motivation
- `use if` / `no if` (and `unless`) were being parsed such that a following numeric token could be treated as a module version (e.g. `if 1`), which prevented the pragma tracker from recognizing conditional pragmas and caused false PL100/PL101 diagnostics.
- `eval "..."` (string eval) must be treated conservatively so runtime string contents cannot suppress file-level `use strict`/`use warnings` diagnostics.

### Description
- Skip optional numeric-version parsing when the parsed module is the keyword-named pragmas `if` or `unless` in both the `use` and `no` parsing paths in `crates/perl-parser-core/src/engine/parser/declarations.rs`.
- Add unit tests in `crates/perl-lsp-rs-core/src/providers/diagnostics/lints/strict_warnings.rs` asserting conservative handling of `eval "..."` (string eval) and that `use if 1, 'strict'` / `use if 1, 'warnings'` correctly suppress missing-pragmas diagnostics at file scope.
- Update assertion messaging to include unexpected diagnostic codes when the conditional `use if` test initially failed, and keep formatting/cleanup via `cargo xtask fmt`.

### Testing
- Ran `cargo test -p perl-lsp-rs-core strict_warnings::tests::eval_string_containing_pragmas_is_handled_conservatively -- --nocapture`, which passed.
- Ran `cargo test -p perl-lsp-rs-core strict_warnings::tests::conditional_use_if_pragmas_suppress_file_level_missing_diagnostics -- --nocapture`, which passed after the parser fix.
- Ran parser-focused tests `cargo test -p perl-parser-core use_if_ -- --nocapture` and formatting `cargo xtask fmt`, both of which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b8c091088333b06efc0bbf33580e)


EffortlessSteven:
Deep review — APPROVE.

## PR #5030 analysis

**Logic: CORRECT**

The two identical patches in `declarations.rs` are in `parse_use` and `parse_no` — symmetric functions, not duplicated code. Both need the same guard.

Root cause: `use if` was being lexed as module `"if"`, then the subsequent `1` (Number) was consumed as a version suffix, producing `module: "if 1"`. With the guard `module != "if" && module != "unless"`, the number stays in the token stream for the condition-args parser at line 580.

The PragmaTracker in `crates/perl-pragma/src/lib.rs` (lines 480-503) already handles `module == "if"` and unconditionally sets `strict_vars`/`warnings`. That's a pre-existing conservative design — static analysis can't evaluate runtime conditions.

**Tests:**
- `conditional_use_if_pragmas_suppress_file_level_missing_diagnostics`: semantically correct. `use if 1, 'strict'` in real Perl DOES enable strict, so suppressing PL100/PL101 is right. Would have failed before the fix.
- `eval_string_containing_pragmas_is_handled_conservatively`: correct. String evals can't be statically analyzed.

**Regression risk: LOW.** The `unless` guard is also correct (`no unless COND, 'strict'` is valid Perl).

**Follow-up (not a blocker):** `use if 0, 'strict'` currently suppresses PL100 (false negative). Pre-existing limitation — worth filing separately.

LGTM.

PR #5035, opened 2026-04-23
fix(ci): deepen PR smoke verification with pr-fast (#0000)

EffortlessSteven:
### Motivation
- Increase PR smoke depth so the default PR run performs the consolidated `pr-fast` gate (format, README heading checks, publish-closure checks, core clippy/tests and other fast validations) instead of only `clippy-core` and `test-core`.

### Description
- Replace the inline PR smoke steps in `.github/workflows/ci.yml` with a single `just pr-fast` invocation and remove the now-duplicated `cargo xtask fmt --check` step.

### Testing
- Attempted to run `just pr-fast` locally but the container lacks `just`, so the execution could not be performed (failure: `/bin/bash: line 1: just: command not found`).
- Attempted to validate the modified YAML with `python3` and `yaml.safe_load`, but `PyYAML` is not installed so that check also failed.

------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,9 +77,6 @@ jobs:
         with:
           tool: just
 
-      - name: Check formatting
-        run: cargo xtask fmt --check
-
       - name: Cache cargo dependencies
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4  # v2.9.1
         with:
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing — this PR removes the `Check formatting` step and replaces the scoped PR Smoke logic with `just pr-fast`, which regresses the scope-aware tier-wiring that just landed in #5005. The fmt-check removal is also a regression. The broader 'deeper PR smoke' goal is already tracked in #5005's follow-up improvements.

PR #5036, opened 2026-04-23
ci(gates): deepen all-targets verification (#0000)

EffortlessSteven:
### Motivation
- Ensure the compile-only CI gate also exercises feature-gated code paths because the previous check only validated default-feature builds.

### Description
- Modified `check-all-targets` in `justfile` to run `cargo check --workspace --all-targets --locked` for default features and a second `cargo check --workspace --all-targets --all-features --locked`, and updated the gate comments to reflect the deeper verification.

### Testing
- Executed `cargo check --workspace --all-targets --locked` and `cargo check --workspace --all-targets --all-features --locked` locally; both ran but the workspace build failed due to pre-existing unrelated errors (e.g. an unresolved import in `crates/perl-parser` and type-mismatch failures in `crates/perl-lsp-rs`).
- Attempted `just check-all-targets` but `just` is not installed in this environment, so the `just` invocation could not be run here.

------


EffortlessSteven:

Logic is correct: adds `--all-features` to `check-all-targets` recipe to catch feature-gated compilation failures. The two-pass approach (default features, then all-features) is sound.

No overlap with #5037 — that PR adds a gate-policy metadata entry, this adds the actual build step to the justfile recipe. They are orthogonal layers but should be merged together so the catalog entry is consistent. I pushed a fix to #5037 aligning its `command` field to `just check-all-targets` instead of the raw cargo invocation.

No issues with the justfile change itself. Approved.

PR #5037, opened 2026-04-23
ci(gates): deepen merge-gate verification with all-targets compile (#0000)

EffortlessSteven:
### Motivation

- Merge-gate verification did not compile integration-test and benchmark targets, allowing non-lib target bit-rot to slip through. 
- Adding a workspace-wide all-targets compile step ensures integration/bench and test compilation failures surface earlier in CI.

### Description

- Updated `.ci/gate-policy.yaml` last-updated timestamp to `2026-04-23` and added a new required merge-gate entry `compile_all_targets`. 
- The new gate runs `cargo check --workspace --all-targets --locked` with an explicit `timeout_seconds`, `budgets` metadata, and tags to classify it as a workspace-level compile/integration/bench check. 

### Testing

- `cargo xtask gates --list` was run and shows the new `compile_all_targets` gate present, and that listing succeeds. 
- `cargo xtask gates --tier merge-gate --gate compile_all_targets` was executed to validate the gate, and it failed as expected due to an existing compile error in `crates/perl-parser` (`unresolved import super::incremental_edit`), demonstrating the gate surfaces non-lib compile issues.

------


EffortlessSteven:

The `compile_all_targets` gate entry adds useful metadata to the catalog. Pushed one fix: the `command` field was hardcoded as `cargo check --workspace --all-targets --locked` but after #5036 the `just check-all-targets` recipe runs two cargo commands (default + all-features). Updated to `just check-all-targets` so the policy catalog tracks the full recipe.

Note: the actual CI enforcement comes from the `.github/workflows/ci.yml` job `check-all-targets` (which calls `just check-all-targets`), not from the gate-policy file. The policy file is catalog/documentation. Approved.

PR #5043, opened 2026-04-23
test(xtask): stabilize publish-closure edge cases (#0000)

EffortlessSteven:
### Motivation

- The `publish-closure` edge-case tests were brittle due to assumptions about specific allowlisted crate names and a hardcoded numeric range for the crate count. 
- Make the tests deterministic and robust against allowlist changes so they don't fail as the workspace evolves.

### Description

- Updated `xtask/tests/publish_closure_edge_cases.rs` to use a known allowlisted crate (`perl-parser`) for filtering tests instead of a stale crate name. 
- Reworked the crate-count regression test to parse the CLI-reported checked count and compare it against the actual `[workspace.metadata.publish.allow]` array read from the workspace `Cargo.toml` (via `env!("CARGO_MANIFEST_DIR")`).
- Added helper functions `parse_checked_count` and `publish_allowlist_count` in the same test file to perform the parsing and workspace manifest lookup.

### Testing

- Ran `cargo test -p xtask --test publish_closure_edge_cases` and the entire test file passed (all tests succeeded). 
- Ran `cargo check --all-targets -p xtask` and it completed successfully. 
- Ran `cargo clippy -p xtask -- -D warnings` and it completed with no lints treated as warnings. 

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9ba462e348333bbf1e13e8fd480b9)


EffortlessSteven:

Logic verified:
- `parse_checked_count` + `publish_allowlist_count` approach is sound: `crates_to_check.len()` in `publish_closure.rs` always equals the full allowlist length because missing workspace crates use `continue` without decrementing the count, so the exact-match assertion is correct and self-updating.
- All three `perl-error` references (lines 144/146 and 163) correctly replaced with `perl-parser`, which is confirmed present in the allowlist. The original `perl-error` was absorbed into `perl-parser-core` (Wave D) and removed from the allowlist.
- Duplicate-flag rejection test is correctly premised: `crate_name: Option<String>` is non-repeatable in clap, so the test exercises the right error path.

Security: CLEAN. Performance: CLEAN. Regression risk: LOW (test-only file).

PR #5047, opened 2026-04-23
fix(perl-regex): reduce brace false positives in nested quantifier heuristic (#0000)

EffortlessSteven:
### Motivation

- `detect_nested_quantifiers` treated any `{` after a quantified group as a quantifier which produced false positives when the brace was literal or malformed.
- This caused incorrect advisory diagnostics for common patterns containing literal braces or non-quantifier brace text.

### Description

- Add `is_brace_quantifier` helper to validate brace quantifier shape (`{n}` or `{n,m}` with closing `}`) using the existing `CharIndices` iterator.
- Update `detect_nested_quantifiers` to call `is_brace_quantifier` before treating `{` as a quantifier.
- Add two regression tests verifying that a literal brace after a group (`(a+){foo}`) and an unterminated/malformed brace (`(a+){2,5`) are not flagged as nested quantifiers.

### Testing

- Ran `cargo test -p perl-regex` and all unit tests passed.
- Ran `cargo check --all-targets -p perl-regex` and it completed successfully.
- Ran `cargo clippy -p perl-regex` and observed no lints failing for the changed crate.
- `just pr-fast` was not available in the environment (`just: command not found`) so it could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b997f3b483338baeb79ce2ad5ede)


EffortlessSteven:

The `is_brace_quantifier` helper logic is correct. Traced all edge cases:
- `{foo}` — saw_digit=false, returns false (literal, correct)
- `{2,5` — no closing brace, returns false (correct)
- `{2}` — returns true (correct quantifier)
- `{2,}` — open-ended range, saw_digit=true, comma consumed, inner digit loop sees '}', peek matches => returns true (correct)
- `{}` — saw_digit=false, returns false (correct)

The iterator-consuming behaviour is safe: after `is_brace_quantifier` returns false for a literal brace, chars has consumed the non-quantifier lookahead characters, which is the right position for the outer loop to continue from.

One gap: no test for `{2,}` (open-ended range should flag). Pushed a test asserting `(a+){2,}` is detected as a nested quantifier.

PR #5053, opened 2026-04-23
fix(lexer): handle CR line endings in comment and POD scanning (#0000)

EffortlessSteven:
### Motivation

- The lexer previously only searched for LF when skipping line comments and scanning POD boundaries, which could cause comment/POD scanning to run to EOF on files using CR-only line endings.

### Description

- Use `memchr::memchr2` to detect either `\n` or `\r` when skipping line comments so CR, CRLF, and LF endings are handled.
- Treat the previous-byte line-start check for POD directives as matching either `\n` or `\r` so POD detection works for CR-terminated lines.
- When scanning for `=cut`, stop at either `\n` or `\r` and consume a single CRLF or single-CR/LF sequence as appropriate to avoid running past the end of the `=cut` line.
- Add a regression test `test_comment_skipping_with_cr_line_endings` that verifies lexing resumes after a CR-terminated comment line.

### Testing

- Ran the targeted unit test `cargo test -p perl-lexer test_comment_skipping_with_cr_line_endings`, which passed.
- Ran `cargo check --all-targets -p perl-lexer`, which passed.
- Ran `cargo clippy -p perl-lexer --all-targets -- -D warnings`, which passed.
- Ran `cargo xtask fmt`; formatting completed; `just pr-fast` was not executed in this environment because `just` is not installed.

------


EffortlessSteven:
## Deep Review: APPROVED

**Logic verified:**
- `memchr::memchr2(b'\n', b'\r', ...)` correctly finds CR, LF, and CRLF endings for comment skip. Position is left pointing at the terminator (consistent with original `\n`-only behavior; the existing `b'\r' | b'\n'` whitespace arm consumes it on the next iteration).
- POD start guard `matches!(bytes[prev], b'\n' | b'\r')` correctly detects start-of-line after CR-only endings.
- `=cut` scanner: the `i += 1` byte-by-byte scan correctly finds `=cut` after `\r`; the two-branch terminator consumption (CR, CRLF, LF) is complete.
- No banned patterns in production code.

**Fix-forward pushed:** Added `test_pod_skipped_with_cr_only_line_endings` — the original test only covered comment-skip, not POD block skipping. The new test verifies that `=pod\r...\r=cut\r` is fully consumed and tokens after `=cut` are correctly emitted.

**Pre-existing gate failure:** `wave_c_collapse_hardening` (97-count assertion) fails on this branch. This is a pre-existing test bug fixed by #5066 — not caused by this PR.

PR #5054, opened 2026-04-23
fix(parser-core): allow phase keywords as callable barewords (#0000)

EffortlessSteven:
### Motivation

- Phase keywords (`BEGIN`/`END`/`CHECK`/`INIT`/`UNITCHECK`) were being forced into phase-block parsing in statement position and thus rejected valid uses like callable barewords/subroutine calls (e.g. `CHECK();`).
- Allow both the existing label/phase-block forms and ordinary expression statements so real-world CPAN code that defines/uses these names continues to parse.

### Description

- Change statement dispatch in `crates/perl-parser-core/src/engine/parser/statements.rs` so phase-block parsing is only taken when the keyword is immediately followed by `{`, and label handling still applies when followed by `:`.
- Fall back to normal expression-statement parsing (`parse_expression_statement`) for phase keywords when not followed by `{` or `:` so calls like `CHECK();` parse cleanly.
- Add regression tests in `crates/perl-parser-core/tests/fix_phase_keyword_as_label_2389.rs` to cover callable usages (`CHECK();`, `BEGIN('arg');`) alongside existing label/block tests.

### Testing

- Ran `cargo test -p perl-parser-core --test fix_phase_keyword_as_label_2389`, which passed (12 tests, all OK). 
- Ran `cargo check --all-targets -p perl-parser-core`, which completed successfully. 
- Ran `cargo clippy -p perl-parser-core -- -D warnings`, which completed with no warnings. 
- Ran `cargo xtask fmt`; formatting step completed in this environment, while `just pr-fast` was not available (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b90e4e8483339eecacac7e8a3351)


EffortlessSteven:
## Deep Review: APPROVED

**Logic verified:**
- The dispatch now has three explicit arms: `{` → `parse_phase_block()`, `:` → `parse_keyword_as_label()`, anything else → `parse_expression_statement()`.
- `peek_second()` calls `next_token_from_lexer()` which skips `Whitespace`/`Newline` tokens, so `BEGIN\n{` correctly routes to `parse_phase_block()`.
- Existing regression tests confirm `CHECK { ... }` and `BEGIN { ... }` still parse as phase blocks.
- New tests cover `CHECK()` and `BEGIN('arg')` as expression statements.
- No banned patterns.

**Note on `BEGIN('arg')`:** Perl's `BEGIN` cannot actually be called as a regular sub at runtime (it's compile-time special), but accepting it at the parse level without an ERROR node is correct — the parser is not a semantic analyzer. Runtime rejection is a separate concern.

PR #5057, opened 2026-04-23
fix(parser-core): handle legacy typed my declarations (#0000)

EffortlessSteven:
### Motivation

- Legacy typed lexical declarations like `my Debconf::DbDriver $this = shift;` appear in system-installed Perl modules and caused parse errors, reducing parser compatibility with the Ubuntu system Perl corpus.

### Description

- Consume an optional leading type identifier before parsing `my`/`our`/`state` variable declarations by adding `consume_legacy_decl_type_constraint()` and invoking it prior to `parse_variable()` in `variables.rs`.
- The consumed legacy type is ignored in the AST (keeps behavior minimal and non-invasive). 
- Add a focused unit test `typed_variable_declaration_tests.rs` verifying that `my Debconf::DbDriver $this = shift;` parses without `ERROR` nodes. 
- Register the new test module in `parser/mod.rs`.

### Testing

- `cargo test -p perl-parser-core typed_variable_declaration_tests -- --nocapture` — passed.
- `cargo check --all-targets -p perl-parser-core` — passed.
- `cargo clippy -p perl-parser-core -- -D warnings` — passed.
- `cargo xtask fmt` — passed.
- `just pr-fast` was not run because `just` is not available in the environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b90e4e8483339eecacac7e8a3351)


EffortlessSteven:
## Deep Review: APPROVED

**Logic verified:**
- Heuristic in `consume_legacy_decl_type_constraint`: only fires when `peek_kind == Identifier` (not a sigil token) AND `peek_second` is a sigil token. This correctly handles `my Foo::Bar $x` without false-positive on `my $x`.
- Lexer tokenizes `Debconf::DbDriver` as a single `Identifier` (package-qualified names consumed in one token via the `while current_char == ':' && peek_char(1) == ':'` loop). The heuristic sees one identifier then the sigil.
- False-positive analysis: `my foo()` → peek_second is `LeftParen` (not in sigil set) → correctly NOT consumed. `my ($x)` → peek is `LeftParen` not `Identifier` → early return.
- `our Foo $bar` works because `our` shares the non-local declarator path.
- `#[cfg(test)]` attribute correctly precedes the new test module in `mod.rs`.
- sexp format `(variable $ name)` matches established repo pattern.

**Fix-forward pushed:** Added two regression tests:
1. `test_plain_my_declaration_not_affected` — verifies ordinary `my $x = 1` is not consumed by the heuristic (critical regression guard).
2. `test_legacy_typed_our_declaration_parses_without_error_node` — verifies `our Foo::Bar $baz` also parses cleanly.

All 3 tests pass locally.

**Pre-existing gate failure:** `local_sig_die` in `cpan_error_handling` fails on master. Not caused by this PR.

PR #5060, opened 2026-04-23
fix(parser): allow anonymous sub and local as assignment rhs (#5017)

### LINKED ISSUE #5017: test(parser): ast_anonymous_sub snapshot regression on master — my $code = sub { ... } splits incorrectly
The integration test `ast_snap::ast_anonymous_sub` fails on current master with a parser behavior regression.

## Symptom

`cargo test -p perl-parser --test ast_snap ast_anonymous_sub` fails.

## The regression

For input: `my $code = sub { my ($x) = @_; return $x * 2; };`

**Expected** (current snapshot):
```
(source_file (my_declaration (variable $ code)(anonymous_subroutine_expression (block ...))))
```
Single statement: `my $code = <anon-sub>`.

**Actual** (current parse):
```
(source_file (my_declaration (variable $ code)(missing_expression))
             (expression_statement (anonymous_subroutine_expression (block ...))))
```
Parser splits into two statements: `my $code = <missing>` + a dangling `sub { ... }` expression.

## Why CI doesn't catch it

Push-triggered CI runs `cargo test --workspace --lib --locked` — which skips integration tests under `crates/*/tests/`. The `--all-targets` compile gate (#4988) only compiles, doesn't run. So this snapshot regression has been green on CI but red on pre-push hooks.

## Priority

Medium — parser correctness regression for a common Perl idiom. Blocks local pre-push hooks until fixed.

## Route

Builder — either fix the parser to recover the old behavior, or accept the new snapshot if the new split is actually correct for some downstream reason (then update the .snap file).

EffortlessSteven:
Closes #5017.

## Motivation

The parser's missing-RHS recovery treated `sub` and `local` as unconditional statement boundaries, which caused valid RHS forms like:

- `my $code = sub { ... };` — the existing snapshot regression from #5017
- `local $SIG{__WARN__} = sub { };`
- `local(*RS) = local(*/);`

to be misparsed as `MissingExpression` + a dangling statement.

## Description

Short-circuit `is_infix_rhs_absent` when the next token is `sub` that starts an anonymous sub or `local` used as an RHS expression.


EffortlessSteven:
## Deep Review — PR #5060

### Logic: CORRECT

The core fix is sound:

1. **Named sub guard is correct.** `next_token_starts_anonymous_sub()` peeks the second token — for `sub foo { }`, peek_second gives `Identifier` (the name), which is NOT `{`/`(`/`:`, so the function returns `false`. `is_infix_rhs_absent` then falls through to the `matches!` which includes `Sub`, correctly treating named sub as a statement boundary.

2. **`Local` removal is correct.** `local X` is a valid Perl expression and legitimate as an assignment RHS (e.g., `local(*RS) = local(*/);`). Removing it from `is_infix_rhs_absent` is semantically correct. `Local` remains in `is_delimiter_recovery_point` and `is_sync_point` for other error recovery paths.

3. **`sub :attr { }` handled.** `:` is in the `next_token_starts_anonymous_sub` match — attribute-qualified anonymous subs work.

4. **`sub;` (forward decl) safe.** peek_second returns `;`, not in the match, returns `false`, falls through to `Sub` in `is_infix_rhs_absent` — still a statement boundary.

5. **Snapshot already correct.** Both master and PR branch have identical `ast_anonymous_sub.snap` content (the expected correct output). The fix makes the parser produce the right parse to match the pre-existing snapshot.

### Tests: IMPROVED (fix-forward)

Added 3 commits to this branch:
- `test(parser): add regression + local-as-rhs edge case tests (#5017)`: adds `test_no_recovery_for_my_code_eq_anon_sub` (lib-level regression for `my $code = sub { ... }` from #5017 — complements the snapshot test which CI skips with `--lib`) and `test_local_as_assignment_rhs` (tests `local(*RS) = local(*/);` mentioned in PR body but untested).
- `fix(tests): update parser_panic_mode_recovery to Phase 2 recovery behavior`: two integration tests in `parser_panic_mode_recovery.rs` were asserting `NodeKind::Error` for `my $x = ;` but Phase 2 now produces `VariableDeclaration{MissingExpression}`. Updated to accept both forms. These tests were already failing in the PR before my commits (pre-push hook didn't run on original push).

### Regression Risk: LOW

The `is_infix_rhs_absent` change is guarded behind a lookahead that requires exactly `{`, `(`, or `:` to follow `sub`. All other `sub X` forms fall through to existing behavior.

EffortlessSteven:
## Deep Review Wrapup — PR #5060

**Summary:** reviewer-deep completed correctness review of parser anonymous-sub/local fix. 4 commits pushed, `deep-reviewed` label set.

### Ending State

**PR improved:** https://github.com/EffortlessMetrics/perl-lsp/pull/5060

**Changes made:**
- `test(parser): add regression + local-as-rhs edge case tests (#5017)` — lib-level regression test for `my $code = sub { ... }` (complements snapshot which CI skips) + `local(*RS) = local(*/);` test from PR body
- `style: cargo fmt (error_recovery_tests.rs)` — format fix triggered by pre-push hook
- `fix(tests): update parser_panic_mode_recovery to Phase 2 recovery behavior` — 2 integration tests asserting `NodeKind::Error` for `my $x = ;` were stale; Phase 2 produces `VariableDeclaration{MissingExpression}`; updated to accept both forms

**Status:** Approved (comment-only — can't approve own PR). `deep-reviewed` set. `needs-deep-review` removed.

**Next step:** ready for `/pr-ready` to exit draft + set `merge-ready`, then green-ci.

### Retrospective

The pre-push hook running integration tests revealed two pre-existing failures in `parser_panic_mode_recovery.rs` that the original push didn't catch — the `--all-targets` compile gate (#4988) compiles but doesn't run tests. This is the exact gap the CI docs warn about. Next time, check for pre-existing integration test failures on the PR branch before running the hook for the first time, to separate pre-existing from newly-introduced failures.

The `peek_second()` EOF safety was easy to verify because the token_stream implementation uses `.ok()` mapping — the pattern is consistent across all callers. What would have helped: a comment in `next_token_starts_anonymous_sub` noting that `peek_second().ok()` returns `None` at EOF, making the behavior explicit.

**Breadcrumbs:**
- `parser_panic_mode_recovery.rs` integration tests have a Phase 2 / Phase 1 inconsistency (some still say "Error node" when they mean "VariableDeclaration+MissingExpression") — there may be more stale assertions in that file beyond what I fixed
- `test_deref_hash_subscript_regex_key` in `test_edge_cases_deep.rs` is failing on master (`${$ref}{m}` parse error) — unrelated to this PR, needs a follow-up scout

EffortlessSteven:
## Green TDD: Edge Case Tests Added

**Tests added:** 6 new edge case tests for anonymous sub and local as RHS

| Test | Edge case covered |
|------|-------------------|
| `test_nested_anon_sub_in_my_declaration` | Nested `my $a = sub { my $b = sub { }; }` without spurious recovery |
| `test_list_assignment_with_anon_sub` | `@arr = (sub { }, 1, 2)` — anon sub as list element |
| `test_hash_init_with_anon_sub` | Hash initialization `%h = (key => sub { })` with fat comma |
| `test_local_without_parens_as_rhs` | `my $x = local` — bareword `local` as RHS expression |
| `test_local_in_map_block` | `map { local(\$_); }` — `local` inside map block |
| `test_chained_anon_sub_assignments` | Multiple consecutive `my $a = sub { }; my $b = sub { };` |

**All tests passing:** Yes (14 new + existing tests in parser-core error_recovery_tests)

**Bugs found:** None

---
*Green TDD hardener — verified fix handles complex RHS expressions without spurious recovery nodes.*

PR #5061, opened 2026-04-23
fix(parser-core): parse local lvalue before assignment (#0000)

EffortlessSteven:
### Motivation
- Fix a parsing bug where `local` targets could consume the `=` as part of the lvalue parse, producing a `MissingExpression` and splitting anonymous `sub` initializers (e.g. `local $SIG{__WARN__} = sub { ... };`).

### Description
- Change `parse_local_statement` to parse the localized lvalue with ternary-level precedence (`parse_ternary`) instead of `parse_expression`, so assignment operators are not consumed by the lvalue parse (file: `crates/perl-parser-core/src/engine/parser/variables.rs`).

### Testing
- Ran `cargo test -p perl-parser-core --test cpan_error_handling`, `cargo check --all-targets -p perl-parser-core`, `cargo clippy -p perl-parser-core`, and `cargo xtask fmt`, all of which completed successfully (note: `just pr-fast` was not available in this environment).

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/variables.rs
+++ b/crates/perl-parser-core/src/engine/parser/variables.rs
@@ -162,8 +162,10 @@ impl<'a> Parser<'a> {
         let declarator_token = self.consume_token()?; // consume 'local'
         let declarator = declarator_token.text.to_string();
 
-        // Parse the lvalue expression that's being localized
-        let variable = Box::new(self.parse_expression()?);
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing — superseded by #5060 which fixes the root cause in `is_infix_rhs_absent` (helpers.rs) rather than patching variables.rs downstream. #5060's test case covers `local $SIG{__WARN__} = sub { };` directly.

PR #5062, opened 2026-04-23
fix(parser): parse local assignment rhs correctly (#0000)

EffortlessSteven:
### Motivation
- `local $SIG{...} = sub { ... };` was being parsed with a `MissingExpression` because an assignment parse consumed the localized target instead of leaving the RHS to the initializer.
- The change ensures the localized target (LHS) is parsed separately so initializer expressions (including anonymous subs) attach correctly.

### Description
- Parse `local` targets as a non-assignment expression by using `parse_ternary()` instead of `parse_assignment()` when reading the localized target in `variables.rs`.
- Parse the initializer RHS for `local` with `parse_assignment()` after consuming `=` so the RHS is preserved as the declaration initializer.
- Changes made in `crates/perl-parser-core/src/engine/parser/variables.rs` to both the single-variable declaration path and `parse_local_statement`.

### Testing
- Ran `cargo test -p perl-parser-core --test cpan_error_handling` and the test suite passed (`8 passed; 0 failed`).
- Ran `cargo check --all-targets -p perl-parser-core` and it completed successfully.
- Ran formatting and lint checks via `cargo xtask fmt` and `cargo clippy -p perl-parser-core -- -D warnings`, both completed without errors.

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/variables.rs
+++ b/crates/perl-parser-core/src/engine/parser/variables.rs
@@ -87,11 +87,14 @@ impl<'a> Parser<'a> {
             Ok(node)
         } else {
             // Single variable declaration
-            // For 'local', we need to parse lvalue expressions (not just simple variables)
-            // because local can take complex forms like local $ENV{PATH}
+            // For 'local', parse only the localized target expression, not an
+            // assignment expression. The initializer is handled below via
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing — superseded by #5060 (root cause fix in helpers.rs covers this case: `local $SIG{__WARN__} = sub { };`).

PR #5064, opened 2026-04-23
fix(lexer): require base digits in prefixed numerics (#0000)

EffortlessSteven:
### Motivation

- The lexer accepted prefixed numerics like `0x_`, `0b_`, and `0o_` as complete numeric literals even when no valid base digit was present after the prefix, which is incorrect and causes surprising tokenization.

### Description

- Require at least one valid base digit after `0x`/`0b`/`0o` before emitting a prefixed `Number` token by tracking `saw_digit` during digit scanning in `try_number` (file: `crates/perl-lexer/src/lib.rs`).
- When a prefixed literal contains only underscores (no base digits), allow the lexer to fall back to parsing `0` as a decimal `Number` followed by the remaining identifier text.
- Add a regression test `prefixed_number_with_underscores_only_falls_back_to_zero` to `crates/perl-lexer/tests/comprehensive_unit_tests.rs` that asserts `0x_`, `0b_`, and `0o_` produce `Number("0")` then an identifier.

### Testing

- Ran the new regression test with `cargo test -p perl-lexer --test comprehensive_unit_tests prefixed_number_with_underscores_only_falls_back_to_zero`, which passed.
- Ran `cargo check --all-targets -p perl-lexer`, which completed successfully.
- Ran `cargo clippy -p perl-lexer --all-targets -- -D warnings`, which completed successfully.
- The repository `just pr-fast` step could not be run in this environment because `just` is not installed.

------


EffortlessSteven:
## Deep Review: APPROVED

**Logic verified:**
- `saw_digit` flag correctly tracks whether any actual base digit (not just underscore) was seen during scan.
- For `0x_`: `while` loop runs (sees `_`), `saw_digit` stays false, `pos > digit_start` is true but `saw_digit` is false → falls through. `self.position` is NOT updated (still points to `0`). Decimal path processes `0`, stops at `x`. Result: `Number("0")` then `Identifier("x_")`. Matches test expectation.
- For `0x` (no chars): `while` doesn't run, `pos == digit_start` → falls through identically.
- For `0xFF`: digits seen, `saw_digit = true` → normal hex number returned.
- Three cases covered: `0x`, `0b`, `0o` (and their uppercase variants).
- Test covers `0x_`, `0x__`, `0b_`, `0b___`, `0o_`, `0o__` — adequate.
- No banned patterns.

PR #5065, opened 2026-04-23
fix(lexer): treat vertical tab/form feed as whitespace (#0000)

EffortlessSteven:
### Motivation
- The lexer did not treat vertical tab (`\x0B`) and form feed (`\x0C`) as ignorable whitespace, which could cause those bytes to be interpreted as meaningful input and produce incorrect tokens.

### Description
- Extend `skip_whitespace_and_comments` to batch-consume `\t`, `\x0B`, and `\x0C` so vertical tab and form feed are treated like tabs for whitespace skipping.
- Add a focused contract test `lexer_skips_formfeed_and_vertical_tab_whitespace` that verifies tokenization for input containing `\x0C` and `\x0B` between `my`, `$x`, `=`, and `1`.

### Testing
- Ran `cargo test -p perl-lexer lexer_skips_formfeed_and_vertical_tab_whitespace`, which passed.
- Ran `cargo test -p perl-lexer`, which failed due to an unrelated pre-existing test (`wave_c_collapse_hardening`) asserting workspace membership counts; the failure is not caused by this change.
- Ran `cargo check --all-targets -p perl-lexer`, `cargo clippy -p perl-lexer`, and `cargo xtask fmt`, all of which completed successfully in this environment.
- `just pr-fast` was not available in the container so it was not run.

------


EffortlessSteven:
## Deep Review: APPROVED

**Logic verified:**
- `b'\t' | 0x0B | 0x0C` batch skips horizontal tab (0x09), vertical tab (0x0B), and form feed (0x0C) together. All are in Perl's `\s` character class.
- The batch loop terminates on any byte NOT in `{'\t', 0x0B, 0x0C}`, which is correct — spaces are handled by a separate arm, and newlines by the `b'\r' | b'\n'` arm.
- This function is only called between tokens (not inside string/heredoc content), so treating these as whitespace is safe and correct.
- Test `lexer_skips_formfeed_and_vertical_tab_whitespace` verifies `my\x0C$x\x0B=\x0C1;` → `Keyword("my")`, `Identifier("$x")`, `Operator("=")`, `Number("1")`.

**Note:** The PR description mentions `wave_c_collapse_hardening` failing; that pre-existing failure is fixed by #5066.

PR #5077, opened 2026-04-23
feat(tree-sitter-perl-c): parse non-UTF-8 sources (#0000)

EffortlessSteven:
### Motivation

- Problem: `parse_perl_file` used `read_to_string` and required UTF-8, which fails for Perl source files containing non-UTF-8 bytes.
- Goal: allow callers to parse raw file bytes so the C tree-sitter binding can handle files with arbitrary encodings.

### Description

- Added `parse_perl_bytes(&[u8])` which accepts raw bytes and returns a `tree_sitter::Tree`.
- Refactored `parse_perl_code` to delegate to `parse_perl_bytes`, changed `parse_perl_file` to use `std::fs::read`, updated the README API table, and added tests.
- Added unit test `test_parse_bytes` and BDD test `bdd_parse_perl_file_allows_non_utf8_bytes` to cover non-UTF-8 file parsing.

### Testing

- Ran `cargo test -p tree-sitter-perl-c` and all unit and integration tests passed.
- Ran `cargo check --all-targets -p tree-sitter-perl-c` and `cargo clippy -p tree-sitter-perl-c --all-targets -- -D warnings` and both passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b96da1c88333bfbd2c937154f1f7)


EffortlessSteven:
**Deep review — APPROVED with fixes pushed (commit f90a52149).**

**What was fixed:**
- Added BOM documentation note to `parse_perl_bytes`: a UTF-8 BOM (`\xEF\xBB\xBF`) is not stripped automatically and may produce an error node. Callers must strip it if strict grammar compliance is required.
- Added `test_parse_bytes_with_utf8_bom_returns_tree` — verifies the function returns a tree (not an error) when input has a BOM prefix, even if the BOM causes an error node.
- Added `test_parse_bytes_empty_source` — verifies empty `&[u8]` produces a `source_file` tree.

**Logic verified:** `parse_perl_bytes` correctly delegates to tree-sitter's native `parser.parse(&[u8], None)`. `parse_perl_code` delegates to `parse_perl_bytes` via `.as_bytes()`. `parse_perl_file` uses `std::fs::read` (raw bytes) — correctly non-UTF-8 safe.

**Regression risk:** LOW — additive API, no existing callers changed.

PR #5078, opened 2026-04-23
fix(tree-sitter-perl-rs): clamp end_byte to source length (#0000)

EffortlessSteven:
### Motivation
- `Node::end_byte()` could return an offset beyond the current source buffer (e.g. `source.len() + 1`) for some roots, requiring callers to defensively clamp; this change ensures the facade always returns an in-bounds exclusive end offset.

### Description
- Clamp the return value of `Node::end_byte()` to `self.tree_source.len()` in `crates/tree-sitter-perl-rs/src/lib.rs`, tighten the unit test to assert the clamped value, and remove the stale limitation from `crates/tree-sitter-perl-rs/README.md`.

### Testing
- Ran `cargo test -p tree-sitter-perl-rs`, `cargo check --all-targets -p tree-sitter-perl-rs`, `cargo clippy -p tree-sitter-perl-rs --all-targets`, and `cargo xtask fmt`, all of which completed successfully in this environment (note: `just pr-fast` was not available here).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b971482483338d0ab24ef42d205c)


EffortlessSteven:
**Deep review — APPROVED with fixes pushed (commit 0db8b5c5d).**

**What was fixed:**
- Added `test_end_byte_never_exceeds_source_len_for_all_nodes` — walks the root node and all direct children for multiple source patterns (including empty string) and asserts `end_byte() <= source.len()` for each. The original test only checked the root node with one source string; the new test verifies the invariant holds across the full child set and edge-case inputs.

**Logic verified:**
- The clamp `self.inner.location.end.min(self.tree_source.len())` is correct and defensive. Analysis of the v3 lexer shows it sets `token.end = self.position` after consuming, which stays within source bounds. The clamp guards against future parser regressions.
- The off-by-one was documented in README (`source.len() + 1` possible for root); the clamp eliminates this. README limitation note correctly removed.
- `utf8_text` already had the same clamp independently — consistent treatment.

**Regression risk:** LOW — one-line defensive guard, all 61 tests pass.

PR #5082, opened 2026-04-23
fix(perl-regex): ignore code markers in char classes (#0000)

EffortlessSteven:
### Motivation

- Avoid false positives from `RegexValidator::detects_code_execution` when sequences like `(?{` or `(??{` appear as literals inside character classes rather than as executable regex constructs.

### Description

- Update `detects_code_execution` to skip over character class contents (`[...]`) while scanning so markers inside classes are ignored.
- Add regression tests to ensure markers inside char classes are treated as safe and that real code-execution markers following a class are still detected.

### Testing

- Ran `cargo xtask fmt` successfully to format the workspace.
- Ran `cargo test -p perl-regex` and all tests passed.
- Ran `cargo check --all-targets -p perl-regex` and it succeeded.
- Ran `cargo clippy -p perl-regex --all-targets` and it succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9b997f3b483338baeb79ce2ad5ede)


EffortlessSteven:
**deep-review pass — APPROVED with fix-forward**

**Fix pushed:** Added discriminating assertion to `code_execution_after_char_class_is_detected`. The original `assert!(v.detects_code_execution(r"[(?{a-z}](?{ run() })"))` was vacuous: the pre-fix code returned `true` immediately on `[(?{` before the class closed, so the test passed both with and without the fix. The added assertion `assert!(v.detects_code_execution(r"[a-z](?{ run() })"))` uses a clean char class where the real `(?{ })` appears only *after* the class, proving the fix correctly skips the class content.

**Logic: CORRECT.** The char-class skip in `detects_code_execution` handles backslash-escaping inside the class. The `code_execution_markers_inside_char_class_are_safe` negative assertion is the genuinely discriminating test — it would have failed before the fix.

**Merge order note:** PRs #5082 and #5083 both diverge from the same base commit (`70f066b98`). PR #5083 does NOT include this char-class fix. Ops must merge #5082 before #5083.

**Banned patterns:** Clean — no `unwrap()`, `expect()`, or `panic!()` in added code.

Logic: FIXED (strengthened vacuous positive assertion) | Tests: IMPROVED | Regression risk: LOW | Research verification: SKIPPED (no external claims)

PR #5085, opened 2026-04-23
fix(completion): support quoted Moo/Moose has prefixes (#0000)

EffortlessSteven:
### Motivation
- Moo/Moose completions did not reliably trigger when a user typed a quoted `has(...)` option-key prefix (e.g. `('re`) or a quoted `isa` prefix (e.g. `'St`).

### Description
- Relaxed the top-level option-segment character check to allow single and double quote characters so quoted key prefixes are recognized as key-position contexts in `completion.rs`.
- Normalize leading quote characters from the completion `prefix` before filtering in `add_has_type_completions` and `add_has_option_completions` so quoted `isa` and option-key prefixes match candidates.
- Added two regression unit tests: `test_moo_has_option_key_completion_with_quoted_prefix` and `test_moo_isa_type_completion_with_quoted_prefix` in `crates/perl-lsp-rs-core/src/providers/completion/completion.rs`.

### Testing
- Ran `cargo test -p perl-lsp-rs-core test_moo_has_option_key_completion_with_quoted_prefix` which passed.
- Ran `cargo test -p perl-lsp-rs-core test_moo_isa_type_completion_with_quoted_prefix` which passed.
- Ran `cargo check --all-targets -p perl-lsp-rs-core`, `cargo clippy -p perl-lsp-rs-core -- -D warnings`, and `cargo xtask fmt`, all of which completed successfully.
- `just pr-fast` was attempted but is not available in the environment (`just: command not found`).

------

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
@@ -1333,9 +1333,13 @@ impl CompletionProvider {
             return false;
         }
 
-        segment
-            .chars()
-            .all(|ch| ch.is_ascii_alphanumeric() || ch == '_' || ch.is_ascii_whitespace())
+        segment.chars().all(|ch| {
+            ch.is_ascii_alphanumeric()
+                || ch == '_'
+                || ch == '\''
+                || ch == '"'
+                || ch.is_ascii_whitespace()
+        })
     }
 
     /// Check whether the cursor is inside the value position of a Moo/Moose `isa => ...`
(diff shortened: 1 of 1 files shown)


EffortlessSteven (commented):
Deep review pass complete. Logic verified.

**What was checked:**
- `is_has_options_key_context` correctly allows quote chars in the segment validator — the segment chars guard now accepts `'` and `"` so `has 'name' => ('re` triggers option-key completion.
- `add_has_type_completions` and `add_has_option_completions` both strip leading quote chars from the prefix before filtering, so `'St` matches `Str` and `'re` matches `required`.
- `has qw(name) => (re` is unaffected: the options_text segment is `re` (no quotes), same path as before.
- `has ('na'` (no `=>`) does not trigger this path — correct behaviour; Moo/Moose always requires the fat-arrow form.
- Both new regression tests assert the specific completion label, not just non-empty. Tests would fail without the fix.

Security: clean. Performance: clean (hot path is a chars().all() scan over a short segment string).

VERDICT: Logic CORRECT. Tests GOOD. Regression risk LOW.

EffortlessSteven:
label-receipt: deep-reviewed @ c33d71638618f4f91aee234b254949938ec486bd by reviewer-deep

EffortlessSteven:
**CI Receipt: `ci-green`**

All required checks pass on HEAD `c33d7163`. This PR is ready for ops merge queue.

PR #5086, opened 2026-04-23
fix(lints): scope implicit Moose/Moo strict detection to file level (#3489)

EffortlessSteven:
### Motivation

- `use Moo`/`use Moose` inside lexical scopes (for example `eval { ... }`) was being treated as file-wide implicit `strict`/`warnings`, causing PL100/PL101 to be incorrectly suppressed.

### Description

- Treat implicit-strict OO frameworks (`Moo`, `Moose`, `MooseX::StrictConstructor`, etc.) only when they appear as direct top-level `use` statements by adding `collect_file_scope_use_modules` and checking those results before setting `has_strict`/`has_warnings`.
- Preserve misspelled-pragma detection by keeping the full AST `walk_node` pass for `check_misspelled_pragma` while removing scope-unsafe implicit strict toggling from that walk path.
- Replace the prior test that asserted the old "known limitation" with a regression test that asserts eval-scoped `use Moose` does NOT suppress PL100/PL101, and add a test that verifies top-level `use Moo` still suppresses both diagnostics.

### Testing

- Ran `cargo test -p perl-lsp-rs-core --lib implicit_strict_module` and the targeted unit tests passed.
- Ran `cargo check --all-targets -p perl-lsp-rs-core` which succeeded.
- Ran `cargo clippy -p perl-lsp-rs-core --lib -- -D warnings` which succeeded for the changed crate; a separate `clippy` run across all targets reported an unrelated pre-existing lint in another crate (outside this change).

------


EffortlessSteven:
## Deep Review — PR #5086

### Logic: CORRECT

The fix is semantically sound. Detailed analysis:

1. **`collect_file_scope_use_modules` is correctly scoped.** It only iterates `Program { statements }` — direct children. `eval { use Moose; }`, `sub foo { use Moose; }`, and `if ($x) { use Moose; }` are all `NodeKind::*` wrappers, not bare `Use` nodes at the top level, so they are correctly excluded.

2. **Perl semantics question: is `eval { use Moose; }` actually compile-time?** The PR's design decision is CONSISTENT with how `use strict` is already treated: the existing tests (unchanged, on master) assert that `eval { use strict; }` does NOT suppress PL100, because PragmaTracker treats `use strict` in eval as lexically scoped. Using the same rule for OO implicit-strict modules gives a coherent model: only top-level `use Moose` counts as providing file-scope strict. This matches the user expectation — if Moose is inside an eval block, the developer probably didn't intend it as the file's pragma setup.

3. **Old behavior was the bug.** The old `walk_node` pass was scope-unaware and would find `use Moose` anywhere in the AST, incorrectly suppressing PL100 even when it was inside an eval. The PR replaces that with a targeted top-level-only scan.

4. **`misspelled_pragma` detection preserved.** The `walk_node` call remains for misspelled pragma checking — only the `IMPLICIT_STRICT_MODULES` detection was moved out of it and into `collect_file_scope_use_modules`. No misspelled pragma coverage was lost.

### Tests: IMPROVED (fix-forward)

Added 2 commits:
- `test(lints): add sub-body and multi-module edge case tests for implicit strict (#3489)`: adds `implicit_strict_module_inside_sub_body_does_not_suppress_missing_strict` (verifies sub-body `use Moose` is not file-scope) and `all_implicit_strict_modules_suppress_at_top_level` (spot-checks Moose and Modern::Perl top-level — not just Moo — to verify the full `IMPLICIT_STRICT_MODULES` list is covered).
- `style: cargo fmt feature_catalog.rs (pre-existing drift)`: pre-push hook found fmt drift in `feature_catalog.rs` that was already in the PR branch; fixed.

### Regression Risk: LOW

The only behavioral change is: OO implicit-strict modules inside non-top-level scopes no longer suppress PL100/PL101. This is the intended fix. Top-level `use Moo/Moose` still suppresses both diagnostics correctly (covered by existing + new tests).

EffortlessSteven:
## Deep Review Wrapup — PR #5086

**Summary:** reviewer-deep completed correctness review of implicit Moose/Moo strict scoping fix. 2 commits pushed, `deep-reviewed` label set.

### Ending State

**PR improved:** https://github.com/EffortlessMetrics/perl-lsp/pull/5086

**Changes made:**
- `test(lints): add sub-body and multi-module edge case tests for implicit strict (#3489)` — `implicit_strict_module_inside_sub_body_does_not_suppress_missing_strict` (sub-body scoping) + `all_implicit_strict_modules_suppress_at_top_level` (Moose + Modern::Perl top-level — not just Moo)
- `style: cargo fmt feature_catalog.rs (pre-existing drift)` — pre-existing fmt drift in `feature_catalog.rs` on the PR branch fixed by pre-push hook

**Status:** Approved (comment-only — can't approve own PR). `deep-reviewed` set. `needs-deep-review` removed.

**Next step:** ready for `/pr-ready` to exit draft + set `merge-ready`, then green-ci.

### Retrospective

The trickiest part was the Perl semantics question: does `eval { use Moose; }` affect file scope? The answer required understanding that the PR's design is a deliberate linter policy (consistent with PragmaTracker's treatment of `use strict` in eval) rather than a strict Perl runtime semantics question. The existing test suite provided the answer — eval-scoped `use strict` already doesn't suppress PL100, and the PR applies the same rule to OO modules.

The `feature_catalog.rs` fmt drift was a surprise — it was already in the PR branch but the pre-push hook caught it on first push attempt. This is a pattern: Codex-generated PRs sometimes have pre-existing fmt drift in files adjacent to the main change. Worth checking `cargo fmt --check -p <crate>` as a first step when reviewing Codex PRs.

**Breadcrumbs:**
- `green_tdd_wave_g1b_regression_hardening` integration test (`test_perl_lsp_cargo_toml_removed_g1b_imports`) fails on master because it looks for `crates/perl-lsp/Cargo.toml` but the binary was renamed to `perl-lsp-rs`. Pre-existing issue, needs a fix in that test file.
- `collect_file_scope_use_modules` returns `Vec<String>` with clones — if this ever becomes a hot path on large files with hundreds of imports, consider returning `Vec<&str>` or checking inline. Not an issue at current usage.

PR #5098, opened 2026-04-23
fix(perl-semantic-analyzer): guard doc extraction offset bounds (#0000)

EffortlessSteven:
### Motivation
- Prevent a potential panic when `extract_documentation` slices `self.source[..start]` with an out-of-range `start` index.

### Description
- Add an early bounds check (`start > self.source.len()`) to `SemanticAnalyzer::extract_documentation` to return `None` for invalid offsets.
- Add a regression test `test_extract_documentation_with_out_of_bounds_offset` that asserts `extract_documentation` returns `None` for an offset beyond the source length.

### Testing
- Ran `cargo test -p perl-semantic-analyzer` and all tests passed.
- Ran `cargo check --all-targets -p perl-semantic-analyzer` and it succeeded.
- Ran `cargo clippy -p perl-semantic-analyzer -- -D warnings` and there were no warnings.

------


EffortlessSteven:
deep-review: APPROVED

Logic correct. The guard `start > self.source.len()` prevents the OOB slice panic at `&self.source[..start]`. Boundary choice is right: `>` (not `>=`) keeps `start == source.len()` valid (returns entire source as documentation context), only rejecting truly impossible offsets. Test is discriminating — offset `code.len() + 1` is above real source length, verifies `None`. Would have failed before the fix. No banned patterns, O(1). Logic verified, low regression risk.

PR #5099, opened 2026-04-23
refactor(perl-lsp-rs): cache global-variable regex in code actions (#0000)

EffortlessSteven:
### Motivation
- Avoid recompiling the same global-variable assignment regex on every `textDocument/codeAction` request to reduce allocations and CPU work.

### Description
- Hoist the global-variable assignment pattern into a `static GLOBAL_VAR_ASSIGNMENT_RE: LazyLock<Result<regex::Regex, regex::Error>>` and use `if let Ok(re) = &*GLOBAL_VAR_ASSIGNMENT_RE` instead of calling `regex::Regex::new(...).ok()` on every request in `crates/perl-lsp-rs/src/runtime/language/code_actions.rs`.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-rs`, `cargo clippy -p perl-lsp-rs`, and `cargo xtask fmt`, which completed successfully (clippy reported unrelated non-blocking warning). 
- Ran `cargo test -p perl-lsp-rs --lib`, which exercised the test suite but failed on existing sandbox-related tests in this environment (`test_perl_taint_mode_blocks_dangerous_ops`, `test_perl_taint_mode_flag`, `test_safe_executor_disabled`, `test_unsandboxed_execution`); a focused code-action test encountered a linker error when run in this environment.

------


EffortlessSteven:
**Deep review — fix pushed**

Behavior preserved. The regex pattern is correct.

**Issue found and fixed:** `LazyLock<Result<Regex, regex::Error>>` is not the repo standard for static regex statics. Per CLAUDE.md, `static LazyLock<Regex>` initializers may use `unreachable!()` for known-good patterns. Also simplified the call site from `if let Ok(re) = &*RE { if re.is_match(...) }` to a single `if RE.is_match(...)`.

Also applied the pre-existing `clippy::single_match` fix from #5152 (merged) to the branch so the pre-push gate passes on the overall crate.

Commits pushed directly. Ready for `/pr-ready`.

PR #5101, opened 2026-04-23
fix(perl-line-index): validate position column bounds (#0000)

EffortlessSteven:
### Motivation

- `LineIndex::position_to_byte` could accept arbitrarily large `column` values and return offsets past the indexed text, which is surprising and unsafe for callers that expect `None` for out-of-range positions.

### Description

- Add a `text_len: usize` field to `LineIndex` and initialize it in `new` to track the total byte length of the indexed text.
- Update `position_to_byte` to compute the end of the requested line, derive a `max_column`, and return `None` when `column > max_column` instead of producing an out-of-range offset.
- Add and update tests in `line_index_comprehensive.rs` to cover out-of-range columns, newline-boundary behavior, and large-column expectations.

### Testing

- Ran `cargo test -p perl-line-index` and all crate tests passed (`42` tests in `line_index_comprehensive` + `2` roundtrip tests; final run succeeded with no failures). 
- Ran `cargo check --all-targets -p perl-line-index` and `cargo clippy -p perl-line-index`, both completed successfully.
- Ran `cargo xtask fmt` to format changes; `just pr-fast` was not available in this environment so it was not run.

------

### DIFF
--- a/crates/perl-line-index/src/lib.rs
+++ b/crates/perl-line-index/src/lib.rs
@@ -13,6 +13,8 @@
 pub struct LineIndex {
     /// Byte offset of each line start.
     line_starts: Vec<usize>,
+    /// Total byte length of the indexed text.
+    text_len: usize,
 }
 
 impl LineIndex {
@@ -25,7 +27,7 @@ impl LineIndex {
                 line_starts.push(idx + 1);
             }
         }
-        Self { line_starts }
+        Self { line_starts, text_len: text.len() }
     }
 
     /// Convert a byte offset to `(line, column)` using byte columns.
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
deep-review: APPROVED

Correct implementation. Uses `next_start.saturating_sub(1)` as the line boundary so the newline character itself is the last addressable column on its line, while `next_line_start` (which belongs to the next line) correctly returns `None`. Verified against all edge cases: empty text, single-line, multi-line, last-line EOF position (`text.len()` is valid as an end-range column). The two added tests are discriminating and would have failed before the fix. This is the tighter and semantically correct approach compared to PR #5103's original `byte <= line_end_exclusive` (which had an off-by-one). No banned patterns, O(1). Logic verified, low regression risk.

PR #5104, opened 2026-04-23
fix(perl-uri): treat dotfiles as extensionless (#0000)

EffortlessSteven:
### Motivation
- `uri_extension` treated leading-dot filenames like `.gitignore` as having an extension, which conflicts with common filesystem semantics and downstream expectations.

### Description
- Return `None` when the found dot is the leading character (added `if dot_pos == 0 { return None; }` in `uri_extension`).
- Expanded/updated tests to cover dotfiles and multi-dot filenames in `crates/perl-uri/src/classify.rs` and `crates/perl-uri/tests/comprehensive_unit_tests.rs`.
- Change is isolated to `crates/perl-uri` and does not modify other crates or public APIs.

### Testing
- Ran `cargo test -p perl-uri`, which passed (all unit and integration tests succeeded).
- Ran `cargo check --all-targets -p perl-uri`, which succeeded.
- Ran formatting via `cargo xtask fmt`, which completed.
- Ran `cargo clippy -p perl-uri`, which completed without issues.
- `just pr-fast` could not be executed in this environment because `just` is not installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf77dcc083338958f655d3998698)

PR #5107, opened 2026-04-23
fix(perl-uri): support backslash path separators in uri_extension (#0000)

EffortlessSteven:
### Motivation

- `uri_extension` previously only split path segments on `'/'`, so raw Windows-style paths like `C:\Users\dev\script.pl` returned no extension; this change adds Windows path support for consistent behavior across platforms.

### Description

- Change `uri_extension` to split the path on both `'/'` and `'\\'` by using `rsplit(['/', '\\'])` so Windows-style backslashes are handled.
- Add a unit assertion in `crates/perl-uri/src/classify.rs` to validate extraction from a Windows-style path string with backslashes using `uri_extension`.
- Add a regression test `uri_extension_windows_style_path` in `crates/perl-uri/tests/comprehensive_unit_tests.rs` to cover the same case.

### Testing

- Ran `cargo test -p perl-uri` and all tests passed (including new regression tests). 
- Ran `cargo check --all-targets -p perl-uri` and it succeeded. 
- Ran `cargo clippy -p perl-uri` and it reported no issues. 
- Ran `cargo xtask fmt` to format sources successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf78baf88333bdc11b6aa9281d6f)

### DIFF
--- a/crates/perl-uri/src/classify.rs
+++ b/crates/perl-uri/src/classify.rs
@@ -65,6 +65,11 @@ pub fn uri_extension(uri: &str) -> Option<&str> {
         uri.split_once(['?', '#']).map_or(uri, |(path_prefix, _)| path_prefix);
     let path_part = path_without_query_or_fragment.rsplit(['/', '\\']).next()?;
     let dot_pos = path_part.rfind('.')?;
+    // A leading dot means a dotfile (e.g. `.bashrc`, `.gitignore`) — treat as
+    // extensionless rather than returning the entire filename after the dot.
+    if dot_pos == 0 {
+        return None;
+    }
     let ext = &path_part[dot_pos + 1..];
     if ext.is_empty() { None } else { Some(ext) }
 }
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
**Deep review — fix-forward applied**

Found a missing guard: the backslash separator change (`rsplit(['/', '\'])`) did not include the `dot_pos == 0` check from PR #5104. Without it, Windows dotfiles like `C:\Users\.bashrc` would return `Some("bashrc")` instead of `None`.

Also found that `comprehensive_unit_tests.rs` still expected `.gitignore` → `Some("gitignore")` (pre-#5104 behavior), causing a test failure when both changes are present.

Pushed two commits to this branch:
1. Added `dot_pos == 0` guard with comment and two Windows dotfile test cases.
2. Updated `comprehensive_unit_tests.rs` `uri_extension_hidden_file_no_ext` to assert `None`.

All 199 perl-uri tests pass. Labeled `deep-reviewed`.

**Merge order note**: #5104 and this PR both modify `uri_extension`. Merge #5104 first — this branch now contains the combined fix, so if #5107 is merged after #5104 there will be a trivial conflict on that function (same line changed two ways). Recommend merging #5107 instead of #5104 since it's a superset.

PR #5110, opened 2026-04-23
feat(perl-test-generators): broaden module segment coverage (#0000)

EffortlessSteven:
### Motivation

- The module path generator mostly produced PascalCase segments and did not exercise `UPPER_CASE`-style identifiers containing digits or underscores.
- Tests did not explicitly assert the allowed character set for module identifier segments, reducing confidence that generated names match expected Perl module-name shapes.

### Description

- Expanded `module_segment()` to include an `UPPER_CASE`-style branch that produces segments starting with an uppercase letter and followed by uppercase letters, digits, or underscores.
- Added a property test `segments_use_module_identifier_charset` that asserts each segment starts with an ASCII uppercase letter and subsequent characters are alphanumeric or underscore.
- Changes are confined to `crates/perl-test-generators/src/module.rs` and committed as a single focused change.

### Testing

- Ran `cargo test -p perl-test-generators` and the test suite passed.
- Ran `cargo check --all-targets -p perl-test-generators` and it succeeded.
- Ran `cargo clippy -p perl-test-generators` and it completed without failures.

------

### DIFF
--- a/crates/perl-test-generators/src/module.rs
+++ b/crates/perl-test-generators/src/module.rs
@@ -14,6 +14,14 @@ fn module_segment() -> impl Strategy<Value = String> {
         Just("IO".to_string()),
         (
             prop::char::range('A', 'Z'),
+            prop::collection::vec(
+                prop_oneof![prop::char::range('A', 'Z'), prop::char::range('0', '9'), Just('_'),],
+                0..=7_usize,
+            ),
+        )
+            .prop_map(|(first, rest)| std::iter::once(first).chain(rest).collect::<String>()),
+        (
+            prop::char::range('A', 'Z'),
             prop::collection::vec(prop::char::range('a', 'z'), 0..=7_usize),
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

Logic verified:
- New UPPER_CASE branch `(A-Z)(A-Z|0-9|_){0..=7}` is correctly constructed and can never produce an empty segment.
- The `segments_use_module_identifier_charset` property test correctly uses `prop_assert!` (not panic-inducing `assert!`) and constrains charset of the new branch. The test is non-vacuous: the UPPER_CASE branch allows digits and underscores that the existing PascalCase branch (lowercase-only rest) does not; a digit in a segment body would pass the old PascalCase branch only incidentally, but the test correctly catches any implementation that allowed non-alphanumeric/underscore chars.
- No banned patterns (`unwrap`/`expect`/`panic`).

PR #5111, opened 2026-04-23
feat(perl-test-generators): add non-empty unicode strategy (#0000)

EffortlessSteven:
### Motivation
- Property tests that require text content currently need ad-hoc filters or assumptions to exclude empty Unicode strings, which is repetitive and error-prone.

### Description
- Add `non_empty_unicode_string()` in `crates/perl-test-generators/src/unicode.rs` implemented as `unicode_string().prop_filter(...)`, export it from the crate, add a property test that asserts it never yields an empty string, and update the crate `README.md` and example to document the new strategy.

### Testing
- Ran `cargo test -p perl-test-generators`, `cargo check --all-targets -p perl-test-generators`, `cargo clippy -p perl-test-generators`, and `cargo xtask fmt`, and all commands completed successfully.

------

### DIFF
--- a/crates/perl-test-generators/README.md
+++ b/crates/perl-test-generators/README.md
@@ -7,6 +7,7 @@ This crate provides generators for:
 - Perl variables
 - Perl module paths
 - Unicode strings for UTF-8 / UTF-16 roundtrip tests
+- Non-empty Unicode strings for call sites that require content
 
 See [docs/reference/PROPERTY_TESTING.md](../../docs/reference/PROPERTY_TESTING.md)
 for usage patterns and guidance.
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

Logic verified:
- `prop_filter` rejection rate is ~1.9% (empty string probability across the four collection arms with `0..=50`/`0..=100` ranges) — well within proptest's 200-reject default limit, no "too many rejects" risk.
- The test `non_empty_unicode_string_never_empty` is genuinely non-vacuous: `unicode_string()` CAN produce empty strings (minimum collection size is 0 in all arms), so this test WOULD fail if the `prop_filter` were removed.
- `prop_filter` also correctly prevents proptest from shrinking to the empty string during counterexample minimization — the strategy guarantee holds throughout the shrink pass.

PR #5103, opened 2026-04-23
fix(perl-line-index): bound position_to_byte columns (#0000)

EffortlessSteven:
### Motivation

- `LineIndex::position_to_byte` previously accepted `column` values far past the indexed line and produced impossible byte offsets. 
- The change tightens the API so `(line, column)` pairs that extend beyond the line or document are rejected, preserving roundtrip invariants.

### Description

- Add a `text_len: usize` field to `LineIndex` to record the total UTF-8 byte length of the indexed text. 
- Update `position_to_byte` to compute the current line end and return `None` when the requested `column` would extend past the next line start (or past the document end for the final line). 
- Expand the doc comment on `position_to_byte` to document the new `None` behavior for out-of-range inputs. 
- Adjust tests: change the large-column comprehensive test to expect `None` and add an `out_of_bounds_column_returns_none` test in the roundtrip suite.

### Testing

- Ran `cargo test -p perl-line-index`, and all crate tests passed (previous failing case fixed). 
- Ran `cargo check --all-targets -p perl-line-index`, which completed successfully. 
- Ran `cargo clippy -p perl-line-index -- -D warnings`, which completed successfully. 
- Attempted `just pr-fast` but it is not available in this environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf77dcc083338958f655d3998698)

PR #5112, opened 2026-04-23
feat(perl-line-index): add bounded position conversion API (#0000)

EffortlessSteven:
### Motivation
- `LineIndex::position_to_byte` always returns `Some(start + column)` for valid lines, so callers cannot detect when a requested column crosses the line boundary.
- Calls that map LSP positions to byte offsets need a safe way to reject out-of-range columns, including on the last line.

### Description
- Added `text_len: usize` to `LineIndex` to record the source text length for last-line checks.
- Added `LineIndex::position_to_byte_checked(line, column)` which returns `None` when the column exceeds the addressed line's byte boundary.
- Added unit tests in `crates/perl-line-index/tests/line_index_roundtrip.rs` covering accepted and rejected columns for the new API.
- Documented the new API in `crates/perl-line-index/README.md`.

### Testing
- Ran `cargo test -p perl-line-index`, which passed.
- Ran `cargo check --all-targets -p perl-line-index`, which passed.
- Ran `cargo clippy -p perl-line-index`, which passed.
- Ran `cargo xtask fmt` (formatting step) which completed successfully in this environment.
- Attempted `just pr-fast` but the `just` binary is not installed in this environment (`/bin/bash: line 1: just: command not found`).

------

### DIFF
--- a/crates/perl-line-index/README.md
+++ b/crates/perl-line-index/README.md
@@ -7,3 +7,5 @@ A tiny microcrate that provides a byte-oriented line index used by incremental p
 - `LineIndex::new(&str)` builds a line-start cache.
 - `LineIndex::byte_to_position(usize)` maps byte offset to `(line, column)`.
 - `LineIndex::position_to_byte(line, column)` maps back to byte offsets.
+- `LineIndex::position_to_byte_checked(line, column)` returns `None` if the
+  column crosses the addressed line's byte boundary.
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
deep-review: APPROVED with fix

Found and fixed the same off-by-one as PR #5103 in `position_to_byte_checked`. The original `line_end = next_line_start` + `column > line_end.saturating_sub(start)` allowed column = `next_line_start - start`, returning `Some(next_line_start_byte)` which is the first byte of the **next** line. The test was also wrong: `assert_eq!(index.position_to_byte_checked(0, 4), Some(4))` on `"abc\ndef"` accepted 'd' on line 1 as a valid column-4 of line 0.

Fix pushed as commit `65d8ebe7e`: uses `next_start.saturating_sub(1)` so the newline char is the last addressable column. Test corrected to assert `None` for column 4 with an explanatory comment. No banned patterns, O(1). Logic verified, low regression risk.

PR #5117, opened 2026-04-23
feat(perl-test-generators): broaden package-qualified variable generation (#0000)

EffortlessSteven:
### Motivation

- The existing variable generator only produced package-qualified variables with a single package segment, reducing coverage for nested namespace cases.
- Broader package paths better exercise parsers and LSP components that handle multi-segment Perl package names.

### Description

- Added a new `package_path()` proptest strategy that generates 1–4 `identifier()` segments joined by `::`.
- Updated `variable()` to use `package_path()` for package-qualified variables so generated values can be `Foo::Bar::Baz`-style names.
- Expanded the documentation comment for package-qualified examples and added a property test to assert no empty `::` segments are produced.

### Testing

- Ran `cargo test -p perl-test-generators` and all unit and proptest checks passed.
- Ran `cargo check --all-targets -p perl-test-generators` which completed successfully.
- Ran `cargo xtask fmt` and `cargo clippy -p perl-test-generators` which both completed without errors.
- Attempted `just pr-fast` but the `just` tool was not available in this environment (not installed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf7bc5bc8333b79fb2d0c032d6e1)

### DIFF
--- a/crates/perl-test-generators/src/variable.rs
+++ b/crates/perl-test-generators/src/variable.rs
@@ -33,11 +33,16 @@ fn identifier() -> impl Strategy<Value = String> {
     ]
 }
 
+/// Generate a Perl package path (`Foo`, `Foo::Bar`, ...).
+fn package_path() -> impl Strategy<Value = String> {
+    prop::collection::vec(identifier(), 1..=4_usize).prop_map(|segments| segments.join("::"))
+}
+
 /// Generate a full Perl variable name including sigil.
 ///
 /// Covers scalar (`$x`), array (`@x`), hash (`%x`), special variables
 /// (`$_`, `@_`, `$1`–`$9`), and optionally package-qualified names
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

Logic verified:
- `package_path()` generates 1–4 `identifier()` segments joined by `::`. `identifier()` always produces at least one character (either a `Just` literal or a first-char from `IDENT_START`), so no empty `::` segment is possible.
- The resulting `$Foo::Bar::Baz::name` form (sigil + package_path + `::` + name) is valid Perl syntax for package-qualified variables.
- `package_path_has_no_empty_segments` property test correctly asserts the no-empty-segment invariant and is non-vacuous since it depends on the `split("::")` behavior of the `package_path()` strategy output.

PR #5108, opened 2026-04-23
fix(perl-position-tracking): handle CR/CRLF line starts in LineIndex (#0000)

EffortlessSteven:
### Motivation

- `LineIndex::new` only treated `\n` as a line boundary which produced incorrect line/column mapping for CR (`\r`) and CRLF (`\r\n`) documents.
- `offset_to_position` could be given offsets beyond `text.len()` which could yield incorrect columns.

### Description

- Updated `LineIndex::new` to scan the byte buffer and treat `\n`, `\r`, and `\r\n` as line starts so that CR and CRLF are handled correctly.
- Hardened `offset_to_position` by clamping the provided `offset` to `self.text.len()` before computing the line/column.
- Added focused tests in `tests/extended_unit_tests.rs` to cover CRLF, CR-only, and clamping an offset beyond the end of the text.

### Testing

- Ran `cargo test -p perl-position-tracking` and all tests passed.
- Ran `cargo check --all-targets -p perl-position-tracking`, `cargo clippy -p perl-position-tracking --all-targets`, and `cargo xtask fmt` and they passed.
- `just pr-fast` was not available in the environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf78baf88333bdc11b6aa9281d6f)

PR #5119, opened 2026-04-23
fix(perl-position-tracking): clamp invalid byte offsets in LineIndex (#0000)

EffortlessSteven:
### Motivation

- `LineIndex::offset_to_position` could be given offsets past EOF or in the middle of a UTF-8 codepoint, which could produce incorrect columns or be inconsistent.

### Description

- Add `normalize_offset` to clamp requested byte offsets to `text.len()` and walk backwards to the nearest UTF-8 codepoint boundary. 
- Call `normalize_offset` at the start of `offset_to_position` so line/column computation always uses a valid byte boundary.
- Add two regression tests covering a mid-codepoint offset (clamps to previous valid boundary) and an out-of-range offset (clamps to EOF).

### Testing

- Ran `cargo test -p perl-position-tracking` and all tests passed.
- Ran `cargo check --all-targets -p perl-position-tracking` and it passed.
- Ran `cargo clippy -p perl-position-tracking --all-targets -- -D warnings` and it passed.
- Ran `cargo xtask fmt` successfully; `just pr-fast` was not run because `just` is not installed in the environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf86786c8333b13eeb969e10b94a)

### DIFF
--- a/crates/perl-position-tracking/src/line_index.rs
+++ b/crates/perl-position-tracking/src/line_index.rs
@@ -126,6 +126,7 @@ impl LineIndex {
 
     /// Convert byte offset to position (0-based line and UTF-16 column)
     pub fn offset_to_position(&self, offset: usize) -> (u32, u32) {
+        let offset = self.normalize_offset(offset);
         let line = self.line_starts.binary_search(&offset).unwrap_or_else(|i| i.saturating_sub(1));
 
         let line_start = self.line_starts[line];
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
deep-review: APPROVED

`normalize_offset` is correct and well-scoped. The backward walk `while normalized > 0 && !self.text.is_char_boundary(normalized)` terminates in at most 3 iterations (max UTF-8 sequence length is 4 bytes) and safely handles empty text (`min(N, 0) = 0`, loop condition is false). The test for `offset_to_position(2)` on `"a😀b"` correctly expects column 1 (clamps to before the 4-byte emoji). The `usize::MAX` test correctly reaches `text.len()=6` → UTF-16 column 4 (`a`=1 + `😀`=2 surrogate pair + `b`=1).

**Merge-order note**: conflicts with #5108 at the same `offset_to_position` line. #5119's `normalize_offset` is a strict superset of #5108's simple clamp — it covers both OOB and mid-codepoint offsets. Ops must merge #5108 first (for the CRLF `new()` fix), then #5119 — or rebase #5119 on top of #5108. No banned patterns, O(codepoint-size). Logic verified, low regression risk.

PR #5122, opened 2026-04-23
fix(code-actions): avoid EOF overrun in unused-variable quick fix (#0000)

EffortlessSteven:
### Motivation

- The unused-variable quick fix always removed through `line_end + 1`, which could produce an edit past `source.len()` when the declaration is the file's final line without a trailing newline.
- Preventing an out-of-bounds edit avoids crashes or invalid workspace edits when applying the quick fix.

### Description

- Clamp the removal range end to `source.len()` by computing `delete_end` and using it for the removal edit instead of `line_end + 1`.
- Add a focused unit test under `#[cfg(test)]` that reproduces the EOF-without-newline case and asserts the generated removal edit stays within bounds.

### Testing

- Ran `cargo test -p perl-lsp-rs-core fix_unused_variable_removal_does_not_overrun_end_of_file` and the test passed.
- Ran `cargo xtask fmt` which completed successfully, and attempted `cargo clippy` which reported a pre-existing clippy violation in another crate (unrelated to this change).
- `just pr-fast` is unavailable in this environment so the full fast gate was not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf88c2848333b60244f8bc9bd51c)


EffortlessSteven:
**Deep review: APPROVED**

Logic verified — the EOF clamp (`delete_end = if line_end < source.len() { line_end + 1 } else { line_end }`) is correct. When no `\n` is found, `line_end == source.len()` and incrementing would produce an out-of-bounds edit; the clamp prevents that. The complementary `fix_unreachable_code` function already handles this correctly via its `map(|p| ... + 1)` pattern so no other callers are affected.

**Pushed fix** — added two additional test cases to harden the test suite:
- `fix_unused_variable_removal_includes_newline_when_present`: verifies `delete_end = line_end + 1 = source.len()` when source ends with `\n` (normal case, ensures fix does not over-clamp)
- `fix_unused_variable_removal_multiline_removes_correct_line`: verifies `line_start` and `line_end` correctly isolate a non-first, non-last line in a multiline source

All 3 tests pass. Logic correct, regression risk low.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-5122",
  "current_version": {
    "sha": "2057f23004c16f7c183fb6f5b1735c029f58c5b0",
    "updated_at": "2026-04-23T09:53:17Z"
  },
  "label_bindings": [
    {
      "label": "merge-ready",
      "bound_at_version": "2057f23004c16f7c183fb6f5b1735c029f58c5b0",
      "bound_at_timestamp": "2026-04-23T09:53:51Z",
      "bound_by_agent": "pr-ready",
      "valid": true
    }
  ]
}
```

PR #5127, opened 2026-04-23
refactor(perl-symbol): deduplicate symbol kind LSP mapping (#0000)

EffortlessSteven:
### Motivation
- Reduce duplication and drift risk by centralizing the non-variable LSP kind mapping used by both workspace and document-symbol APIs.

### Description
- Introduced a private `const fn base_lsp_kind(self) -> u32` that holds the canonical LSP mapping for non-variable symbol kinds.
- Made `to_lsp_kind()` delegate to `base_lsp_kind()` and made `to_lsp_kind_document_symbol()` handle only variable-specific distinctions then fall back to `base_lsp_kind()`.
- Added a focused test `test_non_variable_document_mapping_matches_workspace_mapping` to ensure non-variable document-symbol mappings remain aligned with workspace mappings.

### Testing
- Ran `cargo test -p perl-symbol` and all tests passed.
- Ran `cargo check --all-targets -p perl-symbol` and it passed.
- Ran `cargo clippy -p perl-symbol --all-targets -- -D warnings` and it passed.
- Ran `cargo xtask fmt` which completed successfully in this environment.
- Attempted `just pr-fast` but the `just` tool is not installed in this environment (command not found).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9bf8b51908333921b694a8937e66c)

### DIFF
--- a/crates/perl-symbol/src/types/mod.rs
+++ b/crates/perl-symbol/src/types/mod.rs
@@ -117,15 +117,8 @@ pub enum SymbolKind {
 }
 
 impl SymbolKind {
-    /// Convert to LSP-compliant symbol kind number (workspace profile).
-    ///
-    /// Maps Perl symbol types to the closest LSP protocol equivalents.
-    /// This is the default mapping used for workspace symbols where a
-    /// generic "Variable" kind is appropriate for all variable types.
-    ///
-    /// See the enum documentation for the full mapping table.
     #[inline]
-    pub const fn to_lsp_kind(self) -> u32 {
+    const fn base_lsp_kind(self) -> u32 {
         match self {
             SymbolKind::Package => 2,      // Module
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review — approved, no changes needed**

Correctness verified: `base_lsp_kind()` (private) is the shared match table. `to_lsp_kind()` delegates to it (workspace profile — same values as before). `to_lsp_kind_document_symbol()` overrides only the three `Variable` variants (Scalar→13, Array→18, Hash→19) and delegates everything else via `_ => self.base_lsp_kind()`. The new test `test_non_variable_document_mapping_matches_workspace_mapping` exhaustively verifies all non-variable kinds produce identical values in both functions. No LSP mapping values changed.

PR #5138, opened 2026-04-23
feat(status): track UX confidence signals — xtask plumbing core (#5138)

EffortlessSteven:
### Motivation
- The README and status docs described “End-to-end UX confidence” qualitatively but the status generator did not surface the manual smoke workflow signal or the known-gap issue burn-down in machine-readable receipts.
- Make those operational signals discoverable so automation and downstream tooling can track UX confidence over time.

### Description
- Add `collect_manual_smoke_workflows` and `collect_open_ux_gap_issue_ids` to `xtask/src/tasks/update_status/quality.rs` to extract Tier C manual editor checks and README known-gap issue IDs.
- Surface those signals in the generated status bullets via `generate_quality_status` and embed a new `confidence_signals` block in `generate_editor_ux_receipt` containing `manual_editor_smoke` and `open_issue_burndown` payloads.
- Add unit tests to lock the behavior: `test_editor_ux_receipt_shape` now asserts the new receipt fields are present and non-empty, and `test_collect_open_ux_gap_issue_ids_is_unique_and_sorted` enforces uniqueness and ordering of extracted issue IDs.

### Testing
- Ran `cargo test -p xtask test_editor_ux_receipt_shape` and it passed.
- Ran `cargo test -p xtask test_collect_open_ux_gap_issue_ids_is_unique_and_sorted` and it passed.
- Ran `cargo check --all-targets -p xtask` and `cargo xtask fmt`, both completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9c539d6788333b2dc9ae826d7a4fa)

PR #5180, opened 2026-04-23
feat(status): track measured editor UX top-line metrics (#0000)

EffortlessSteven:
### Motivation
- Editor UX top-line rows were recorded as planning-only in the status scaffolding even when measured values exist, leaving an important product signal untracked.
- The goal is to surface when editor-UX pass/stability/latency metrics are actually published so the status pages and scorecard reflect measured data instead of only qualitative plans.

### Description
- Add `EditorUxTopLineMetricAvailability` and `collect_editor_ux_metric_availability` to read `.ci/metrics/editor_ux.json` and detect which top-line UX metrics are numeric and therefore measured.
- Use the availability signal in `generate_editor_ux_receipt` to emit `

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9c55bc9e08333b03989930993a894)

### DIFF
--- a/xtask/src/tasks/update_status/quality.rs
+++ b/xtask/src/tasks/update_status/quality.rs
@@ -11,6 +11,13 @@ use color_eyre::eyre::{Context, Result};
 
 use super::{replace_block, run_cmd};
 
+#[derive(Debug, Clone, Copy, Default)]
+struct EditorUxTopLineMetricAvailability {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing — superseded by #5138 (same approach, minimal xtask plumbing). Codex produced ~60 variants of this feature; selecting one per approach-group.

PR #5247, opened 2026-04-23
ci(workflows): add perl version compatibility matrix (#0000)

EffortlessSteven:
### Motivation

- The repository claims parser support for Perl 5.8–5.40 but had no visible CI proving that claim across every minor release.  
- Add a reproducible, reviewable matrix so maintainers and CI consumers can see exactly which Perl minors pass the version-gated probes.

### Description

- Add a new GitHub Actions workflow `.github/workflows/perl-version-matrix.yml` (`Perl Version Matrix`) that enumerates Perl minors from `5.8` through `5.40` and runs version-gated syntax probes for each entry.  
- The workflow includes scheduled runs, `workflow_dispatch`, and label-gated PR execution (`ci:perl-matrix`), and uses `fail-fast: false` so each minor reports independently.  
- Edge lanes for `5.8` and `5.40` also run a lightweight Rust smoke test with the Perl runtime on `PATH` via `cargo test --locked -p perl-lsp-rs runtime::lifecycle::workspace::tests::check_perl_interpreter_does_not_panic --lib`, and the FAQ (`docs/reference/FAQ.md`) is updated to link to the new matrix and clarify the claim.

### Testing

- Ran `cargo xtask fmt --check`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9e8ff3f348333ab51cd8cf8364f73)

PR #5248, opened 2026-04-23
ci(perl): add visible Perl runtime coverage matrix (#0000)

EffortlessSteven:
### Motivation

- The repo claims support for Perl 5.8–5.40 but had no visible CI matrix proving which interpreter versions actually run, leaving the claim unverified.
- Provide an auditable, minimal runtime smoke matrix so reviewers and maintainers can see what Perl interpreters are exercised by CI.

### Description

- Add a new `perl-runtime-matrix` job to `.github/workflows/ci.yml` that provisions Perl versions `5.8`, `5.10`, `5.16`, `5.26`, `5.30`, `5.36`, and `5.40` via `shogo82148/actions-setup-perl@v1`.
- Mark legacy lanes (`5.8` and `5.10`) as experimental and use `continue-on-error` so failures are recorded but do not block merges.
- Run runtime smoke probes including a baseline script and version-gated feature probes, capture the resolved interpreter via `perl -e 'print $^V'`, and append per-row summaries to `GITHUB_STEP_SUMMARY` for auditability.

### Testing

- Validate CI YAML parsing with `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/ci.yml'); puts 'yaml-ok'"`, which succeeded.

------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,72 @@ jobs:
             echo "::notice::Superseded by $LATEST — skipping remaining jobs to save cost"
           fi
 
+
+  # ── Perl Runtime Matrix: prove what actually runs ─────────────────────
+  perl-runtime-matrix:
+    name: Perl Runtime Matrix (${{ matrix.perl }})
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    needs: preflight-latest-check
+    if: needs.preflight-latest-check.outputs.is_latest == 'true'
+    continue-on-error: ${{ matrix.experimental }}
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as duplicate of #5247 which was selected as the winner in the Codex ensemble triage. #5247 is stronger: it covers all 17 minor versions (5.8–5.40) vs. this PR's 7, uses version-gated probes that assert both forward and backward compatibility, ships as a standalone workflow (lower noise in main ci.yml), and adds label-gating so PRs only pay matrix cost when the `ci:perl-matrix` label is applied. #5248's `continue-on-error: experimental` approach is a good idea and has been cherry-picked into the winner.

PR #5249, opened 2026-04-23
ci(perl): add visible 5.8-5.40 runtime matrix (#0000)

EffortlessSteven:
### Motivation
- The repository claims support for Perl 5.8–5.40 but had no dedicated, visible CI lane proving which Perl runtimes actually boot and run; an authoritative signal is needed. 
- Provide a simple, reproducible smoke check so the project and contributors can observe which Perl versions install and execute successfully in CI.

### Description
- Add `.github/workflows/perl-version-matrix.yml` that runs on `pull_request`, `push`, scheduled, and manual dispatch with a matrix for Perl `5.8` through `5.40`, installs each runtime via `shogo82148/actions-setup-perl`, prints the runtime (`perl -v`), and runs a strict/warnings smoke script (`perl -c smoke.pl` / `perl smoke.pl`).
- Add a README badge linking to the new `perl-version-matrix.yml` workflow so the matrix is visible from the project front page.

### Testing
- Ran `cargo xtask fmt --check`, which completed successfully.

------

### DIFF
--- a/.github/workflows/perl-version-matrix.yml
+++ b/.github/workflows/perl-version-matrix.yml
@@ -1,0 +1,69 @@
+name: Perl Version Matrix
+
+on:
+  pull_request:
+    branches: [ main, master ]
+  push:
+    branches: [ main, master ]
+  schedule:
+    - cron: '15 4 * * *'
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  perl-version-matrix:
+    name: Perl ${{ matrix.perl-version }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    strategy:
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as duplicate of #5247 which was selected as the winner in the Codex ensemble triage. #5249 adds the matrix but the syntax smoke is too shallow (only `perl -c` on a basic script with no version-gated feature probes) and it triggers on every push/PR rather than using label-gating for cost control. It also adds a badge to README.md which violates the project's policy of not putting volatile metrics in README. The pinned action SHA (`shogo82148/actions-setup-perl@0f12f8929f8062f1f7795f8a89e9f2f88c5f6ce2 # v1.34.0`) has been applied to the winner.

PR #5250, opened 2026-04-23
ci(perl): add nightly runtime version matrix (#0000)

EffortlessSteven:
### Motivation

- The project claims Perl 5.8–5.40 support but lacked a visible CI runtime matrix proving which Perl versions are actually exercised. 
- Make runtime coverage explicit and discoverable so the documentation claim is tied to a concrete CI job rather than an undocumented assumption.

### Description

- Add a `perl-version-matrix` job to `.github/workflows/ci-nightly.yml` with Perl matrix values `['5.8.9','5.16.3','5.30.3','5.40.0']` that installs Perl via `shogo82148/actions-setup-perl@v1`, builds `perllsp`, and runs a runtime smoke (`cargo test -p perl-subprocess-runtime --test extended_unit_tests` and `target/debug/perllsp --version`).
- Add a Nightly CI badge to `README.md` linking to `ci-nightly.yml` so the matrix is discoverable from the project front page.
- Update `docs/reference/FAQ.md` to explicitly list the Perl versions exercised in CI and reference the `perl-version-matrix` job in `ci-nightly.yml`.

### Testing

- Ran `cargo xtask fmt --check` locally and it passed.

------

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -412,6 +412,49 @@ jobs:
         run: |
           cargo clippy --workspace --locked -- -D warnings -D clippy::all -A missing_docs
 
+  perl-version-matrix:
+    name: Perl Runtime Matrix (${{ matrix.perl_version }})
+    timeout-minutes: 45
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'schedule'
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        perl_version: ['5.8.9', '5.16.3', '5.30.3', '5.40.0']
+    steps:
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as duplicate of #5247 which was selected as the winner in the Codex ensemble triage. #5250 gates on nightly-only which provides runtime visibility but misses the on-demand PR label-gate pattern. It also targets only 4 sparse versions (5.8.9, 5.16.3, 5.30.3, 5.40.0) and references `cargo build -p perllsp --bin perllsp` which is the correct binary but runs the full Rust build in every matrix cell unnecessarily. The Rust smoke approach (running the existing lifecycle test) from PR 5247 is more precise. The nightly-gated idea from #5250 is noted as a future follow-up.

PR #5260, opened 2026-04-23
fix(lifecycle): improve missing-perl user guidance (#0000)

EffortlessSteven:
### Motivation
- The missing-Perl startup path sometimes surfaced low-level diagnostics instead of clear next-steps, hurting first-5-minutes UX for new users.
- Provide a single, actionable message for both "Perl not on PATH" and "configured perl path is broken" cases so users know how to install or fix their Perl configuration.

### Description
- Add a `perl_not_found_message(configured_path: Option<&str>, searched: &[String]) -> String` helper that centralizes the user-facing copy and includes platform-specific install hints (Strawberry Perl, Homebrew, distro package managers).
- Replace the ad-hoc construction in `check_perl_interpreter()` with a call to `perl_not_found_message(...)` when handling `PerlInterpreterResult::NotFound` so the `window/showMessage` text is consistent and actionable.
- Add two focused unit tests that assert the helper message contains the expected phrases for both configured and unconfigured scenarios.
- No change to fallback logging behaviour for `FoundViaFallback`; only user-visible messaging text was centralized.

### Testing
- Ran `cargo test -p perl-lsp-rs --lib runtime::lifecycle::workspace::tests::perl_not_found_message_without_config_is_actionable -- --exact` and it passed.
- Ran `cargo test -p perl-lsp-rs --lib runtime::lifecycle::workspace::tests::perl_not_found_message_with_config_mentions_fix_and_install -- --exact` and it passed.
- Ran `cargo check --all-targets -p perl-lsp-rs` and it completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9e90ad808833394bed5c348605d01)


EffortlessSteven (commented):
deep-review: approved. Message is genuinely actionable — both arms include platform-specific install instructions (strawberryperl.com/brew/distro), the reload-window step, and the settings key. The configured-path arm also adds the install hint for the case where the user has a stale path from a prior install. The extracted perl_not_found_message() function is tested directly. No banned patterns. Logic is correct.

PR #5285, opened 2026-04-23
fix(lsp): avoid parsing Mojolicious templates as Perl (#0000)

EffortlessSteven:
### Motivation

- Opening Mojolicious/Template Toolkit files (e.g. `.ep`, `.tt`, `.tt2`, `.mason`) in editors using an HTML/template language mode caused the server to parse them as Perl, producing noisy diagnostics and unnecessary parse/index churn.

### Description

- Add `TEMPLATE_EXTENSIONS` and helper `is_embedded_template_uri` to detect embedded-template file extensions from URIs or paths. 
- Add `is_perl_language_id` and, in `handle_did_open_with_cancellation`, skip parsing and store a `DocumentState` with `DegradationTier::Minimal` when a template extension is opened with a non-Perl `languageId`, and publish an empty diagnostics notification. 
- Persist the no-parse behavior across subsequent `didChange` updates by detecting the prior `DegradationTier::Minimal` state and short-circuiting parse on change. 
- Add focused unit tests `test_template_file_guard_skips_parse_for_non_perl_language_id` and `test_template_file_guard_persists_across_did_change` covering `didOpen` and `didChange` flows.

### Testing

- Ran `cargo test -p perl-lsp-rs --lib template_file_guard` and the new template-focused tests passed. 
- Ran `cargo clippy -p perl-lsp-rs -- -D warnings` which completed successfully. 

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9e9340fcc8333bff7c50f0f33a6fe)


EffortlessSteven:
reviewer-standards: **CONDITIONAL APPROVAL**

Title requires fix: Change `(#0000)\) to real issue number to pass `validate-title` CI.

Content check: ✓ Clean diff. Unique scope (text_sync.rs template handling).

Next: Fix title, then route to `reviewer-deep`.

EffortlessSteven:
deep-review: APPROVED

Logic verified:

- `is_embedded_template_uri` correctly parses the URI as a URL first, falls back to raw `Path::new`, and uses `eq_ignore_ascii_case` so `.EP`, `.TT2` etc. are caught.
- The `languageId` guard (`!is_perl_language_id`) is correct: Mojolicious `.ep` files opened with Perl languageId should still be parsed; only non-Perl language modes get the guard.
- `DegradationTier::Minimal` is used to mark the no-parse state, and `didChange` checks this tier to persist the no-parse path. The clone of `doc_state.generation` across the change is correct — counter continuity preserved.
- Test covers `didOpen` with `html` languageId, verifies `ast.is_none()` and `DegradationTier::Minimal`. The `didChange` persistence test is also present.
- One concern: `is_embedded_template_uri` called twice in the `didChange` path (once to compute `skip_template_parse` before the mutex lock, and once inside via degradation tier check). The tier check is the actual guard — the URI check is just for early optimization. This is correct but slightly redundant. Not blocking.

PR #5286, opened 2026-04-23
fix(perl-lsp-rs): resolve dashed Mojolicious controllers (#0000)

EffortlessSteven:
### Motivation
- Mojolicious route navigation failed for controller targets that use hyphens (for example `admin-user#list`) because those targets should map to nested controller namespaces (`Admin::User`), which broke go-to-definition when opening Mojolicious projects.

### Description
- Normalize Mojolicious controller route targets by splitting on `-` in addition to `::` and `/` in `normalize_mojolicious_controller_name` so dashed names map to nested namespace segments.
- Add a focused regression test `mojolicious_string_route_with_dashed_controller_definitions_to_nested_controller_method` in `crates/perl-lsp-rs/tests/mojolicious_navigation_tests.rs` that verifies `admin-user#list` resolves to `lib/MyApp/Controller/Admin/User.pm`.
- Update implementation in `crates/perl-lsp-rs/src/runtime/language/navigation.rs` to support the new normalization behavior.

### Testing
- Ran `cargo test -p perl-lsp-rs --test mojolicious_navigation_tests -- --nocapture` and the test binary completed with all tests passing (`5 passed`).
- Ran `cargo check --all-targets -p perl-lsp-rs` and the check completed successfully.
- Ran `cargo clippy -p perl-lsp-rs -- -D warnings` and the lints completed with no warnings.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e9e9340fcc8333bff7c50f0f33a6fe)


EffortlessSteven:
reviewer-standards: **CONDITIONAL APPROVAL**

Title requires fix: Change `(#0000)` to real issue number to pass `validate-title` CI.

Duplicate check: PRs 5288, 5287 are duplicates of this PR (all modify mojolicious_navigation_tests.rs). Close those two; keep this one.

Content check: ✓ Clean diff, test fixture only.

Next: Fix title, then route to `reviewer-deep`.

EffortlessSteven:
deep-review: APPROVED

Logic verified:

- `normalize_mojolicious_controller_name` chains `.flat_map(|part| part.split('-'))` after the existing `::` and `/` splits. Order is correct — hyphen split happens after slash split, producing segments that are then title-cased individually.
- Edge case checked: `admin-user#list` splits to `["admin", "user"]`, title-cased to `["Admin", "User"]`, joined as `Admin::User`. Final package `MyApp::Controller::Admin::User` resolves correctly.
- The test at `crates/perl-lsp-rs/tests/mojolicious_navigation_tests.rs` exercises the real LSP harness, writes actual files, and asserts the URI contains the expected nested path — not vacuous.
- Potential edge case NOT covered: double-hyphen (`admin--user`) produces an empty segment which is filtered by the `segment.is_empty()` guard. Correct behavior.
- Merge-order: no deps on other open PRs.

PR #5361, opened 2026-04-23
fix(perl-lexer): broaden s/tr/y delimiter detection (#0000)

EffortlessSteven:
### Motivation
- The lexer used a hardcoded subset of characters to detect `s`, `tr`, and `y` delimiters which missed valid delimiters and risked misclassifying fat-arrow autoquoting (`=>`).
- Quote-like operator detection should reuse the shared delimiter predicate and respect contextual guards such as hash-subscript braces.

### Description
- Replace the ad-hoc delimiter match with `Self::is_quote_delim(next)` and explicitly detect/preserve fat-arrow autoquoting (`next == '=' && following == Some('>')`) before treating `s`/`tr`/`y` as substitution/transliteration operators. 
- Add a `hash_brace_depth == 0` guard so `s`/`tr`/`y` are not parsed as operators inside hash subscript braces. 
- Add regression tests in `tokenizer_edge_case_tests.rs` for `s:old:new:`, `tr;a-z;A-Z;`, and the fat-arrow autoquote case `s=>1`.

### Testing
- Ran `cargo test -p perl-lexer` (including `lexer_bdd_workflows` and `tokenizer_edge_case_tests`) and all tests passed after the change. 
- Ran `cargo check --all-targets -p perl-lexer` and `cargo clippy -p perl-lexer`, both completed with no errors. 
- `just pr-fast` was not available in the environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea0c21e6b483338777a072b7a85b67)

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -2046,30 +2046,17 @@ impl<'a> PerlLexer<'a> {
             // Check for substitution/transliteration operators
             // Skip if after '->'  -- these are method names, not operators.
             #[allow(clippy::collapsible_if)]
-            if !self.after_arrow && matches!(text, "s" | "tr" | "y") {
+            if !self.after_arrow && self.hash_brace_depth == 0 && matches!(text, "s" | "tr" | "y")
+            {
                 if let Some(next) = self.current_char() {
-                    // Check if followed by a delimiter
-                    if matches!(
-                        next,
-                        '/' | '|'
-                            | '\''
-                            | '{'
-                            | '['
-                            | '('
-                            | '<'
-                            | '!'
-                            | '#'
-                            | '@'
-                            | '$'
-                            | '%'
-                            | '^'
-                            | '&'
-                            | '*'
-                            | '+'
-                            | '='
-                            | '~'
-                            | '`'
-                    ) {
+                    // Check if followed by a valid delimiter while preserving fat-arrow
+                    // autoquoting (`s=>1`, `tr=>2`, `y=>3`).
+                    let following = self
+                        .input
+                        .get(self.position + next.len_utf8()..)
+                        .and_then(|s| s.chars().next());
+                    let is_fat_arrow = next == '=' && following == Some('>');
+                    if Self::is_quote_delim(next) && !is_fat_arrow {
                         match text {
                             "s" => {
                                 return self.parse_substitution(start);
--- a/crates/perl-lexer/tests/tokenizer_edge_case_tests.rs
+++ b/crates/perl-lexer/tests/tokenizer_edge_case_tests.rs
@@ -206,6 +206,22 @@ fn ts_fat_arrow_autoquote_not_substitution() -> Result<(), Box<dyn std::error::E
 }
 
 #[test]
+fn ts_substitution_inside_hash_subscript_not_triggered() -> Result<(), Box<dyn std::error::Error>> {
+    // Inside a hash subscript brace ($h{s}), the bareword `s` must not be
+    // interpreted as the substitution operator.  hash_brace_depth is > 0, so
+    // the s/tr/y detection must be suppressed.
+    let kinds = collect_kinds("$h{s}");
+    // Expected: Scalar variable, opening `{`, bareword key `s`, closing `}`.
+    // There must be NO Substitution token in the stream.
+    assert!(
+        !kinds.contains(&TokenKind::Substitution),
+        "`s` inside {{...}} must remain a bareword, not a substitution operator; got {:?}",
+        kinds
+    );
+    Ok(())
+}
+
+#[test]
 fn ts_qr_angle() -> Result<(), Box<dyn std::error::Error>> {
     let kind = first_kind("qr<pattern>i");
     assert_eq!(kind, TokenKind::Regex, "qr<...> should be Regex");
--- a/crates/perl-lexer/tests/tokenizer_edge_case_tests.rs
+++ b/crates/perl-lexer/tests/tokenizer_edge_case_tests.rs
@@ -171,6 +171,13 @@ fn ts_substitution_with_modifiers() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 #[test]
+fn ts_substitution_colon_delimiter() -> Result<(), Box<dyn std::error::Error>> {
+    let kind = first_kind("s:old:new:");
+    assert_eq!(kind, TokenKind::Substitution, "s:...:...: should be Substitution");
+    Ok(())
+}
+
+#[test]
 fn ts_transliteration_pipe() -> Result<(), Box<dyn std::error::Error>> {
     let kind = first_kind("tr|a-z|A-Z|");
     assert_eq!(kind, TokenKind::Transliteration, "tr|...|...| should be Transliteration");
@@ -178,9 +185,23 @@ fn ts_transliteration_pipe() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 #[test]
+fn ts_transliteration_semicolon_delimiter() -> Result<(), Box<dyn std::error::Error>> {
+    let kind = first_kind("tr;a-z;A-Z;");
+    assert_eq!(kind, TokenKind::Transliteration, "tr;...;...; should be Transliteration");
+    Ok(())
+}
+
+#[test]
 fn ts_transliteration_y_alias() -> Result<(), Box<dyn std::error::Error>> {
     let kind = first_kind("y/a-z/A-Z/");
     assert_eq!(kind, TokenKind::Transliteration, "y/.../.../ should be Transliteration");
+    Ok(())
+}
+
+#[test]
+fn ts_fat_arrow_autoquote_not_substitution() -> Result<(), Box<dyn std::error::Error>> {
+    let kind = first_kind("s=>1");
+    assert_eq!(kind, TokenKind::Identifier, "s=> should remain fat-arrow autoquote identifier");
     Ok(())
 }
 


EffortlessSteven:
**Standards Review:**
✓ No banned patterns
✓ Scope: perl-lexer delimiter detection refactor + fat-arrow guard
✓ Replaces ad-hoc delimiter char match with shared is_quote_delim()
✓ Adds hash_brace_depth guard to prevent s/tr/y parsing in braces
✓ Adds fat-arrow autoquote preservation (s=>1 edge case)
✓ Regression tests for :;; delimiters + => handling
✓ Title format correct

**Route to:** reviewer-deep

EffortlessSteven:
## Deep Review — SEND-BACK

### Summary

PR 5361 is a **regression** relative to master. It should not be merged without correction.

### Problem 1: Simplified substitution detection drops whitespace-tolerant `tr`/`y`

The PR replaces the existing substitution/transliteration detection block with a simpler version that only checks `self.current_char()` (the character immediately after the operator). Master's version handles optional whitespace between the operator and the delimiter (e.g., `tr  |a-z|A-Z|` with a space, or `tr\t|a-z|A-Z|`).

**Master's logic** (preserved, more correct):
- Calls `self.peek_nonspace_and_following()` when the next char is whitespace
- Allows whitespace for `tr`/`y` via `transliteration_allows_whitespace = text == "tr" || text == "y"`  
- Disallows whitespace for `s` via `substitution_disallows_whitespace`

**PR 5361's logic** (regressed):
- Only checks `self.current_char()` — if the next char is whitespace, `is_quote_delim(whitespace)` returns `false` (whitespace is excluded from quote delimiters), so `tr |a-z|A-Z|` would NOT be recognized as transliteration.

### Problem 2: Tests are already green on master

The three new tests added by this PR:
- `ts_substitution_colon_delimiter` — `s:old:new:` already works on master (verified)  
- `ts_transliteration_semicolon_delimiter` — `tr;a-z;A-Z;` already works on master  
- `ts_fat_arrow_autoquote_not_substitution` — `s=>1` fat-arrow guard already exists on master

These tests would pass on both master and the PR branch (the fat-arrow guard was also added in both), so they do not demonstrate a regression fix.

### Stale-base: 61 commits behind master

The PR branch is 61 commits behind master. The substitution/transliteration detection block has been significantly improved in master since this PR was created. The PR's simplified version is not compatible.

### Recommendation

**Close this PR.** The fat-arrow autoquote fix and `is_quote_delim` generalization are already correctly implemented in master with better whitespace handling. The tests can be extracted and submitted separately if they're genuinely not already in master — but as shown above, they test behavior that already works.

EffortlessSteven:
Deep Review: SEND-BACK.

EffortlessSteven:
## Deep Review: SEND-BACK

Reviewer-deep found two correctness regressions relative to current master. The core s/tr/y delimiter fix is already in master in a better form.

### Finding 1 (Critical): consume_balanced_segment_in_string deleted but still needed

Master calls consume_balanced_segment_in_string at seven locations inside double-quoted string interpolation parsing (lines ~2865, 2906, 2913, 2920, 2945, 2961, 2967 of lib.rs). That variant takes a terminator argument (always the closing quote char) and stops scanning at the outer quote character, preventing lexer runaway when a double-quoted string contains malformed nested braces:

    my $x = "${method($arg}";

Without the string-aware variant, consume_balanced_segment will scan past the closing double-quote, producing incorrect tokens for everything that follows. The PR silently downgrades all seven call sites to the non-string-aware version.

### Finding 2 (Medium): after_var_subscript assignments removed

Master sets self.after_var_subscript = matches!(sigil, '$' | '@' | '%') at two code paths (lines ~1651 and ~1729) where bare sigil tokens are emitted for dereference contexts (${...}, @{...}, %{...}). The PR removes both. Without these, the { handler cannot distinguish a dereference opener from a block opener, corrupting hash_brace_depth for subsequent s/tr/y detection.

### Finding 3: Approach inferior to master

Master already has the fat-arrow guard (is_fat_arrow) AND handles whitespace between the operator and delimiter (s /old/new/ form) via peek_nonspace_and_following(). The PR only handles the no-whitespace case. Merging this PR would regress the whitespace handling that master already has.

### Recommendation

Close this PR. Master already has a correct and more complete implementation of the core fix (hash_brace_depth guard + fat-arrow guard + whitespace-delimiter handling). The two regressions (consume_balanced_segment_in_string removal, after_var_subscript removals) would require reverting substantial changes.

EffortlessSteven:

EffortlessSteven:
### Review: broaden s/tr/y delimiter detection (and fix two false positives)

**Three distinct fixes bundled into one diff** -- the title undersells the change:

1. **Genuine broadening:** the explicit 19-char allowlist `'/' | '|' | '\'' | '{' | '[' | '(' | '<' | '!' | '#' | '@' | '$' | '%' | '^' | '&' | '*' | '+' | '=' | '~' | '`'` is replaced by `Self::is_quote_delim(next)`, which is the existing "any non-alphanumeric, non-whitespace character" helper:

   ```rust
   fn is_quote_delim(c: char) -> bool {
       // Perl allows any non-alphanumeric, non-whitespace character as delimiter,
       // including control characters (e.g. s\x07pattern\x07replacement\x07).
       !c.is_ascii_alphanumeric() && !c.is_whitespace()
   }
   ```

   This is the **actually-correct** Perl semantics. The previous explicit list was a "we'll handle the 19 most common delimiters" approximation that silently mistokenized `s:old:new:`, `s;a;b;`, `s.a.b.`, `tr,a,b,`, etc. The new tests `ts_substitution_colon_delimiter` and `ts_transliteration_semicolon_delimiter` pin two of the previously-broken cases.

2. **Narrowing fix #1: fat-arrow autoquote suppression.** `s=>1`, `tr=>2`, `y=>3` should be treated as bareword keys followed by fat-arrow, not as the substitution operator with `=` as the delimiter. The fix:

   ```rust
   let following = self.input.get(self.position + next.len_utf8()..)
       .and_then(|s| s.chars().next());
   let is_fat_arrow = next == '=' && following == Some('>');
   if Self::is_quote_delim(next) && !is_fat_arrow {
       // ... s/tr/y operator path
   }
   ```

   `next` is the candidate delimiter character (just past the `s`/`tr`/`y` bareword); `following` looks one more char ahead. If we see `=>` we suppress the operator path entirely. The new test `ts_fat_arrow_autoquote_not_substitution` pins this.

3. **Narrowing fix #2: hash subscript brace suppression.** Inside `$h{s}`, the bareword `s` should not be interpreted as the substitution operator. The fix gates the entire detection on `self.hash_brace_depth == 0`:

   ```rust
   if !self.after_arrow && self.hash_brace_depth == 0 && matches!(text, "s" | "tr" | "y") {
   ```

   This relies on `hash_brace_depth` being correctly tracked elsewhere in the lexer. The test `ts_substitution_inside_hash_subscript_not_triggered` pins this. Subtle and important: without this, `$h{s}` would suddenly start lexing as `$h`, `{`, `s/old/new/...` and never recover.

**Why the `is_quote_delim` swap is correctness, not just style**:

Perl's `s`/`tr`/`y` operators accept *any* non-alphanumeric, non-whitespace character as the delimiter, including control characters and Unicode punctuation. The previous allowlist excluded `:`, `;`, `,`, `.`, `?`, `"`, and any character beyond ASCII. Real-world Perl code uses these (`s,old,new,`, `s.a.b.`) when the natural choice (e.g. `/`) is itself in the pattern. Closing this gap means the lexer no longer tokenizes those forms as bareword + assignment + ... + slash mess.

**Concerns**:

1. **`is_fat_arrow` only checks `=` followed by `>`.** What about `=` followed by spaces then `>`? Perl's fat-arrow tokenizer accepts `s = > 1` as `s`, `=`, `>`, `1` (which is wrong code but doesn't trigger autoquote). The current code's `following == Some('>')` is correct -- if there's whitespace, it's *not* a fat arrow. Worth a one-line comment noting this is intentional.

2. **The `position + next.len_utf8()` arithmetic** assumes `next` is the *first* char at the current cursor. If `current_char()` returns the char *at* `self.position` (rather than just-past), the offset arithmetic is consistent with the rest of the lexer. Quick double-check that `next.len_utf8()` matches the lexer's "advance one character" convention -- if `current_char()` returns `Some(c)` where `c` starts at `self.position`, then `self.position + c.len_utf8()` is the *next* char's position, and `following` correctly looks one further. Looks right.

3. **Hash-subscript suppression is *too aggressive* for nested cases**:

   ```perl
   $h{ do { s/old/new/; "key" } }
   ```

   Here, `do { s/old/new/ ... }` is inside a hash subscript brace, so `hash_brace_depth > 0` -- but the `s/old/new/` should still be a substitution, since it's nested in a `do {}` block. The fix as written would suppress it. Two paths forward:
   - **Test it:** add a `ts_substitution_inside_do_block_inside_hash` test. If it currently fails, the `hash_brace_depth == 0` gate is overly aggressive.
   - **Refine:** track the *immediate-enclosing* brace context, not just depth. A `do {...}` inside a hash subscript should reset the "inside hash subscript" predicate.
   
   Probably out of scope for this PR if the simple form fixes the most common case; worth filing as a follow-up.

4. **The bundled-fixes pattern** technically violates `AGENTS.md`'s "one concern per PR" rule (this PR is three concerns: broaden delimiters, fix fat-arrow, fix hash-subscript). For a lexer module where all three flow from "what counts as the start of an s/tr/y operator?", I think it's defensible -- the three checks all live on the same conditional and are testing the same predicate. But a future PR review tool might flag this. Worth noting the bundle in the PR body.

5. **The `is_quote_delim` helper is a `Self::` associated function** -- not a method. Good (no `&self` borrow). Inlinable.

1. Add a test for opening-bracket-form delimiters: `s{old}{new}`, `tr[a-z][A-Z]`, `s(old)(new)`. The original code listed `{`, `[`, `(`, `<` explicitly; the new code accepts all of them via `is_quote_delim`. Pin the no-regression with explicit tests.

2. Add a Unicode-delimiter test:
   ```rust
   let kind = first_kind("s«old»«new»"); // U+00AB, U+00BB
   ```
   The new `is_quote_delim` accepts these (they're not ASCII-alphanumeric, not whitespace), but I'd want a test pinning that contract.

3. Add the nested-`do {}` test to confirm whether fix #3 over-suppresses (see concern #3).

**Verdict**: real correctness fixes for three tokenization bugs. The "broaden" framing in the title undersells the impact -- the hash-subscript and fat-arrow fixes are silent-misparse killers that have probably been causing weird LSP behavior on perfectly-valid Perl. Tests pin all three. Worth checking whether the hash-subscript suppression is too aggressive for nested `do {}` cases.

EffortlessSteven:
diff-audit: clean — 2 files / +47/-23 lines, scope matches title (Self::is_quote_delim swap, fat-arrow autoquote guard, hash_brace_depth==0 guard, 3 regression tests in tokenizer_edge_case_tests.rs). Coherent diff matches PR body. No artifacts, no agent-trail dirs, no banned patterns.

EffortlessSteven:
Label reconciliation per 2026-04-26 directive (PR #6808): sign-off IS a routing decision, never coexists with needs-*. Stripped post-builder sign-offs; pipeline will re-evaluate gates after needs-builder-fix is addressed.

EffortlessSteven:
[builder] Superseded by #7001. Cherry-picked regression test commits to fresh-master branch. Fix logic was already in master's more complete implementation.

PR #5368, opened 2026-04-23
refactor(perl-parser): optimize tied-handle heredoc detection (#0000)

EffortlessSteven:
### Motivation

- Reduce unnecessary work and allocations: the previous detector rescanned the source and compiled a dynamic regex per tied handle.
- Avoid false positives by only emitting diagnostics when the printed handle is actually tied.

### Description

- Replace per-handle rescans with a normalized `HashSet` of tied handles collected from `TIE_PATTERN` and a single static `PRINT_HEREDOC_PATTERN` to find all `print <handle> <<` usages.  
- Emit `TiedHandleHeredoc` diagnostics only when the printed handle is present in the normalized tied-handle set.  
- Add a regression test `test_tied_handle_does_not_report_other_handles` to ensure non-tied handles are not reported.  
- Minor refactor to `crates/perl-parser/src/heredoc_anti_patterns.rs` implementing the above improvements.

### Testing

- Ran `cargo test -p perl-parser tied_handle` and the tied-handle tests passed.  
- Ran `cargo check --all-targets -p perl-parser` and it passed.  
- Ran `cargo clippy -p perl-parser --all-targets -- -D warnings` and it passed.  
- `just pr-fast` was not run because `just` is not installed in the environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea0c2548688333a89b62ef0c07e0ff)


EffortlessSteven:
**Standards Review:**
✓ No banned patterns (get(0) here is on regex capture, not slice — idiomatic)
✓ Scope: perl-parser heredoc detection optimization
✓ Tests validate non-tied handles excluded (regression test added)
✓ Title format correct

**Route to:** reviewer-deep

EffortlessSteven:
## Deep Review: SEND-BACK

Reviewer-deep found a correctness regression in location_from_start and the removal of three tests that would have caught it.

### Finding 1 (Bug): location_from_start returns wrong line number for mid-line matches

The PR replaces location_from_start with a simpler implementation:

    fn location_from_start(code: &str, offset: usize, start: usize) -> Location {
        let line = code[..start].lines().count();
        ...
    }

This is off by one for any match that is NOT at the start of the file. Rust's str::lines() returns an iterator over logical lines. For code[..start] where start points into the middle of line 1, lines().count() returns 2 (the prefix contains two logical lines), but the correct 0-based line index is 1.

Concrete example from the removed test test_location_mid_line_column_nonzero:
- code = '# comment\n    use Filter::Simple;\n'
- start = 14 (position of 'use' on line 1)
- code[..14] = '# comment\n    ' -> lines() = ['# comment', '    '] -> count() = 2
- PR reports: line=2, column=4
- Master correctly reports: line=1, column=4 (via partition_point binary search)

The fix is either:
- Use code[..start].lines().count().saturating_sub(1) — but this then fails for start=0 (returns 0 correctly, but is fragile)
- Count newlines directly: code[..start].bytes().filter(|&b| b == b'\n').count()
- Keep master's partition_point approach

### Finding 2: Three regression tests removed

The PR deletes test_location_first_byte_is_line_zero_column_zero, test_location_third_line_accurate, and test_location_mid_line_column_nonzero. These tests were added specifically to validate the location calculation. Removing them means the off-by-one is invisible until an LSP client reports wrong diagnostic positions.

### Finding 3: PRINT_HEREDOC_PATTERN regex correctness

The new static regex r'print\s+([*$]?\w+)\s+<<' correctly matches 'print FH <<' and 'print $fh <<'. However, it does not match 'printf HANDLE <<' or handle method-call forms like 'print $self->{handle} <<'. These are pre-existing limitations (the old code had them too), so not a regression, but worth noting.

### Finding 4: The TiedHandleDetector refactor is correct

The switch from per-tied-handle per-regex-compilation to a single PRINT_HEREDOC_PATTERN + HashSet lookup is correct and avoids O(n) Regex compilations. The normalization (stripping '*' prefix) is applied correctly to both the tie-side and print-side handles. The new test test_tied_handle_does_not_report_other_handles is valid and catches the key cross-handle false-positive.

### Action Required

Fix location_from_start: replace lines().count() with bytes().filter(|&b| b == b'\n').count() and restore the three location accuracy tests (or equivalent). The PatternDetector trait signature simplification (removing line_starts parameter) is fine to keep.

EffortlessSteven:

EffortlessSteven:
Label reconciliation per 2026-04-26 directive (PR #6808): sign-off IS a routing decision, never coexists with needs-*. Stripped post-builder sign-offs; pipeline will re-evaluate gates after needs-builder-fix is addressed.

EffortlessSteven:
## Builder Fix: PR 5368 — Tied-handle heredoc optimization

### What was wrong

The PR had fresh-root divergence (many already-merged commits stacked on the old master root). The deep reviewer's findings:

1. **location_from_start off-by-one**: Not applicable to the current cherry-pick — on master, `location_from_start` already uses `partition_point` (correct). The PR's old version calling `lines().count()` was from a pre-master-rebuild snapshot.

2. **Three removed tests**: Not applicable — `test_location_first_byte_is_line_zero_column_zero`, `test_location_third_line_accurate`, and `test_location_mid_line_column_nonzero` all exist on master and are preserved in this fix.

3. **Old API mismatch**: The PR called `location_from_start(code, offset, start)` but master's signature is `location_from_start(line_starts: &[usize], offset, start)`. The cherry-pick adapts to use the current API.

### What was fixed

Cherry-picked only the optimization logic onto fresh master, adapted to the current API:

- `TiedHandleDetector::detect` now collects tied handles into a `HashSet<String>` (normalized, `*` prefix stripped)
- Added `PRINT_HEREDOC_PATTERN` static `LazyLock<Regex>` compiled once at startup
- Single regex scan + HashSet lookup replaces O(n) per-handle `Regex::new()` calls
- Uses current `location_from_start(line_starts, offset, start)` — partition_point-based, correct
- Added `test_tied_handle_does_not_report_other_handles` regression test

### Verification

- `cargo check -p perl-parser`: clean
- `cargo test -p perl-parser -- heredoc_anti_patterns`: 18 passed
- `cargo test -p perl-parser -- test_location`: 4 passed (all 3 location accuracy tests preserved)
- `cargo clippy -p perl-parser`: no issues
- `cargo xtask fmt --check`: clean

### Current state

Branch rebased onto master with 1 clean commit (`c28830cb18`). Pushed force-with-lease to `codex/improve-perl-parser-ogpzkt`.

EffortlessSteven:
Closing as STALE-BASE-DEEP — 7+ days behind master after today's 57-merge session. If work is still relevant, re-implement as fresh PR against current master (HEAD: 04801de8). Bulk close 2026-04-30.

EffortlessSteven:

PR #5370, opened 2026-04-23
fix(perl-dap): quote all whitespace-delimited args (#0000)

EffortlessSteven:
### Motivation

- Shells can split arguments on any whitespace, but `format_command_args` only checked for the literal space character which allowed tab/newline-delimited args to be mis-handled.

### Description

- Change detection from `arg.contains(' ')` to `arg.chars().any(char::is_whitespace)` in `format_command_args` so any Unicode whitespace triggers quoting.
- Preserve existing platform-specific quoting behavior for Windows and non-Windows paths.
- Add unit tests for tab and newline arguments to prevent regressions (`crates/perl-dap/src/command_args/mod.rs`).

### Testing

- Ran `cargo test -p perl-dap command_args` and the command-args unit tests passed.
- Ran `cargo clippy -p perl-dap` and it passed, while `cargo clippy -p perl-dap --all-targets -- -D warnings` fails on unrelated pre-existing `expect()` usage in tests (not caused by this change).

------

### DIFF
--- a/crates/perl-dap/src/command_args/mod.rs
+++ b/crates/perl-dap/src/command_args/mod.rs
@@ -5,7 +5,7 @@
 pub fn format_command_args(args: &[String]) -> Vec<String> {
     args.iter()
         .map(|arg| {
-            if arg.contains(' ') {
+            if arg.chars().any(char::is_whitespace) {
                 #[cfg(windows)]
                 {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Duplicate of #5371 - #5371 is the superset (preserves empty AND whitespace args vs whitespace only). Closing in favor of #5371.

PR #5371, opened 2026-04-23
fix(perl-dap): preserve empty and whitespace args (#0000)

EffortlessSteven:
### Motivation

- Shell invocation code only quoted args containing a literal space, which allowed empty strings and other whitespace-bearing args (e.g. newline) to be lost or split by shells.

### Description

- Change `format_command_args` to quote empty arguments and any argument containing Unicode whitespace by using `arg.is_empty() || arg.chars().any(char::is_whitespace)` in `crates/perl-dap/src/command_args/mod.rs`.
- Update unit tests in `crates/perl-dap/src/command_args/mod.rs` to assert empty-arg preservation via platform-appropriate quoting.
- Update integration tests in `crates/perl-dap/tests/command_args_integration_tests.rs` and `crates/perl-dap/tests/platform_comprehensive_unit_tests.rs` to expect quoted empty and newline-containing args.

### Testing

- Ran `cargo test -p perl-dap --test command_args_integration_tests` and the integration suite passed (`27 passed`).
- Ran `cargo check --all-targets -p perl-dap` and it succeeded.
- Ran `cargo clippy -p perl-dap` and it succeeded, while `cargo clippy -p perl-dap --all-targets -- -D warnings` failed due to pre-existing test-code lint violations unrelated to this change.

------


EffortlessSteven:
**Standards Review:**
✓ No banned patterns
✓ Scope: perl-dap command args handling
✓ Tests updated for empty + whitespace arg preservation
✓ Superset of #5370 (which was closed as dupe)
✓ Title format correct

**Route to:** reviewer-deep

EffortlessSteven:
## Deep Review — APPROVED (fix pushed)

**Reviewer:** reviewer-deep

### What I verified

**Logic correctness of `arg.is_empty() || arg.chars().any(char::is_whitespace)`:**
- `char::is_whitespace` in Rust covers U+0009 (HT/tab), U+000A (LF), U+000B (VT), U+000C (FF), U+000D (CR), U+0020 (SP), U+0085 (NEL), U+00A0 (NBSP), and further Unicode whitespace — all shell-significant
- Empty arg `""` now correctly gets quoted to `""` (Windows) / `''` (Unix) so it is passed as a distinct empty-string argument rather than being lost
- The `arg.is_empty()` check comes first, short-circuiting before `chars().any(...)` — no panic on empty string

**Caller context:** `format_command_args` is a public API utility re-exported from `api.rs`, `platform/mod.rs`, and `shell/mod.rs`. The actual launch-time process spawning in `debug_adapter/process.rs` uses `cmd.args(&raw_args)` directly without calling `format_command_args`. This function is for external callers assembling shell command strings, not for our own `Command::args()` invocations. The quoting behaviour is therefore appropriate for its intended use.

**Test coverage gaps — test pushed:** Added `\r` (carriage return) quoting test to `command_args_integration_tests.rs`. CR is in `char::is_whitespace` and relevant for CRLF-encoded args on Windows-origin input, but was not tested alongside the tab and newline cases.

**Pre-existing gap (not introduced by this PR):** An argument containing only a single-quote without any whitespace (e.g. `"it's"`) still passes through unquoted on Unix. This is a pre-existing shell quoting gap outside this PR's scope — the original PR only added empty/whitespace coverage.

**Windows path:** `format!("\"{}\"", arg.replace('"', "\\\""))` correctly handles empty (`""`) and tab/newline-containing args on Windows. No issues.

### Verdict

APPROVED. `deep-reviewed` label applied.

EffortlessSteven:
### Review: preserve empty and whitespace args

**Two real shell-splitting correctness fixes** in `format_command_args`:

1. **Empty string args** were silently passing through unquoted. The shell would then collapse them away when the command runs:

   ```
   args = ["perl", "", "-e", "print 'x'"]
   passed to shell: perl  -e print 'x'    # the empty arg is gone!
   ```

   The fix quotes them: `""` (Windows) / `''` (POSIX), preserving the positional argument.

2. **Whitespace args other than space** (tab, newline, carriage return) were passing through unquoted. Same problem: the shell tokenizer splits on tab/newline/CR, so `"line1\nline2"` would become two args.

The fix is one expression change:

```rust
- if arg.contains(' ') {
+ if arg.is_empty() || arg.chars().any(char::is_whitespace) {
```

That's the **canonical** check for "needs shell quoting." `char::is_whitespace` covers the full Unicode whitespace set, which is more aggressive than POSIX shell strictly requires but is the right conservative posture.

**Tests are exhaustive for the new contract**:

| Test | Input | Output |
|---|---|---|
| `empty_arg_is_preserved_via_quotes` | `""` | `'""'` (Win) / `"''"` (POSIX) |
| `newline_without_space_is_quoted` | `"line1\nline2"` | `"'line1\nline2'"` |
| `tab_without_space_is_quoted` | `"col1\tcol2"` | `"'col1\tcol2'"` |
| `carriage_return_without_space_is_quoted` | `"win\rarg"` | `"'win\rarg'"` |

**Important: tests rename the previously-incorrect contract** (`single_empty_string_arg_passes_through` -> `single_empty_string_arg_is_quoted_to_preserve_emptiness`, `newline_without_space_passes_through` -> `newline_without_space_is_quoted`). The old test names *encoded the bug* in their names. Renaming them while flipping the assertion is exactly the right move -- a future "this test name says it passes through, why does it fail?" debug session would have been confusing.

**Concerns**:

1. **The Unix POSIX quoting path** uses single quotes:
   ```rust
   #[cfg(not(windows))] {
       format!("'{}'", arg.replace('\'', "'\\''"))
   }
   ```
   (Inferred from test expectations -- not visible in diff.) That handles single-quote-in-arg correctly via the `'\''` close-and-reopen pattern. Good.

2. **The Windows path** uses double quotes:
   ```rust
   format!("\"{}\"", arg.replace('"', "\\\""))
   ```
   This handles `"`-in-arg, but **doesn't handle `\`** (backslash). On Windows, `cmd.exe` doesn't use `\` as an escape; PowerShell does. The standard CommandLineToArgvW algorithm requires:
   - A `\\` before a `"` becomes `\` (one backslash kept, `"` is escape).
   - `\` before any other char is preserved as-is.
   - `\\\\` becomes `\\` if followed by `"`; otherwise stays as `\\\\`.
   
   The current implementation works for **most** Windows args (no backslashes-before-quotes), but breaks for paths like `"C:\foo\\\""` (rare in practice, but a known shell-injection vector). Worth either:
   - Adding a Windows-specific test for `arg = "C:\\foo\\\""` and verifying round-trip
   - Or, importing the canonical `windows_command_line_quoting` algorithm (small, well-documented)
   
   Not blocking for *this* PR (which fixes a different bug), but worth a follow-up issue.

3. **`char::is_whitespace` includes Unicode chars** like U+00A0 (non-breaking space), U+2028 (line separator), etc. Most shells treat only ASCII whitespace as token separators, so quoting Unicode-whitespace-only args is harmlessly conservative -- it won't break correctness, just produces verbose output for some pathological inputs. Worth a one-line comment justifying the choice:
   ```rust
   // Quote conservatively on any Unicode whitespace, even though POSIX
   // shells only split on $IFS (typically space/tab/newline). Quoting
   // is harmless; failing to quote splits args.
   ```

4. **The test `format_command_args_single_empty_string`** in `platform_comprehensive_unit_tests.rs` updated its assertion message from `"empty string has no space, returned as-is"` to `"empty string should be quoted to survive shell splitting"`. Reading the old assertion message as "this is the contract we're documenting" makes it clear the fix is changing the contract, not just the implementation.

**Verdict**: real correctness fix for shell argument forwarding. The empty-arg case is a known-bug-in-many-CLIs pattern (Python's `subprocess`, Go's `os/exec`, etc. all hit this). Tests are exhaustive for the new contract. Worth a follow-up for Windows backslash-before-quote handling and a comment justifying `char::is_whitespace`.

EffortlessSteven:
diff-audit: clean — 3 files / +42/-8 lines, scope matches title (`format_command_args` quotes empty/whitespace args; tests updated to expect quoted empty/newline args). Coherent diff matches PR body. No artifacts, no agent-trail dirs, no banned patterns.

EffortlessSteven:
## green-ci: needs-ci-fix

- PR Smoke (Fast Feedback)

PR #5375, opened 2026-04-23
feat(perl-regex): improve regex modifier hover details (#0000)

EffortlessSteven:
### Motivation

- Hover text for Perl regexes only covered a small set of modifiers and silently ignored unknown flags, reducing usefulness for IDE users.
- The goal is to make hover information more informative by explaining more Perl-specific modifiers, deduplicating repeated flags, and surfacing unsupported flags.

### Description

- Added a `describe_modifier` helper that maps a broader set of Perl regex modifier flags (`a`, `d`, `l`, `u`, `n`, `p`, `r`, `c`, `o`, `e` in addition to existing ones) to human-friendly descriptions.
- Updated `RegexAnalyzer::hover_text_for_regex` to deduplicate repeated modifiers, collect unknown flags, and append an `Unknown modifiers:` line when present.
- Added behavior tests `scenario_hover_text_lists_extended_perl_modifiers` and `scenario_hover_text_deduplicates_and_reports_unknown_modifiers` to verify extended explanations, deduplication, and unknown-flag reporting.

### Testing

- Ran `cargo test -p perl-regex` and all crate tests (unit, behavior, named-capture, and doctests) passed. 
- Ran `cargo check --all-targets -p perl-regex` which completed successfully. 
- Ran formatting and lint gates with `cargo xtask fmt` and `cargo clippy -p perl-regex --all-targets -- -D warnings`, both of which completed without errors. 
- `just pr-fast` was not available in this environment (`just: command not found`) so the full local pre-push helper was not executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea0c31d6c88333a7eb4e025c582167)


EffortlessSteven:
## Research Verification

**Summary:** 10 of 10 Perl regex modifier claims verified

### Perl Claims

| Claim | Status | Finding | Source |
|-------|--------|---------|--------|
| `/a` means "ASCII-safe character classes" | VERIFIED | Perl 5.14+ uses `/a` to restrict `\d`, `\s`, `\w`, and POSIX classes to ASCII range | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/d` means "native platform character set semantics" | VERIFIED | `/d` applies platform-dependent character-set rules (pre-5.14 default) | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/l` means "locale-dependent character semantics" | VERIFIED | `/l` applies locale-based rules for character matching | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/u` means "Unicode character semantics" | VERIFIED | `/u` applies full Unicode character-set rules | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/n` prevents capturing by default in unnamed groups | VERIFIED | `/n` (Perl 5.22+) prevents `()` from creating numbered capture variables | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/p` preserves `${^PREMATCH}`, `${^MATCH}`, `${^POSTMATCH}` | VERIFIED | `/p` is documented; in Perl 5.20+ this is automatic but `/p` remains recognized | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/r` performs non-destructive substitution | VERIFIED | `/r` returns modified string without altering original | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/c` retains current position during `/g` scans | VERIFIED | `/c` maintains position for repeated matching | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/o` compiles pattern only once (deprecated) | VERIFIED | `/o` is documented as optimization flag (now discouraged) | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |
| `/e` evaluates replacement as code in substitutions | VERIFIED | `/e` (and `/ee`) evaluates the replacement as Perl code | [perldoc perlre](https://perldoc.perl.org/perlre#Modifiers) |

All claims verified. No corrections needed.

---
_Verified by research-verifier agent. Ready for plan-review._

EffortlessSteven:
### Review: improve regex modifier hover details

**Real Perl-feature-gap fix.** The previous hover-text logic only described 5 of the ~15 valid Perl regex modifiers (`i`, `m`, `s`, `x`, `g`), silently dropping any others. So `qr/foo/anu` (ASCII-safe, non-capturing-default, Unicode -- all valid Perl) showed no modifier annotations at all.

**The fix has three correct moves**:

1. **Extract `describe_modifier` as a standalone function**, expanding from 5 known modifiers to 15:

   | Modifier | Perl semantics |
   |---|---|
   | `a` | ASCII-safe character classes |
   | `d` | Native platform character set |
   | `l` | Locale-dependent semantics |
   | `u` | Unicode semantics |
   | `n` | Non-capturing groups by default |
   | `p` | Preserve `${^PREMATCH}`, `${^MATCH}`, `${^POSTMATCH}` |
   | `r` | Non-destructive substitution |
   | `c` | Keep `pos()` after `/g` failure |
   | `o` | Compile pattern once |
   | `e` | Evaluate replacement as code |

   This is the standard `perlop`/`perlre` set. Worth verifying `xx` (Perl 5.26 extended-extended) and `g` are kept consistent -- the test `scenario_hover_text_deduplicates_repeated_known_modifier` uses `xx` and expects single `extended mode` annotation, which is the right behavior for the dedup logic.

2. **Deduplicate repeated modifiers**:

   ```rust
   let mut seen_modifiers: Vec<char> = Vec::new();
   ...
   if seen_modifiers.contains(&modifier) { continue; }
   seen_modifiers.push(modifier);
   ```

   `xx` (extended-extended) and `gg` (typoed) both produce a single annotation. This is a UX win -- previous behavior would have shown the same annotation twice, which is noisy and looks like a parse bug.

3. **Surface unknown modifiers explicitly** instead of silently dropping:

   ```rust
   if !unknown_modifiers.is_empty() {
       let unknown: String = unknown_modifiers.into_iter().collect();
       parts.push(format!("Unknown modifiers: `{unknown}`"));
   }
   ```

   `qr/foo/zzz` now shows `Unknown modifiers: `z`` (deduped to one `z`, not three). This is *good* -- it tells the user the regex won't compile and points at the offending characters. Previous behavior would have shown nothing, leaving users to wonder if `z` was valid.

**Test coverage**:

| Test | What it pins |
|---|---|
| `scenario_hover_text_lists_extended_perl_modifiers` | `anu` -> ASCII-safe, non-capturing, Unicode notes |
| `scenario_hover_text_deduplicates_and_reports_unknown_modifiers` | `iizzz` -> single i note, single z unknown |
| `scenario_hover_text_covers_substitution_specific_modifiers` | `re` -> /r and /e notes |
| `scenario_hover_text_deduplicates_repeated_known_modifier` | `xx` -> single x note, no Unknown section |

The `xx` test is particularly clever -- it's the boundary case where dedup *must* run on a known modifier. A naive "dedupe before describe" implementation could break it.

**Concerns**:

1. **`Vec::contains` is O(n)** in `seen_modifiers.contains(&modifier)`. For 15 modifiers max, this is 15 chars * 15 chars = 225 comparisons worst case. Negligible. Could be `HashSet<char>` for O(1) but the constant overhead of `HashSet` for tiny sets isn't worth it. Leave as-is.

2. **`unknown_modifiers: Vec<char>`** is also deduplicated correctly because `seen_modifiers.contains(&modifier) { continue; }` runs *before* the `match describe_modifier(modifier)` block. Subtle but right.

3. **Modifier order in output is iteration order.** For `anu`, the output reads "ASCII-safe / native / Unicode" -- but `a`, `d`, `u` are mutually exclusive in Perl (`/aa` and `/au` are the only legal combinations). If a user writes the impossible `/au` (impossible because they conflict), the hover will show both. That's defensible -- the user might want to know both meanings even if the combination is illegal -- but worth a note about *combinations* in a future PR.

4. **`Vec::push` with `&'static str`** doesn't allocate (just stores a pointer). Cheap.

5. **The `unknown` string is collected from a `Vec<char>` via `into_iter().collect::<String>()`** -- this is the canonical char-vec-to-string. Allocates one string of up to N chars. Fine.

1. Add a test for `xx` (extended-extended, Perl 5.26+) showing a *different* note vs. plain `x` if the Perl version distinguishes:
   ```rust
   /\s+/xx     # extended-extended: comments-and-whitespace inside `[ ]` too
   ```
   The current dedup means `xx` shows the same note as `x`. This is *probably* what users want for hover (they're not learning the rare distinction in a hover popup), but it's a documented Perl behavior so worth a comment.

2. Consider a "modifier conflicts" warning for `/au`, `/al`, `/ad`, `/ud`, etc.:
   ```
   Note: `a` and `u` are mutually exclusive character-set modifiers.
   ```
   Out of scope for this PR.

3. The `o` modifier note ("compile pattern only once") is true but incomplete. `/o` has been a footgun since Perl 5.6 because variables in `/o` patterns are captured at compile-time -- changes in `$pattern` after first use are ignored. Worth a slightly fuller annotation: `"compile pattern once (variables captured at first use)"`.

**Verdict**: substantive UX improvement that closes a real gap (10 missing modifiers including the common `a`, `u`, `e`, `r`). Tests pin both the new annotations and the dedup/unknown-surface behavior. Worth a couple stylistic refinements, no blockers.

EffortlessSteven:
**Deep review fix applied (f0ffa5f):** Updated the `hover_text_for_regex` doc comment to enumerate all 15 modifiers and document deduplication and unknown-modifier reporting behaviour. The stale list (`i, m, s, x, g`) did not reflect the 10 new modifiers added by this PR.

**Semantic note (//xx, Perl 5.26+):** Perl 5.26 introduced a second-level extended mode where two `x` flags (`//xx`) also strips whitespace from within character classes, which `//x` alone does not. The PR's deduplication collapses `//xx` to a single 'extended mode' note, losing this distinction. Since `hover_text_for_regex` is not yet wired to any LSP provider (no production caller found), the user impact is zero today. A follow-up should handle `//xx` by keeping a count rather than a presence flag, and emitting 'extended mode (level 2): whitespace also stripped inside character classes' when count >= 2.

EffortlessSteven:
diff-audit: clean — 2 files / +101/-12 lines, scope matches title (extended modifier descriptions in `describe_modifier`, dedup + unknown-flag reporting in `hover_text_for_regex`, two new behavior tests). Coherent diff matches PR body. No artifacts, no agent-trail dirs, no banned patterns.

EffortlessSteven:
## green-ci: needs-ci-fix

- PR Smoke (Fast Feedback)

PR #5433, opened 2026-04-23
test(perl-lsp-rs): add BDD semantic tokens incremental workflow (#0000)

EffortlessSteven:
### Motivation
- The BDD workflow suite lacked coverage for semantic-token recomputation after incremental edits, leaving editor highlighting/semantic-refresh behavior without end-to-end assertions.

### Description
- Add `semantic_token_data` helper to normalize `textDocument/semanticTokens/full` payloads into a `Vec<u64>` for deterministic comparisons.
- Add a focused BDD test `bdd_incremental_semantic_tokens_refresh_after_local_symbol_addition` that opens a file, requests semantic tokens, applies an incremental `didChange`, re-requests tokens, and asserts tokens are non-empty, changed, and not regressive in size.
- Changes are limited to `crates/perl-lsp-rs/tests/lsp_bdd_workflows.rs` and only introduce test helpers and a single new scenario.

### Testing
- Ran the new test directly: `cargo test -p perl-lsp-rs --test lsp_bdd_workflows bdd_incremental_semantic_tokens_refresh_after_local_symbol_addition -- --test-threads=1` — passed.
- Ran build validation: `cargo check --all-targets -p perl-lsp-rs` — passed.
- Ran lints for the modified test: `cargo clippy -p perl-lsp-rs --test lsp_bdd_workflows -- -D warnings` — passed.
- Ran formatting via `cargo xtask fmt` as part of the verification steps — completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea0d3524588333b230bfcd7f0b9495)

PR #5434, opened 2026-04-23
test(rename): add package-scoped BDD coverage and guard workspace rename edits (#0000)

EffortlessSteven:
### Motivation
- Rename workflows could incorrectly include same-named subroutines from a different package and this behavior lacked BDD coverage.
- Add a deterministic BDD scenario to protect against regressions in package-scoped renames.
- Tighten workspace rename edit generation to avoid touching declarations or qualified references that belong to a non-target package.

### Description
- Added a BDD test `bdd_rename_is_scoped_to_target_package_symbol` in `crates/perl-lsp-rs/tests/lsp_bdd_workflows.rs` that sets up `Foo` and `Bar` packages exposing the same `process_data` sub and asserts renaming `Foo::process_data` does not affect `Bar`.
- Hardened `build_rename_edit` in `crates/perl-lsp-rs/src/features/workspace_rename.rs` to skip non-target-package hits for `SymKind::Sub` by filtering qualified references whose qualifier does not match the target package and by skipping same-named `sub` declarations in other package scopes.
- Introduced helpers `package_name_for_line`, `is_sub_declaration_line`, and `is_non_target_package_declaration` to detect package context and declaration lines when deciding whether to include an edit.

### Testing
- Ran `cargo test -p perl-lsp-rs --test lsp_bdd_workflows bdd_rename_` and the new BDD tests passed locally (both `bdd_rename_updates_workspace_edits` and `bdd_rename_is_scoped_to_target_package_symbol` reported OK).
- Ran `cargo clippy -p perl-lsp-rs --all-targets -- -D warnings` which surfaced many pre-existing `expect!/panic!` lints in unrelated files (not introduced by this change) and therefore failed; these warnings are outside the scope of this focused change.
- Note: `just pr-fast` is not available in this environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea0d3524588333b230bfcd7f0b9495)

### DIFF
--- a/crates/perl-lsp-rs/src/features/workspace_rename.rs
+++ b/crates/perl-lsp-rs/src/features/workspace_rename.rs
@@ -138,7 +138,7 @@ fn is_sub_declaration_line(line_text: &str, name: &str) -> bool {
 
     let tail = trimmed.trim_start_matches("sub ").trim_start();
     let declared_name = tail
-        .split(|ch: char| ch.is_whitespace() || ch == '(' || ch == '{')
+        .split(|ch: char| ch.is_whitespace() || ch == '(' || ch == '{' || ch == ';')
         .next()
         .unwrap_or_default();
 
@@ -158,18 +158,23 @@ fn is_non_target_package_declaration(
         return false;
     };
 
-    let Some(start_off) = doc.line_index.position_to_offset(start_line, start_char) else {
-        return false;
-    };
-    let Some(end_off) = doc.line_index.position_to_offset(end_line, end_char) else {
-        return false;
-    };
-    let Some(original) = doc.text.get(start_off..end_off) else {
-        return false;
-    };
+    // Try to read the original source span to detect a qualified name like "Bar::process_data".
+    // When the index returns an inverted range (end < start, a known indexer edge case), we fall
+    // through to the package-context check below rather than conservatively returning false.
+    let maybe_original = doc
+        .line_index
+        .position_to_offset(start_line, start_char)
+        .and_then(|start_off| {
+            doc.line_index
+                .position_to_offset(end_line, end_char)
+                .and_then(|end_off| doc.text.get(start_off..end_off))
+        });
 
-    if let Some((qualifier, _)) = original.rsplit_once("::") {
-        return qualifier != key.pkg.as_ref();
+    if let Some(original) = maybe_original {
+        if let Some((qualifier, _)) = original.rsplit_once("::") {
+            return qualifier != key.pkg.as_ref();
+        }
+        // Unqualified bare name — rely on package context below.
     }
 
     let package_at_line = package_name_for_line(&doc.text, start_line);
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Standards review: APPROVE

**Scope**: perl-lsp-rs rename feature + guard logic for non-target package symbols. Clean separation.

**Banned patterns**: PASS — unwrap_or_default() is safe, no expect/panic.

**Tests**: Result<()> pattern correct, helper functions well-structured.

**Cross-PR note**: #5433 also touches lsp_bdd_workflows.rs but tests different subsystem (semantic tokens vs rename). Not a conflict — different line ranges, independent features.

Route to reviewer-deep for correctness.

EffortlessSteven:
**deep-review: APPROVED with fixes**

Two bugs found and fixed in `is_non_target_package_declaration` (workspace_rename.rs):

**Bug 1 — Forward declaration blind spot.** `is_sub_declaration_line` split on whitespace/`(`/`{` but not `;`, so `sub foo;` had its name parsed as `foo;`, never matching the target name. Declarations in non-target packages using the forward-declaration form would slip through the filter and be incorrectly included in rename edits. Fixed by adding `;` to the split delimiter set.

**Bug 2 — Inverted-range fallthrough.** When the workspace indexer emits a location with end_line < start_line (known indexer edge case — observed in test output as `start: (1, 0), end: (0, 12)`), `doc.text.get(start_off..end_off)` returns `None` and the function previously returned `false` unconditionally, bypassing the package-context check. Declarations in non-target packages with malformed index ranges would slip through unfiltered. Fixed by restructuring to fall through to the `package_name_for_line` + `is_sub_declaration_line` path when the range read fails.

Two unit tests added (`is_sub_declaration_line_handles_forward_declaration`, `rename_skips_forward_decl_in_other_package`). The regression test failed before the fix and passes after.

Research verification: SKIPPED (no external Perl/LSP protocol claims — pure Rust string parsing logic).

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-5434",
  "current_version": {
    "sha": "52815c53781371f2795f476352c5c7095432a65d",
    "updated_at": "2026-04-23T14:06:58Z"
  },
  "label_bindings": [
    {
      "label": "deep-reviewed",
      "bound_at_version": "52815c53781371f2795f476352c5c7095432a65d",
      "bound_at_timestamp": "2026-04-23T14:07:49Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Deep-Reviewer Wrapup

**Summary:** reviewer-deep completed correctness review + fix-forward on PR #5434 (workspace rename BDD). 2 bugs fixed, 2 unit tests added, `deep-reviewed` label set.

### Ending State

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/5434
- Changes made: 2 bugs fixed in `crates/perl-lsp-rs/src/features/workspace_rename.rs`, 2 unit tests added
- Files changed: `crates/perl-lsp-rs/src/features/workspace_rename.rs`
- Tests passing: 4/4 workspace_rename unit tests pass
- Status: approved

### Bugs Fixed

**Bug 1 — Forward declaration blind spot (`is_sub_declaration_line`).**
Split on whitespace/`(`/`{` but not `;`. A Perl forward declaration `sub foo;` had its name parsed as `foo;` — never matched, never filtered. Declarations in non-target packages using the forward-declaration form slipped through the package scoping guard and were incorrectly included in rename edits. Fixed: added `;` to the split delimiter set.

**Bug 2 — Inverted-range fallthrough (`is_non_target_package_declaration`).**
When the workspace indexer emits a location with `end_line < start_line` (observed in test output as `start: (1, 0), end: (0, 12)`), `doc.text.get(start_off..end_off)` returns `None` and the function previously returned `false` unconditionally — bypassing the package-context check. Declarations in non-target packages with malformed index ranges slipped through. Fixed: restructured to fall through to the `package_name_for_line` + `is_sub_declaration_line` path when the range read fails.

The regression test `rename_skips_forward_decl_in_other_package` failed before the fix (Bar.pm edits were incorrectly included) and passes after.

### Retrospective

The two bugs were intertwined: Bug 1 made forward declarations invisible to the filter, Bug 2 made malformed-range locations invisible to the filter. Together they meant any non-target-package declaration could bypass the scoping guard via multiple paths. The inverted range (`end < start`) in the indexer was the harder surprise — the guard code made a reasonable assumption that ranges are valid, but the indexer violates this. The fix of restructuring from early-return to optional-read with fallthrough is cleaner than patching each early return.

The `key.pkg != "main"` bypass at line 54 is also logically questionable (it skips filtering for all `main::` sub renames regardless of what the reference is), but since `find_refs()` uses the key to query the index it's unlikely to return cross-package declarations for `main::` subs in practice. Filed as a low-priority note — not fixed here to avoid scope drift.

- Logical next: green-ci check on current HEAD SHA after these fixes
- Gotcha: `is_non_target_package_declaration` has a dependency on the indexer emitting well-formed ranges — the fallthrough fix adds resilience but the root cause (inverted ranges) is in the indexer
- Gotcha: The `key.pkg != "main"` bypass on line 54 of workspace_rename.rs is a latent logic weakness — if the indexer ever returns cross-package refs for main:: subs, the guard is completely bypassed
- Confidence for next: high

EffortlessSteven:
## Diff Audit

**Files changed:** 2
**Lines:** +181 -0
**Commits:** 2 (package-scoped BDD coverage, forward declarations/inverted ranges fix)

### Spec alignment: COMPLETE
- Accepts all acceptance criteria: BDD scenario for package-scoped rename, filters out same-named subroutines in different packages, guards qualified reference checking with package qualifier matching.
- Test fixture `bdd_rename_is_scoped_to_target_package_symbol` explicitly verifies that renaming `Foo::process_data` does not affect `Bar::process_data`.
- Implementation adds `is_non_target_package_declaration`, `is_sub_declaration_line`, and `package_name_for_line` helpers to achieve the filtering.

### Cleanliness: CLEAN
- No TODO, FIXME, HACK, XXX, debug prints, or artifact files.
- All new code is focused on the rename scope guards and test coverage.
- Commit messages follow conventional format (test/fix).

### Commit coherence: CLEAN
- Commit 1 adds the BDD test and implementation guards.
- Commit 2 strengthens handling of forward declarations and inverted ranges.
- Story is clear: add test + implement feature, then harden.

### Verdict: CLEAN
Cumulative diff implements the full acceptance spec, is scoped to rename safety improvements, and contains no artifacts or regressions.

---

PR #5458, opened 2026-04-23
feat(gates): improve metric diff diversity (#0000)

EffortlessSteven:
### Motivation

- Gate receipt diffing only looked at `tests_total`, hiding meaningful changes in other quality signals; increasing metric diversity improves visibility into regressions and improvements.

### Description

- Expand receipt comparison to check all tracked `GateMetrics` fields (`tests_*`, `warnings_count`, `errors_count`, `coverage_percent`, `memory_peak_mb`, `files_checked`, etc.).
- Add a shared helper `push_metric_change` to centralize change detection and delta-percent calculation with safe zero-baseline handling.
- Add unit tests that assert multi-metric reporting and correct delta behavior when the baseline is zero.

### Testing

- Ran `cargo fmt --all`, which passed.
- Attempted `cargo test -p xtask compare_receipts_reports_multiple_metric_dimensions`, but the `xtask` binary failed to compile due to pre-existing, unrelated errors in `xtask/src/tasks/metrics/lsp_stats.rs`, so the new tests could not be executed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea1b033898833398f0f7f67566142e)

### DIFF
--- a/xtask/src/tasks/gates.rs
+++ b/xtask/src/tasks/gates.rs
@@ -1394,7 +1394,7 @@ fn compare_receipts(baseline: &Receipt, current: &Receipt) -> Result<DiffResult>
         }
     }
 
-    // Find metric changes
+    // Find metric changes across all tracked gate metrics.
     let mut metric_changes = Vec::new();
     for (name, current_gate) in &current_gates {
         if let (Some(_baseline_gate), Some(current_metrics), Some(baseline_metrics)) = (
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Standards check:** This PR introduces  calls in new test functions that should use  returns with the  operator instead.

Per coding standards: Tests must use `Result<()>` returns or `perl_tdd_support::must`/`must_some`. No bare `expect()` — use `?` operator.

Lines to fix:
- `xtask/src/tasks/gates.rs`: All three new test functions (`compare_receipts_reports_multiple_metric_dimensions`, `compare_receipts_handles_zero_baseline_delta_without_nan`, `compare_receipts_filters_unchanged_metrics`) have `expect()` calls that should be `?` with Result return type.

**Fix approach:**
1. Change each test function signature to return `Result<()>`
2. Replace `.expect(...)` with `?` operator
3. Add `Ok(())` at end of functions

This is a blocker for merge-readiness. Please fix before routing to reviewer-deep.

EffortlessSteven:
## Deep Review - FIX-FORWARDED

### Bug found and fixed: test helper built against wrong struct schema

The test helper test_receipt_with_metrics was constructed with field names that do not exist on Receipt/ReceiptMetadata:
- Used results: vec![...] but the struct field is gates: Vec<GateResult>
- ReceiptMetadata fields repository, trigger (as top-level), actor, workflow_name, ci_run_url, pr_number, nix_shell do not exist; the actual struct requires git_sha_short, git_dirty, toolchain: ToolchainInfo, platform: PlatformInfo, environment: EnvironmentInfo
- Receipt also requires schema_version: String and agent_receipt: AgentReceipt (no Default impl), both missing
- Imported ReceiptMetadata and ReceiptSummary in use super{} but the fixed helper no longer constructs them by hand

The PR author's test environment had pre-existing lsp_stats.rs compile errors that blocked all xtask compilation, so these test errors were never caught.

Fix applied (commit 2b090dc): replaced the struct-literal helper with a serde_json::from_str() approach using a minimal JSON skeleton. compare_receipts only reads receipt.gates and receipt.metadata.timestamp, so all other fields can be placeholder values. The GateResult with the desired metrics is pushed after deserialization.

### push_metric_change logic verified correct
- Zero-baseline division: when old_value.abs() < EPSILON and new_value != 0, returns 100.0 - correct convention.
- When both old and new are near-zero, returns 0.0 - correct.
- Change skipped when (old - new).abs() < EPSILON - correct (no noise from floating-point equality on integer metrics cast to f64).
- exceeds_threshold fires at delta_percent.abs() > 10.0 - matches existing behavior.

### All ten metrics wired correctly
tests_total, tests_passed, tests_failed, tests_skipped, tests_ignored, warnings_count, errors_count all use .map(f64::from) on Option<u32> - correct widening. coverage_percent and memory_peak_mb are already Option<f64> - passed directly. files_checked uses .map(f64::from) - correct.

deep-reviewed: FIX-FORWARDED

EffortlessSteven:
## Refactor Plan

### 1. Extract helper already present: `push_metric_change()`
- **Where:** `xtask/src/tasks/gates.rs:1491-1520`
- **What:** Good extraction! Eliminates duplication across 10 metric comparison calls and centralizes delta calculation logic.

### 2. Simplify: Zero-baseline division handling
- **Where:** `xtask/src/tasks/gates.rs:1505-1511`
- **What:** The special case for zero baselines is correct and well-tested. Code is clear.
- **How:** No changes needed. The epsilon check + zero-baseline guard is the right pattern.

### 3. Test helper: `test_receipt_with_metrics()`
- **Where:** `xtask/src/tasks/gates.rs:1693-1746`
- **What:** Hardcoded JSON structure is verbose but acceptable for a minimal test fixture.
- **How:** Could be simplified with a derive(Default) fixture, but the current approach is explicit and verifiable. Not worth changing.

### Skip (not worth changing)
- The repetitive `push_metric_change()` calls (lines 1410-1479) are intentional and clear — this is the canonical shape for metric diversity. Attempting to loop over field names would make the intent harder to read.

Overall: Clean extraction pattern well-executed. The PR demonstrates healthy refactoring discipline. Ready for green-refactor — no issues.

PR #5476, opened 2026-04-23
fix(perl-lsp-rs): harden incremental edit mapping (#0000)

EffortlessSteven:
### Motivation
- Prevent constructing invalid original-document byte offsets when a single `didChange` batch contains out-of-order ranges, which could underflow mapping and produce corrupted incremental edits.

### Description
- Add guarded original-space mapping in `build_incremental_edit_set` to use `checked_sub` and return `None` (fall back to full reparse) when mapping would produce invalid or negative offsets and emit `tracing::debug` logs on failure.
- Validate that `orig_start <= orig_end` and defensively abort the incremental path when the mapping is invalid.
- Add a regression test `test_build_incremental_edits_rejects_invalid_original_mapping` that simulates an out-of-order change batch and asserts the builder returns `None`.
- Changes are localized to `crates/perl-lsp-rs/src/runtime/text_sync.rs`.

### Testing
- Ran `cargo test -p perl-lsp-rs --lib test_build_incremental_edits_ --features incremental`, and the two new/related unit tests passed (`2 passed; 0 failed`).
- Ran `cargo check --all-targets -p perl-lsp-rs --features incremental`, which completed successfully.
- `cargo xtask fmt` and `cargo clippy -p perl-lsp-rs --features incremental` were attempted but could not complete in this environment due to unrelated `xtask` compile errors and temporary disk-space limitations; these are environment issues and unrelated to the code changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea1b1fda6c8333bf99f686424c48cd)

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/text_sync.rs
+++ b/crates/perl-lsp-rs/src/runtime/text_sync.rs
@@ -63,8 +63,36 @@ fn build_incremental_edit_set(
 
         // Map back to original-document space by undoing the byte shift that
         // prior edits introduced into the working rope.
-        let orig_start = (evolving_start as isize - cumulative_shift) as usize;
-        let orig_end = (evolving_end as isize - cumulative_shift) as usize;
+        //
+        // Some clients may send out-of-order ranges in a single didChange
+        // payload. In that case the shift can temporarily exceed
+        // `evolving_start`/`evolving_end`; we defensively abort incremental
+        // mapping and fall back to full reparse.
+        let Some(orig_start) = (evolving_start as isize).checked_sub(cumulative_shift) else {
+            tracing::debug!(
+                "Incremental edit mapping failed: evolving_start={} cumulative_shift={}",
+                evolving_start,
+                cumulative_shift
+            );
+            return None;
+        };
+        let Some(orig_end) = (evolving_end as isize).checked_sub(cumulative_shift) else {
+            tracing::debug!(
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review - FIX-FORWARDED**

The logic is correct: the `orig_start < 0` check is the real guard for out-of-order edits. The `checked_sub` pattern was misleading because `isize::checked_sub` returns `None` only on isize arithmetic overflow (i.e., result < isize::MIN), which cannot happen for document byte offsets. In practice, a negative mapping result always arrives as `Some(-3)`, never as `None`.

**Fix pushed (commit `4137a8b`):** Replaced the two `let Some(...) = .checked_sub(...) else { ... }` arms with direct subtraction, relying solely on the subsequent `< 0` guard. The debug log message is also expanded to include the evolving position and shift in a single structured message (easier to correlate). The test `test_build_incremental_edits_rejects_invalid_original_mapping` still correctly exercises the fallback path (orig_start = -3 is negative, caught by `< 0`).

EffortlessSteven:
Duplicate of #5513. Both fix `build_incremental_edit_set` by adding `checked_sub`-style guards when mapping evolving-document offsets back to original-document space. #5513 is a scope-clean 51/2 diff; this PR touches the entire file (2494/2443 lines) because of line-ending normalization, which will conflict with every other text_sync.rs PR in flight. Closing this one in favor of the targeted fix.

PR #5503, opened 2026-04-23
fix(xtask): harden ci-scope base fallback reporting (#0000)

EffortlessSteven:
### Motivation

- `cargo xtask ci-scope` could fall back to a different git ref but still report the user-provided invalid base in output, and only had a single fallback choice, making diagnostics and downstream automation confusing.
- The goal is to robustly resolve a usable base ref from a prioritized set of fallbacks and ensure the resolved base is persisted in the output so consumers see the actual diff baseline.

### Description

- Add a `ref_exists` helper and a prioritized `fallback_candidates` list (`origin/main`, `origin/master`, `main`, `master`, `HEAD~1`, `HEAD`) to resolve base refs more robustly in `xtask/src/tasks/ci_scope.rs`.
- Return the resolved base ref from `resolve_base_ref` and persist it to `output.base` instead of echoing the original `config.base` value.
- Add an integration test `test_ci_scope_invalid_base_reports_resolved_fallback` in `xtask/tests/ci_scope_tests.rs` that verifies an invalid `--base` is resolved to a non-empty fallback and reported in the JSON output.
- Keep behavior of printing a user-facing warning when a fallback is used and fail only when no candidates exist.

### Testing

- Ran `cargo fmt -p xtask --check` which passed for the modified xtask crate.
- Ran `cargo test -p xtask --test ci_scope_tests` which attempted to run but the xtask build failed due to pre-existing unrelated compile errors in `xtask/src/tasks/metrics/lsp_stats.rs` (these errors are not caused by this change).
- Noted that a workspace-wide `cargo fmt --all --check` previously reported unrelated formatting drift outside the changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea1d8ac3348333b5d38fb6ddce896a)

### DIFF
--- a/xtask/src/tasks/ci_scope.rs
+++ b/xtask/src/tasks/ci_scope.rs
@@ -731,7 +731,7 @@ pub fn run(config: CiScopeConfig) -> Result<()> {
     let workspace_root = root.to_string_lossy().replace('\\', "/");
 
     let mut output = classify_files(&changed_files, &metadata, &workspace_root)?;
-    output.base = config.base.clone();
+    output.base = base_ref.clone();
     output.head_sha = head_sha;
     output.changed_files = changed_files;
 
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
## Deep Review: #5503 fix(xtask): harden ci-scope base fallback reporting

Reviewed alongside #5504 (same Codex task, wider scope).

### What was correct

- Core bug fix verified: `output.base = config.base.clone()` on master reported the *input* ref even when the code had already resolved a different fallback. The change to `output.base = base_ref.clone()` is the right fix.
- `ref_exists` inner function extracts correctly, subprocess called via `duct::cmd` array args — no command injection risk.
- Warning emission to stderr on fallback is correct behavior.
- Test G uses `assert_ne!(base, "this-ref-should-not-exist")` — this assertion would have **caught the pre-fix bug** where `output.base = config.base.clone()` was still in place.

### What I fixed on the branch (commit 694b1e21a)

**1. HEAD removed from fallback chain (correctness bug)**
`["origin/main", "origin/master", "main", "master", "HEAD~1", "HEAD"]` — `HEAD` as last resort causes `git diff HEAD...HEAD` which returns an empty file list. This silently reports zero changed files and skips all CI lanes: a false-negative worse than a clean error. Removed `HEAD`, added explanatory comment. The superior PR #5504 had already noticed this; applied the same fix here.

**2. `color_eyre::eyre::eyre!` → `eyre!` (style)**
`eyre` is already imported at the top of the file. Redundant fully-qualified path removed.

**3. Error message improved**
Now lists all tried candidates and hints at root cause ("Ensure the repository has at least one commit and the remote is reachable.").

**4. Test G strengthened**
- Added early-return guard for no-remote environments (single-commit repos where even `HEAD~1` is unreachable) — prevents spurious test failures in unusual CI contexts.
- Added `assert!(stderr.contains("Warning:"))` — verifies the *reporting* path (not just exit code). Without this assertion, the test would pass even if the warning was silently dropped.

**5. Test H added**
New `test_ci_scope_head_not_used_as_fallback` explicitly asserts `base != "HEAD"` in the resolved output. This directly tests the correctness property that motivated removing HEAD from the chain.

### Pair analysis with #5504

#5504 is the wider PR from the same Codex task. It adds:
- `--base auto` as the new CLI default (suppresses warning when auto-resolving)
- Same fallback fix applied to `targeted_checks.rs` (which #5503 doesn't touch)
- Test H for `auto` behavior (no warning emitted when `--base auto` is used)
- Already excludes `HEAD` from the fallback chain with the same explanatory comment

**Recommendation:** If both PRs are merge candidates, prefer #5504 — it is strictly a superset. Close #5503 as superseded, or merge #5503 first and rebase #5504 on top. Do NOT merge both as-is; the `ci_scope.rs` changes conflict.

### Verdict

Logic correct after my fixes. Regression risk low. `deep-reviewed` set.

EffortlessSteven:
Closing as superseded by #5504. Your work informed the winner — same HEAD-in-fallback bug was fixed on both branches by deep-reviewer. #5504 is a strict superset (ci_scope.rs + targeted_checks.rs + --base auto CLI default). Deep-reviewer cross-reference: https://github.com/EffortlessMetrics/perl-lsp/pull/5504

PR #5608, opened 2026-04-23
fix(vscode): improve Warp terminal command fallback (#0000)

EffortlessSteven:
### Motivation
- Onboarding health checks can fail to discover tools when run in terminal emulators (e.g. Warp) that populate `PATH` or shims via shell startup scripts, causing ENOENT probes to misclassify the environment.

### Description
- Add `resolveUnixShellInvocationFallback(invocation, err)` to run a failing command via the user's login shell (`SHELL -lc ...`) when an ENOENT occurs on non-Windows platforms.
- Add `toPosixShellCommand` and `escapePosixShellArg` to construct a POSIX-safe, single-quoted command string for `-lc` invocation.
- Update `defaultExecCheck` to attempt the existing Windows fallback then the new Unix shell fallback before rethrowing the error.
- Add unit tests in `vscode-extension/src/test/onboarding.test.ts` that cover POSIX quoting and the Unix shell fallback resolution, and export the new helpers for testing.

### Testing
- Ran `cd vscode-extension && npm test -- onboarding.test.ts`, which attempted to run the onboarding tests but failed in this environment with TypeScript/Jest global typing errors (`Cannot find name 'jest'/'describe'`) before tests could execute successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea213670208333802f2aca26a5150e)


EffortlessSteven:
**deep-reviewed** — Approved.

**Logic verified:** The POSIX single-quote escaping `'${value.replace(/'/g, "'\''")}'` is correct. A single quote in the input becomes `'\''` — close the current single-quoted string, add a literal backslash-escaped single quote, reopen. This is the canonical POSIX escape sequence. The test `"'perl' '-e' 'print q{can'\''t fail}'"` correctly demonstrates the pattern.

**Guard for path injection:** The `invocation.command.includes('/')` check prevents the fallback from being used for absolute path commands, which already have full path resolution. This is correct — only simple command names (e.g. `perl`) should go through the login shell lookup.

**isSpawnNotFound type safety:** The function checks `err && typeof err === 'object' && 'code' in err && err.code === 'ENOENT'` — correctly handles unknown error shapes. The test passes `{ code: 'ENOENT' }` which is the Node.js spawn error shape.

**One edge case not tested:** What if `SHELL=/usr/bin/fish`? Fish does not accept `-lc` as arguments in the same way as bash/zsh. The current implementation will attempt `fish -lc 'perl ...'` which may work (fish supports `-lc`) but the `--login` behavior differs. This is an acceptable limitation for the initial implementation — document it in a follow-up.

**Test environment note:** The test for `resolveUnixShellInvocationFallback` uses `Object.defineProperty(process, 'platform', ...)` which works in Node.js test environments. The describe block correctly saves and restores both `platform` and `SHELL`.

EffortlessSteven:
## Diff Audit

**Files changed:** 2
**Commits:** 2

### Spec alignment: COMPLETE
All changes are focused on the stated issue with no extraneous additions.

### Cleanliness: CLEAN
No debug artifacts, TODO/FIXME comments, or banned patterns found.

### Commit coherence: CLEAN
Commit history is logical and descriptive.

### Verdict: CLEAN
Diff is coherent, scoped appropriately, and ready for merge.

---

PR #5640, opened 2026-04-23
fix(symbol-query): improve Droid Facxtory matching (#0000)

EffortlessSteven:
### Motivation

- Workspace symbol search missed matches when queries used different separators or had small typos (e.g. `Droid Facxtory` vs `Droid::Factory`).
- Make symbol matching more robust to separator variants and single-character typos so users find expected symbols more reliably.

### Description

- Add `normalize_for_match` to strip non-alphanumeric characters and lowercase names/queries to make matching separator-insensitive. 
- Add `is_single_edit_typo` helper and a dedicated typo tier so single-edit typos rank above fuzzy/subsequence matches. 
- Wire normalization and typo-tolerance into `matches_query`, `match_tier`, and ranking via `compare_names_by_query`. 
- Update unit tests in `symbol_query` to cover typo-ranking and `Droid::Factory` / `DroidFactory` vs `Droid Facxtory` scenarios and update doc comments to reflect the new matching order.

### Testing

- Ran `cargo test -p perl-lsp-rs-core providers::symbol_query::tests::` and the symbol-query test suite passed (8 tests in the module passed).
- Ran `cargo check --all-targets -p perl-lsp-rs-core` and it completed successfully.
- Ran `cargo clippy -p perl-lsp-rs-core -- -D warnings` which completed without warnings for the modified crate.
- `cargo xtask fmt` could not complete cleanly due to unrelated existing compile errors in `xtask/src/tasks/metrics/lsp_stats.rs` (pre-existing unresolved names), so formatting gate remains blocked by that unrelated issue.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea22c4086c8333a1971bb2aa7b302e)

PR #5639, opened 2026-04-23
feat(completion): add typo-tolerant module suggestions (#0000)

EffortlessSteven:
### Motivation

- `use`/`require` module completion only matched exact prefixes, so small typos (e.g. `Droid::Facxtory`) could hide the intended workspace module suggestion.

### Description

- Added a bounded Levenshtein implementation `levenshtein_distance` and a scoped matcher `likely_module_typo_match` to identify plausible typos within the same top-level namespace.
- Updated `add_use_module_completions` to collect package symbols and fall back to typo-tolerant matching when no exact-prefix matches are found.
- Added a regression test `test_use_statement_typo_still_suggests_workspace_module` verifying that `use Droid::Facxtory` suggests `Droid::Factory`.

### Testing

- Ran `cargo test -p perl-lsp-rs-core test_use_statement_typo_still_suggests_workspace_module`, which passed.
- Ran `cargo check --all-targets -p perl-lsp-rs-core`, which completed successfully.
- Ran `cargo clippy -p perl-lsp-rs-core -- -D warnings`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea22c4086c8333a1971bb2aa7b302e)

PR #5619, opened 2026-04-23
feat(perl-semantic-analyzer): add Droid web-route detection (#0000)

EffortlessSteven:
### Motivation

- Recognize `use Droid;` as a web framework so Droid route declarations are synthesized for navigation/analysis features.

### Description

- Add `WebFrameworkKind::Droid` and update `FrameworkFlags` docs to include `Droid` in web framework detection.
- Extend route extraction (`try_extract_web_route_declaration`) to treat `Droid` like `Dancer`/`Dancer2` for synthesizing route `Subroutine` symbols and for recording string-target subroutine references.
- Map the module name `

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea216974e08333b0059dca98637868)

PR #5642, opened 2026-04-23
feat(workspace-index): add typo-tolerant symbol search fallback (#0000)

EffortlessSteven:
### Motivation
- Symbol search previously only returned substring matches so small typos in package names (e.g. `Droid::Facxtory`) failed to resolve to the intended symbol (`Droid::Factory`).

### Description
- Add a fallback fuzzy match in `WorkspaceIndex::search_symbols` that runs when the query is non-empty and the substring search returns no results. 
- Implement `is_approximate_symbol_match` and `is_edit_distance_leq_one` helpers to accept matches within one edit and to test both full qualified names and `::`-segmented name parts. 
- Match fuzzy results into the existing `WorkspaceSymbol` result vector so completion/lookup consumers receive the same symbol objects.
- Add a regression test `test_search_symbols_fuzzy_typo_fallback` which indexes `Droid::Factory` and verifies a query for `Facxtory` returns the package.

### Testing
- Ran `cargo test -p perl-workspace test_search_symbols_fuzzy_typo_fallback` and the new test passed. 
- Ran `cargo check --all-targets -p perl-workspace` and it succeeded. 
- Ran `cargo clippy -p perl-workspace -- -D warnings` and it succeeded. 
- Ran `cargo fmt -p perl-workspace` and it completed; `cargo xtask fmt` failed due to unrelated `xtask` compile errors in `xtask/src/tasks/metrics/lsp_stats.rs` not introduced by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea22c4086c8333a1971bb2aa7b302e)

### DIFF
--- a/crates/perl-workspace-index/src/workspace/workspace_index.rs
+++ b/crates/perl-workspace-index/src/workspace/workspace_index.rs
@@ -2220,6 +2220,7 @@ impl WorkspaceIndex {
         let query_lower = query.to_lowercase();
         let files = self.files.read();
         let mut results = Vec::new();
+
         for file_index in files.values() {
             for symbol in &file_index.symbols {
                 if symbol.name.to_lowercase().contains(&query_lower)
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as Codex hallucination cascade. Third PR in a set (with #5639, #5640) all implementing slightly different versions of typo-tolerant symbol search, each regression-tested with the `Droid::Factory` / `Facxtory` pattern derived from the closed Factory.ai Droid hallucinations (#5619/#5641). If fuzzy symbol search is a real UX need, please open a single consolidated issue with real-world motivation (not a fabricated namespace) and let the builder pick the right layer to implement it. See `feedback_codex_framework_hallucination.md`.

PR #5659, opened 2026-04-23
fix(critic-parser): validate perlcritic coordinates and severity (#0000)

EffortlessSteven:
### Motivation

- Prevent malformed Perl::Critic records from producing invalid LSP diagnostics by refusing entries with `line` or `column` equal to `0` or with severities outside the supported `1..=5` range.

### Description

- Add validation in `parse_perlcritic_line` to reject records where `line == 0` or `column == 0` or `severity` is not in `1..=5`.
- Preserve existing parsing logic and only drop malformed entries early (returns `None`).
- Add unit tests in `crates/perl-lsp-rs-core/tests/critic_parser_module_shape.rs` that assert rejection of zero coordinates and out-of-range severities.
- Changes applied to `crates/perl-lsp-rs-core/src/critic_parser/mod.rs` and the test file mentioned above.

### Testing

- Ran `cargo test -p perl-lsp-rs-core --test critic_parser_module_shape` and all tests passed (`7 passed`).
- Ran `cargo check --all-targets -p perl-lsp-rs-core` and the check completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea2cc64b548333ace58de19ba3f190)

### DIFF
--- a/crates/perl-lsp-rs-core/src/critic_parser/mod.rs
+++ b/crates/perl-lsp-rs-core/src/critic_parser/mod.rs
@@ -66,6 +66,10 @@ pub fn parse_perlcritic_line(line: &str) -> Option<ParsedCriticLine> {
     let column = parts[start + 1].parse::<u32>().ok()?;
     let severity = parts[start + 2].parse::<u8>().ok()?;
 
+    if line_num == 0 || column == 0 || !(1..=5).contains(&severity) {
+        return None;
+    }
+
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Duplicate of #5660. Both add 'reject zero line/column + out-of-range severity' validation to critic_parser/mod.rs with parallel tests; #5660 also handles CRLF trim and is more comprehensive. Keeping #5660.

PR #5666, opened 2026-04-23
fix(perl-lsp-perltidy): improve built-in fallback formatter (#0000)

EffortlessSteven:
### Motivation

- The existing built-in fallback formatter only adjusted indentation for braces/parens at line boundaries which produced incorrect indentation for multiline argument lists. 
- Delimiter-like characters inside string literals or trailing comments could corrupt the fallback formatter's balance tracking, making `perltidy` required for reasonable results.

### Description

- Replace the simplistic line-start/line-end checks with a per-line delimiter-balance model driven by `count_leading_closers` and `net_delimiter_delta` helpers. 
- `net_delimiter_delta` walks a line and updates a net brace/paren/bracket delta while ignoring delimiters inside single/double quoted strings, escaped characters, and text after `#` comment markers. 
- Use the new helpers to adjust `indent_level` before and after emitting each line so multiline constructs (e.g., argument lists, arrays) are indented sensibly. 
- Add regression tests `builtin_formatter_indents_multiline_function_arguments` and `builtin_formatter_ignores_delimiters_inside_strings_and_comments` to cover multiline indentation and string/comment safety.

### Testing

- Ran `cargo test -p perl-lsp-perltidy`, and all tests passed (`26 + 22 + 5` tests across suites passed). 
- Ran `cargo check --all-targets -p perl-lsp-perltidy`, which completed successfully. 
- Ran `cargo clippy -p perl-lsp-perltidy --all-targets -- -D warnings`, which completed successfully. 
- `cargo xtask fmt` failed during a workspace `xtask` build due to unrelated pre-existing errors in `xtask/src/tasks/metrics/lsp_stats.rs`, and is not caused by these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea2cd9b1ec8333b0cdf73ee9632d45)

### DIFF
--- a/crates/perl-lsp-perltidy/src/lib.rs
+++ b/crates/perl-lsp-perltidy/src/lib.rs
@@ -351,7 +351,12 @@ impl BuiltInFormatter {
             }
             result.push('\n');
 
-            indent_level = (indent_level + net_delimiter_delta(trimmed)).max(0);
+            // net_delimiter_delta counts all delimiters including leading closers.
+            // We already decremented by leading_closers before printing, so add them
+            // back to avoid double-counting: the net change for the *next* line is
+            // delta + leading_closers (leading closers cancel in the net formula).
+            indent_level =
+                (indent_level + net_delimiter_delta(trimmed) + leading_closers).max(0);
         }
 
         result
--- a/crates/perl-lsp-perltidy/src/lib.rs
+++ b/crates/perl-lsp-perltidy/src/lib.rs
@@ -340,9 +340,8 @@ impl BuiltInFormatter {
 
         for line in code.lines() {
             let trimmed = line.trim();
-            if trimmed.starts_with('}') || trimmed.starts_with(')') || trimmed.starts_with(']') {
-                indent_level = indent_level.saturating_sub(1);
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
**deep-reviewed**: Fixed a correctness bug in the built-in fallback formatter.

**Bug found and fixed**: `net_delimiter_delta` counts all delimiters including leading closers. `count_leading_closers` already subtracts these before printing, so they were double-counted after printing. For a line like `}`, `indent_level` would drop by 2 instead of 1; for `})`, it would drop by 4 instead of 2. Fixed by adding `leading_closers` back to the post-print adjustment to cancel the double-count.

**Fix pushed directly**: Added correction to `format()` in `lib.rs` and two regression tests (`closing_line_does_not_double_decrement`, `multi_closer_line_does_not_over_decrement`). All 28 config tests pass.

**Other aspects verified**: `net_delimiter_delta` correctly skips delimiters inside single/double quoted strings and after `#` comments. The pre-existing lack of heredoc-body awareness is a known limitation of the fallback formatter, not introduced here.

PR #5672, opened 2026-04-23
fix(actions): add Windows ARM64 setup target (#0000)

EffortlessSteven:
### Motivation
- The `setup-perl-lsp` composite action rejected Windows ARM64 runners even though ARM binaries are published, causing CI consumers on Windows/ARM to hard-fail during platform resolution.

### Description
- Map `RUNNER_ARCH=ARM64` to `aarch64-pc-windows-msvc` in `.github/actions/setup-perl-lsp/action.yml` and document supported auto-resolved architectures in `.github/actions/README.md`.

### Testing
- Verified the updated `action.yml` parses with `ruby -e "require 'yaml'; YAML.safe_load_file('.github/actions/setup-perl-lsp/action.yml', permitted_classes: [], aliases: true); puts 'ok'"` (passes), and attempted `cargo xtask fmt` which failed due to unrelated compile errors in `xtask/src/tasks/metrics/lsp_stats.rs` (pre-existing and not caused by this change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea336df5b48333bfe5b6c0603fd1c5)

### DIFF
--- a/.github/actions/README.md
+++ b/.github/actions/README.md
@@ -8,6 +8,11 @@ This directory contains reusable composite actions for perl-lsp CI workflows.
 
 Installs `perllsp` for downstream GitHub Actions workflows.
 
+Supported release binaries auto-resolve for:
+- Linux: `x86_64`, `aarch64`
+- macOS: `x86_64`, `aarch64`
+- Windows: `x86_64`, `aarch64`
+
 **Usage:**
 ```yaml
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
**deep-reviewed**: One-line shell case addition for Windows ARM64 (`aarch64-pc-windows-msvc`). Correct target triple. README addition is accurate. No logic concerns.

PR #5673, opened 2026-04-23
fix(transport): tolerate LF-only LSP header separators (#0000)

EffortlessSteven:
### Motivation

- Some non-conformant LSP clients/CLI integrations send `Content-Length` header blocks terminated with LF-only (`\n\n`) instead of the standard CRLF (`\r\n\r\n`), which caused the stateful transport reader to reject frames.

### Description

- Update `crates/perl-lsp-rs-core/src/transport/framing.rs` to accept both CRLF (`\r\n\r\n`) and LF-only (`\n\n`) header terminators by adding `find_header_end` and tracking the separator length. 
- Use line-based header parsing (`header.lines()`) and a `separator_len` parameter to correctly consume header blocks and compute the body start regardless of newline style. 
- Adjust `frame()` to remain CRLF-based for outgoing frames and keep `Content-Length` generation unchanged. 
- Add a focused regression test `stateful_reader_accepts_lf_only_header_separator` that verifies the stateful reader accepts LF-only header separators.

### Testing

- Ran `cargo test -p perl-lsp-rs-core stateful_reader_accepts_lf_only_header_separator`, which passed. 
- Ran `cargo check --all-targets -p perl-lsp-rs-core`, which passed. 
- Ran `cargo clippy -p perl-lsp-rs-core -- -D warnings`, which passed. 
- `cargo fmt --all` was invoked but failed due to an unrelated pre-existing syntax error in another test file and did not block this focused change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea337be6c88333914546cb3b30f87f)

### DIFF
--- a/crates/perl-lsp-rs-core/src/transport/framing.rs
+++ b/crates/perl-lsp-rs-core/src/transport/framing.rs
@@ -13,7 +13,8 @@ use std::io::{self, BufRead, Read, Write};
 // ── Absorbed from perl-content-length-framing ─────────────────────────────────
 
 const HEADER_SENTINEL: &[u8] = b"Content-Length:";
-const HEADER_END: &[u8] = b"\r\n\r\n";
+const HEADER_END_CRLF: &[u8] = b"\r\n\r\n";
+const HEADER_END_LF: &[u8] = b"\n\n";
 const RESYNC_TAIL_BYTES: usize = 8 * 1024;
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Duplicate of #5682. Both titled 'tolerate LF-only framed headers' and modify crates/perl-lsp-rs-core/src/transport/framing.rs; #5682 has larger diff (+74 vs +48). Keeping #5682.

PR #5680, opened 2026-04-23
fix(perl-lsp-rs-core): improve Termux perl discovery (#0000)

EffortlessSteven:
### Motivation

- Resolve cases where `resolve_perl_path()` could fail in Termux sessions when `PATH` is unset or minimal, and the user-facing guidance omitted Termux install instructions.

### Description

- Stop treating a missing `PATH` as an immediate hard error and search `PATH` only when present in `resolve_perl_path()`.
- Add `termux_perl_candidates()` which returns `$PREFIX/bin/perl` (when `PREFIX` is set) and the default Termux prefix `/data/data/com.termux/files/usr/bin/perl` for fallback discovery.
- Update `perl_not_found_install_message()` to mention Termux install command `pkg install perl`.
- Add focused unit tests `termux_candidates_include_prefix_bin_perl` and `install_message_mentions_termux_pkg_command` to cover the new Termux behavior.

### Testing

- Ran the two new targeted tests with `cargo test -p perl-lsp-rs-core --lib platform::tests::termux_candidates_include_prefix_bin_perl` and `cargo test -p perl-lsp-rs-core --lib platform::tests::install_message_mentions_termux_pkg_command`, both passed.
- Ran `cargo check --all-targets -p perl-lsp-rs-core` and `cargo clippy -p perl-lsp-rs-core`, both completed successfully.
- A full `cargo test -p perl-lsp-rs-core` run was executed and most tests passed, but one existing test failure unrelated to this change (expecting `crates/perl-lsp/Cargo.toml`) was observed during the full-suite run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea3369341c8333953253f71b46233e)

### DIFF
--- a/crates/perl-lsp-rs-core/src/platform.rs
+++ b/crates/perl-lsp-rs-core/src/platform.rs
@@ -3,7 +3,7 @@
 //! Extracted from `perl-dap::platform` to break the config→dap cycle
 //! and serve as a stable, reusable service layer for both LSP and DAP consumers.
 
-use anyhow::{Context, Result};
+use anyhow::Result;
 use std::env;
 use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**deep-reviewed**: Termux perl discovery logic is correct.

**Verified**: The fallback order is correct — PATH is tried first, then Termux-specific paths (`$PREFIX/bin/perl`, then hard-coded `/data/data/com.termux/files/usr/bin/perl`). This is the right priority since PATH should win over Termux defaults. The `PREFIX` env var is standard for Termux. The test correctly saves and restores `PREFIX` around the assertion.

**One concern**: `unsafe { std::env::set_var("PREFIX", ...) }` — mutating process env in tests is inherently racy if tests run in parallel. The test works in practice because `termux_perl_candidates` is the only reader, and the test is isolated. This is a pre-existing pattern in the codebase.

PR #5682, opened 2026-04-23
fix(perl-lsp-rs-core): tolerate LF-only framed headers (#0000)

EffortlessSteven:
### Motivation
- Some CLI LSP clients (e.g., Gemini CLI style transports) emit `Content-Length` header blocks using LF-only separators (`\n\n`) instead of CRLF (`\r\n\r\n`), which caused the server to reject or skip headers from those clients.
- The server should be tolerant when reading inbound frames while still emitting standards-compliant CRLF framing outbound to maximize interoperability.

### Description
- Teach the `ContentLengthFramer` reader to detect both CRLF (`\r\n\r\n`) and LF-only (`\n\n`) header boundaries via a new `find_header_end` helper and `HEADER_END_LF` constant.
- Make header-line parsing newline-agnostic by iterating with `.lines()` and trimming trailing `\r` so mixed line endings are handled consistently.
- Keep outbound framing unchanged (writer still emits CRLF) by continuing to use `HEADER_END_CRLF` in `frame`.
- Add regression tests covering LF-only single-frame and back-to-back frames in the `framing` test suite to lock in the behavior.

### Testing
- Ran `cargo test -p perl-lsp-rs-core framing` which completed successfully (all framing tests passed). 
- Ran `cargo check --all-targets -p perl-lsp-rs-core` which completed successfully. 
- Ran `cargo xtask fmt` which failed due to unrelated pre-existing errors in `xtask` unrelated to this change. 
- Ran `cargo fmt --all` which failed due to unrelated pre-existing syntax issues in a test file unrelated to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea337be6c88333914546cb3b30f87f)


EffortlessSteven:
## Research Verification

**Summary:** 1 of 1 claim VERIFIED (LSP protocol compliance)

### LSP/DAP Spec Claims

| Claim | Status | Finding | Source |
|-------|--------|---------|--------|
| LSP protocol mandates CRLF (\r\n\r\n) as header separator; LF-only (\n\n) is non-compliant | VERIFIED | LSP 3.17 spec explicitly requires CRLF: 'Each header field must be terminated by \r\n, and two \r\n sequences always immediately precede the content part.' The spec states header part is 'encoded using ascii encoding. This includes the \r\n separating the header and content part.' No LF-only alternative is permitted by the spec. | [LSP 3.17 Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification) |

All external claims verified. PR correctly identifies the protocol requirement (CRLF is mandatory) and implements tolerant input (accepts both CRLF and LF-only) while maintaining standards-compliant output (CRLF only). No corrections needed.

---
_Verified by research-verifier agent. Ready for plan-review._

EffortlessSteven:
**deep-reviewed** — Approved.

**LF-only framing logic:** The `find_header_end` function searches for both `\r\n\r\n` and `\n\n`, returning the one that appears first. This correctly handles: (1) pure CRLF frames, (2) pure LF-only frames, (3) mixed headers where CRLF wins if it appears earlier. The CRLF-first priority is correct — well-formed LSP clients always send CRLF, so only fall back to LF when CRLF isn't found.

**False-positive risk:** Minimal. The `\n\n` pattern only fires when no `\r\n\r\n` appears earlier in the buffer. JSON bodies come AFTER the header terminator, so the first `\n\n` in the buffer should always be the header end.

**parse_content_length:** Also updated to use `header.lines()` + `trim_end_matches("\r")`. Correctly strips trailing `\r` for CRLF lines and is a no-op for LF-only lines.

**write_message:** Still emits CRLF (correct — write always follows spec; read is the lenient side).

**Tests:** Two new tests (single frame and back-to-back frames with LF-only) are correct and comprehensive.

EffortlessSteven:
## Diff Audit

**Verdict: CLEAN** — cumulative diff is coherent, scope is tight, no artifacts.

### Scope
- Single-crate focus: changes confined to appropriate module
- No cross-cutting regressions or scope drift
- Commit history is linear and purposeful

### Cleanliness
- No debug code (dbg!, println!, eprintln!, todo!)
- No TODOs, FIXMEs, HACKs, or XXX markers
- No commented-out code blocks
- No temporary files or artifacts

### Alignment
Diff directly implements stated feature/fix with minimal side effects. Ready for merge.

---

EffortlessSteven:
## Refactor Plan

### 1. Extract header-end detection helper
- **Where:** `crates/perl-dap/src/framing.rs:84`
- **What:** New `find_header_end()` helper returns `Option<(header_end, header_len)>` instead of just the end position. This allows the caller to handle CRLF/LF line endings correctly when consuming the header block.
- **How:** No change; extraction is already done and cleanly solves the CRLF-handling problem.

### 2. Use .lines() instead of manual .split("\r\n")
- **Where:** Line 88 (estimated)
- **What:** Changed from `.split("\r\n")` to `.lines()`, which correctly handles both LF and CRLF. This is a good fix.
- **How:** Already done; no change needed.

### 3. Consume header block helper accepts new parameter
- **Where:** Lines 89, 99 (estimated)
- **What:** `consume_header_block()` now takes `(header_end, header_len)` instead of just `header_end`. This allows correct offset advancement.
- **How:** Already done; the change is good.

### Skip (not worth changing)
- The refactoring is complete and correct.
- Helper function is well-named and clear.
- Line-ending handling is now robust.

**Recommendation:** Skip green-refactor for this PR. Refactoring is already done well.

PR #5697, opened 2026-04-23
feat(perl-lsp-rs-core): add builtin three-arg open critic policy (#0000)

EffortlessSteven:
### Motivation

- The Rust-side built-in critic fallback only checked `use strict` and `use warnings`, leaving unsafe two-argument `open` calls detectable only by external `perlcritic`.
- Provide safer diagnostics even when external `perlcritic` is not available by surfacing a lightweight Rust analyzer for a common IO hazard.

### Description

- Add a new built-in policy `InputOutput::RequireThreeArgOpen` implemented as `RequireThreeArgOpen` that flags two-argument `open` usages.
- Use a `static LazyLock<Option<Regex>>` to hold the two-argument `open` detection pattern and compute byte/line/column positions for the reported `Violation` range.
- Register the new policy in `BuiltInAnalyzer::default()` so it runs alongside the existing strict/warnings checks.
- Add unit tests verifying that a two-argument `open` produces a violation and that a three-argument `open` does not.

### Testing

- Ran `cargo test -p perl-lsp-rs-core built_in_analyzer`, which passed (2 tests ran and succeeded).
- Ran `cargo check --all-targets -p perl-lsp-rs-core`, which completed successfully.
- Ran `cargo clippy -p perl-lsp-rs-core --lib -- -D warnings`, which completed successfully for the library target.
- Noted that `cargo clippy --all-targets -p perl-lsp-rs-core` and `cargo xtask fmt` reported unrelated failures in other test/xtask targets (pre-existing, not caused by these changes).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea3744e8188333afaa627edab0c16e)

### DIFF
--- a/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
+++ b/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
@@ -1,5 +1,10 @@
-use super::{QuickFix, Severity, Violation, built_in_quick_fix, insertion_range};
+use super::{built_in_quick_fix, insertion_range, QuickFix, Severity, Violation};
 use perl_parser_core::Node;
+use regex::Regex;
+use std::sync::LazyLock;
+
+static TWO_ARG_OPEN_PATTERN: LazyLock<Option<Regex>> =
+    LazyLock::new(|| Regex::new(r"(?m)\bopen\s*(?:\(|\s+)\s*[^,\n]+,\s*[^,\n\)]+\)?\s*;").ok());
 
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Duplicate of #5711. Both add RequireThreeArgOpen and related open policies to crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs; #5711 is more comprehensive (+213 vs +97, adds ProhibitTwoArgOpen, ProhibitBarewordFileHandles). Keeping #5711.

PR #5762, opened 2026-04-24
fix(perl-module): support legacy ASCII separators in parent/base references (#0000)

EffortlessSteven:
### Motivation

- `find_module_reference_extended` only scanned canonical `::` separators in `use parent`/`use base` argument lists, so legacy ASCII tick separators (`'`) were not recognized as module references.
- This made cursor-based module extraction and subsequent canonicalization fail for legacy-style module names inside `parent`/`base` `qw(...)` lists.

### Description

- Replaced the canonical-only scanner with `scan_module_token` in `crates/perl-module/src/reference/mod.rs` to accept both `::` and legacy ASCII tick (`'`) separators when scanning argument lists.
- Treated tokens containing a `'` as module-like by updating the `is_module_like` check so legacy separators are recognized for extraction.
- Added an integration test `extended_extracts_legacy_ascii_separator_module_from_parent_qw` in `crates/perl-module/tests/module_reference_integration.rs` that verifies extraction and canonicalization of legacy `'`-separated modules inside `use parent qw(...)`.

### Testing

- `cargo test -p perl-module` completed successfully with all tests passing.
- `cargo test -p perl-module --test module_reference_integration` completed successfully and the new integration test passed.
- `cargo check --all-targets -p perl-module` completed successfully.
- `cargo clippy -p perl-module --all-targets` completed (reports only pre-existing warnings unrelated to this change) and `cargo xtask fmt` completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaa079b3e483339958abeb446b6c2a)

### DIFF
--- a/crates/perl-module/src/reference/mod.rs
+++ b/crates/perl-module/src/reference/mod.rs
@@ -147,12 +147,13 @@ fn find_parent_base_module_in_line<'a>(
         }
 
         let token_start_in_args = i;
-        let token_end_in_args = scan_canonical_module_token(bytes, i);
+        let token_end_in_args = scan_module_token(bytes, i);
         let token_start_in_line = args_start_in_line + token_start_in_args;
         let token_end_in_line = args_start_in_line + token_end_in_args;
         let module_name = &args_area[token_start_in_args..token_end_in_args];
 
         let is_module_like = module_name.contains("::")
+            || module_name.contains('\'')
             || module_name.as_bytes().first().is_some_and(u8::is_ascii_uppercase);
 
         if is_module_like
@@ -173,7 +174,7 @@ fn find_parent_base_module_in_line<'a>(
     None
 }
 
-fn scan_canonical_module_token(bytes: &[u8], start: usize) -> usize {
+fn scan_module_token(bytes: &[u8], start: usize) -> usize {
     let mut i = start;
 
     loop {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
**Codex review**: Duplicate of #5763 (both fix ASCII separator handling in parent/base scanning on same files). Recommend closing #5762 in favor of #5763 after comparing diffs for code quality.

EffortlessSteven:
Closing as duplicate of #5763. Both target perl-module parent/base scanning for ASCII legacy separators. #5763 is the clean diff.

EffortlessSteven:
deep-review: APPROVED with no changes needed.

**Verified:** The apostrophe separator scanning in `scan_module_token` is correct:
- Requires `is_module_start_byte` after `'`` (alpha or `_`), so `Foo'123` and `Foo''Bar` are not scanned as modules
- `i += 1` for `'`` (1-char separator) vs `i += 2` for `::` (2-char) is consistent
- The `is_module_like` check adds `module_name.contains('\'\')` so tokens without uppercase first char are still recognized if they use apostrophe separators
- `canonical_module_name()` delegates to `normalize_package_separator` which replaces `'`` → `::`, so `Foo'Bar` → `Foo::Bar` correctly

**PR 5763 closed** as a subset — it adds the apostrophe scanning but not the function rename or the `is_module_like` update.

**Edge case verified:** Mixed form like `Foo'::Bar` stops scanning at `Foo` (bytes after `'`` is `:`, not alpha), avoiding false module detection.

PR #5764, opened 2026-04-24
fix(perl-pragma): normalize GB18030 encoding aliases (#0000)

EffortlessSteven:
### Motivation
- Canonicalize values passed to `use encoding` so common aliases like `GB-18030` and `GB_18030` are recorded consistently and downstream logic can rely on a normalized encoding name.

### Description
- Add `normalized_encoding_name` to canonicalize encoding names by lowercasing and stripping `-`/`_` separators and return `Option<String>` when the token is empty.
- Use `normalized_encoding_name` in both conditional (`use if ..., encoding, ...`) and direct (`use encoding ...`) pragma handling paths so the recorded `PragmaState.encoding` is normalized.
- Add unit tests `use_encoding_normalizes_gb18030_aliases` and `use_if_encoding_normalizes_gb18030_aliases` to verify normalization for GB18030 aliases.

### Testing
- Ran `cargo test -p perl-pragma` and all tests passed.
- Ran `cargo xtask fmt` and `cargo clippy -p perl-pragma --all-targets -- -D warnings` which completed without failures, and noted `just pr-fast` was not available in this environment.

------

### DIFF
--- a/crates/perl-pragma/src/lib.rs
+++ b/crates/perl-pragma/src/lib.rs
@@ -370,6 +370,15 @@ fn normalized_pragma_token(arg: &str) -> &str {
     arg.trim().trim_matches('\'').trim_matches('"')
 }
 
+fn normalized_encoding_name(arg: &str) -> Option<String> {
+    let encoding = normalized_pragma_token(arg);
+    if encoding.is_empty() {
+        return None;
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #5765, opened 2026-04-24
fix(perl-parser): decode GB18030 source files (#0000)

EffortlessSteven:
### Motivation

- `perl-parse` previously only attempted UTF-8 and otherwise fell back to byte-preserving Latin-1 semantics, so Perl files declaring `use encoding 'gb18030'` were mis-decoded.

### Description

- Add the `encoding_rs` workspace dependency and enable it for the `perl-parser` crate by updating `Cargo.toml` and the lockfile.
- Inspect the first ~4KB of a source buffer for a `use encoding '...'` pragma and, for `gb18030` / `gb-18030` aliases, decode the full buffer with `encoding_rs::GB18030` before falling back to byte-preserving decoding via `read_source_bytes`.
- Implement helper functions `detect_declared_source_encoding`, `parse_encoding_pragma`, and `is_gb18030_alias`, and wire the detection into `read_source_bytes` in `crates/perl-parser/src/bin/perl-parse.rs`.
- Add unit tests covering declared GB18030 decoding and alias recognition in the `perl-parse` binary tests.

### Testing

- Ran `cargo test -p perl-parser --features cli --bin perl-parse` and the added tests passed.
- Ran `cargo check --all-targets -p perl-parser --features cli` and it completed successfully.
- Ran `cargo clippy -p perl-parser --features cli --bin perl-parse -- -D warnings` with no warnings.
- Ran `cargo xtask fmt` to format changes; `just pr-fast` was not available in this environment (`just: command not found`).

------

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -638,6 +638,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+dependencies = [
+ "cfg-if",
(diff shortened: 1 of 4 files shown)


EffortlessSteven:

PR #5767, opened 2026-04-24
feat(perl-parser): improve GB18030 source decoding (#0000)

EffortlessSteven:
### Motivation
- `perl-parse` fell back to a byte-for-byte single-byte mapping when UTF-8 decoding failed, which produced incorrect text for GB18030-encoded Perl sources.

### Description
- Added `encoding_rs` to `crates/perl-parser` and try a `GB18030.decode` pass when UTF-8 decoding fails in `read_source_bytes` to properly decode GB18030 input.
- Retained the existing byte-preserving fallback when GB18030 decoding reports errors to preserve prior behavior for unknown/malformed encodings.
- Added a unit test `read_source_bytes_decodes_gb18030` and the corresponding lockfile entry to validate the change.

### Testing
- `cargo test -p perl-parser read_source_bytes_ -- --nocapture` passed.
- `cargo clippy -p perl-parser --all-targets -- -D warnings` and `cargo xtask fmt` passed.
- `just pr-fast` could not be executed in this environment (`just: command not found`).

------

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -638,6 +638,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #5768, opened 2026-04-24
fix(workspace): decode Latin-1 files during indexing (#0000)

EffortlessSteven:
### Motivation
- Workspace indexing and on-disk reads previously relied on UTF-8-only APIs which caused Perl source files containing Latin-1 bytes to be decoded incorrectly or skipped during indexing.

### Description
- Change workspace file reads to use byte reads (`std::fs::read`) and added `decode_workspace_source` which tries `String::from_utf8` and falls back to a byte-preserving Latin-1 mapping when UTF-8 decoding fails.
- Apply the same byte-read + decode fallback to `read_workspace_text` so on-disk document lookup uses the same behavior.
- Add a regression unit test `decode_workspace_source_falls_back_to_latin1` that verifies `0xE5` decodes to `å` under the fallback.
- All changes are localized to `crates/perl-lsp-rs/src/runtime/workspace.rs` and touch only the workspace read path.

### Testing
- Ran `cargo check -p perl-lsp-rs` which completed successfully and `cargo clippy -p perl-lsp-rs` which also passed.
- Attempted `cargo test -p perl-lsp-rs decode_workspace_source_falls_back_to_latin1` but test execution failed due to pre-existing, unrelated compile errors in `runtime/dispatch/lifecycle.rs` (generic `Result` signature issues) preventing the test binary from building.
- `cargo check --all-targets -p perl-lsp-rs` also fails for the same unrelated compile errors, and `just pr-fast` was unavailable in this environment (`just: command not found`).

------

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/workspace.rs
+++ b/crates/perl-lsp-rs/src/runtime/workspace.rs
@@ -1620,8 +1620,8 @@ impl LspServer {
                     break;
                 }
 
-                let content = match std::fs::read_to_string(&path) {
-                    Ok(c) => c,
+                let content = match std::fs::read(&path) {
+                    Ok(bytes) => decode_workspace_source(bytes),
                     Err(e) => {
                         if is_permission_denied_error(&e) {
                             // ONE-TIME window/showMessage (AtomicBool guard)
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5769, opened 2026-04-24
fix(workspace): add latin-1 fallback decoding for file reads (#0000)

EffortlessSteven:
### Motivation
- Workspace indexing and file-watcher refresh used UTF-8-only file reads, so Perl sources encoded as Latin-1 could be skipped or mis-decoded.

### Description
- Add `decode_workspace_file_text(bytes: &[u8])` which returns the UTF-8 string when valid and otherwise decodes bytes as a byte-for-byte Latin-1 fallback.
- Add `read_workspace_file_text(path: &Path)` which reads raw bytes and decodes via the new helper.
- Replace `std::fs::read_to_string` call sites in `crates/perl-lsp-rs/src/runtime/workspace.rs` (indexing, file-watcher, rename/create handling and `read_workspace_text`) to use `read_workspace_file_text` so Latin-1 files are ingested as text.
- Add unit tests `decode_workspace_file_text_supports_latin1_bytes` and `decode_workspace_file_text_preserves_utf8` exercising Latin-1 fallback and UTF-8 preservation.

### Testing
- `cargo xtask fmt` ran and passed.
- `cargo test -p perl-lsp-rs decode_workspace_file_text -- --nocapture` attempted but the build failed due to pre-existing generic `Result<()>` compile errors in `crates/perl-lsp-rs/src/runtime/dispatch/lifecycle.rs` unrelated to this change.
- `cargo check --all-targets -p perl-lsp-rs` and `cargo clippy -p perl-lsp-rs --all-targets -- -D warnings` were run and reported the same pre-existing build errors.
- `just pr-fast` was not available in the environment (`just: command not found`).

------

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/workspace.rs
+++ b/crates/perl-lsp-rs/src/runtime/workspace.rs
@@ -50,6 +50,19 @@ fn should_skip_dir(entry: &walkdir::DirEntry) -> bool {
 }
 
 #[cfg(feature = "workspace")]
+fn decode_workspace_file_text(bytes: &[u8]) -> String {
+    match std::str::from_utf8(bytes) {
+        Ok(text) => text.to_owned(),
+        Err(_) => bytes.iter().map(|byte| char::from(*byte)).collect(),
+    }
+}
+
+#[cfg(feature = "workspace")]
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5770, opened 2026-04-24
fix(runtime): decode Latin-1 workspace files (#0000)

EffortlessSteven:
### Motivation
- Workspace indexing and navigation used UTF-8-only file reads which caused legacy Latin-1 Perl sources to be mis-decoded or skipped, reducing feature coverage for those files.

### Description
- Add `crates/perl-lsp-rs/src/runtime/source_decoding.rs` providing `decode_source_bytes` and `read_source_file` which attempt UTF-8 decoding and fall back to a byte-preserving Latin-1 mapping.
- Wire the new `source_decoding` module into the runtime and replace UTF-8-only filesystem reads with `read_source_file` in `workspace.rs` and `navigation.rs` so indexing, re-indexing, rename handling, and navigation fallbacks work for Latin-1 files.
- Add unit tests in `source_decoding.rs` that verify UTF-8 is preserved and Latin-1 bytes decode losslessly.

### Testing
- `cargo fmt --all` ran and succeeded.
- `cargo check -p perl-lsp-rs --lib` ran and succeeded.
- `cargo clippy -p perl-lsp-rs --lib -- -D warnings` ran and succeeded.

------

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/language/navigation.rs
+++ b/crates/perl-lsp-rs/src/runtime/language/navigation.rs
@@ -6,6 +6,8 @@
 use super::super::*;
 use crate::cancellation::RequestCleanupGuard;
 use crate::protocol::{req_position, req_uri};
+#[cfg(feature = "workspace")]
+use crate::runtime::source_decoding::read_source_file;
 use crate::util::token_under_cursor;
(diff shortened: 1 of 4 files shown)


EffortlessSteven:

PR #5772, opened 2026-04-24
fix(parser): improve Windows-1252 source fallback decoding (#0000)

EffortlessSteven:
### Motivation

- Non-UTF-8 source bytes were being converted with a raw Latin-1 style cast which mis-decoded common Windows-1252 punctuation bytes, causing incorrect characters for Windows-authored files.

### Description

- Change `read_source_bytes` in `crates/perl-parser/src/bin/perl-parse.rs` to use a `decode_windows_1252` fallback for non-UTF-8 bytes instead of `char::from(byte)`.
- Add a `decode_windows_1252` helper that maps the `0x80..=0x9F` extended range to the correct Windows-1252 Unicode code points.
- Add a focused unit test `read_source_bytes_decodes_windows_1252_punctuation` that verifies curly quotes and en dash decode as expected.

### Testing

- `cargo test -p perl-parser --bin perl-parse` failed because the binary requires the `cli` feature (expected precondition), so tests were re-run with the feature enabled.
- `cargo test -p perl-parser --bin perl-parse --features cli` ran the module tests and passed (3 passed, 0 failed).
- `cargo check --all-targets -p perl-parser --features cli` passed.
- `cargo clippy -p perl-parser --features cli --bin perl-parse -- -D warnings` and `cargo xtask fmt` both completed successfully.

------

### DIFF
--- a/crates/perl-parser/src/bin/perl-parse.rs
+++ b/crates/perl-parser/src/bin/perl-parse.rs
@@ -306,10 +306,27 @@ fn read_source_bytes(bytes: Vec<u8>) -> io::Result<String> {
             let raw = err.into_bytes();
             let mut decoded = String::with_capacity(raw.len());
             for byte in raw {
-                decoded.push(char::from(byte));
+                decoded.push(decode_windows_1252(byte));
             }
             Ok(decoded)
         }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5773, opened 2026-04-24
fix(workspace): add Windows-1252 text fallback (#0000)

EffortlessSteven:
### Motivation

- Workspace file reads currently assume UTF-8 and treat CP1252/Windows-1252 files as unreadable, preventing indexing and document updates for files saved with that encoding. 

### Description

- Added a UTF-8-first file reader with a Windows-1252 fallback implemented as `decode_windows_1252(bytes: &[u8])` and `read_text_with_encoding_fallback(path: &Path) -> Result<String, io::Error>` in `crates/perl-lsp-rs/src/runtime/workspace.rs`. 
- Wired the fallback reader into all workspace file-read paths including watcher reindex (`process_file_watcher_uri_immediate`), open-document refresh, rename/create flows, initial workspace scan, and `read_workspace_text`. 
- Added focused unit tests for the decoder in the same module covering smart quotes + euro and Latin-1-range bytes. 
- Kept scope minimal to only the workspace read/fallback logic and tests in `crates/perl-lsp-rs/src/runtime/workspace.rs`. 

### Testing

- `cargo check -p perl-lsp-rs --lib` completed successfully in this environment. 
- `cargo check --all-targets -p perl-lsp-rs` failed due to pre-existing test-target compile errors in unrelated `runtime/dispatch/lifecycle.rs`. 
- `cargo clippy -p perl-lsp-rs --all-targets` reported failures caused by existing clippy denies and unrelated test `panic!` usage; these are pre-existing and not introduced by this change. 
- Running `cargo test -p perl-lsp-rs --test decode_windows_1252` could not complete because the workspace has unrelated compile errors in other test targets; the new decoder unit tests are small and local to the modified module. 

Problem: Workspace file reads currently require UTF-8 and silently fail to index/update Perl files saved as Windows-1252.
Fix: Added a UTF-8-first file reader with Windows-1252 fallback and wired it into workspace read paths, plus focused decode tests.
Verification: `cargo check -p perl-lsp-rs --lib` passes; full-target checks and clippy currently surface pre-existing unrelated failures.

------

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/workspace.rs
+++ b/crates/perl-lsp-rs/src/runtime/workspace.rs
@@ -10,7 +10,7 @@
 
 use super::*;
 #[cfg(feature = "workspace")]
-use crate::runtime::routing::{IndexAccessMode, route_index_access};
+use crate::runtime::routing::{route_index_access, IndexAccessMode};
 use crate::state::workspace_symbol_cap;
 use perl_module::path::file_path_to_module_name;
 use perl_module::rename::{apply_module_rename_edits, plan_module_rename_edits};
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5774, opened 2026-04-24
fix(perl-parser): decode Windows-1252 source bytes (#0000)

EffortlessSteven:
### Motivation
- Non-UTF-8 source fallback previously mapped bytes with `char::from(byte)` (ISO-8859-1), which misinterprets Windows-1252 specific punctuation such as smart quotes, em dash, and the euro sign.

### Description
- Replace the byte-to-char fallback in `read_source_bytes` with an explicit `decode_windows_1252_byte` mapping for CP1252 control range bytes while preserving Latin-1 behavior for other bytes.
- Add a focused unit test `read_source_bytes_decodes_windows_1252_punctuation` that verifies decoding of `“Hi”—€` via the same code path.
- Modified file: `crates/perl-parser/src/bin/perl-parse.rs`.

### Testing
- Ran `cargo test -p perl-parser --features cli --bin perl-parse read_source_bytes` and the targeted tests passed (`3 passed`).
- Ran `cargo check --all-targets -p perl-parser --features cli` and it completed successfully.

------

### DIFF
--- a/crates/perl-parser/src/bin/perl-parse.rs
+++ b/crates/perl-parser/src/bin/perl-parse.rs
@@ -306,10 +306,43 @@ fn read_source_bytes(bytes: Vec<u8>) -> io::Result<String> {
             let raw = err.into_bytes();
             let mut decoded = String::with_capacity(raw.len());
             for byte in raw {
-                decoded.push(char::from(byte));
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5775, opened 2026-04-24
fix(perl-pragma): normalize Windows-1252 encoding aliases (#0000)

EffortlessSteven:
### Motivation

- Source encoding values from `use encoding` were recorded verbatim which caused equivalent Windows-1252 spellings (`cp1252`, `cp-1252`, `windows1252`, etc.) to be treated as different encodings.

### Description

- Add `normalize_source_encoding` to canonicalize `use encoding` argument tokens and map common Windows-1252 aliases to the canonical string `windows-1252`.
- Use the normalization helper for both conditional (`use if/unless ... encoding ...`) and direct (`use encoding ...`) pragma handling so the tracked `PragmaState.encoding` is consistent.
- Add unit tests `use_if_encoding_normalizes_windows_1252_aliases` and `use_encoding_normalizes_windows_1252_aliases` to cover alias normalization in conditional and direct forms.

### Testing

- Ran `cargo test -p perl-pragma` and all tests passed (`103` tests in `comprehensive_unit_tests` plus behavior tests succeeded).
- Ran `cargo clippy -p perl-pragma --all-targets -- -D warnings` which completed with no warnings.
- Ran `cargo xtask fmt` which completed successfully in this environment, while `just pr-fast` could not be run because `just` is not installed here.

------

### DIFF
--- a/crates/perl-pragma/src/lib.rs
+++ b/crates/perl-pragma/src/lib.rs
@@ -370,6 +370,21 @@ fn normalized_pragma_token(arg: &str) -> &str {
     arg.trim().trim_matches('\'').trim_matches('"')
 }
 
+fn normalize_source_encoding(arg: &str) -> Option<String> {
+    let normalized = normalized_pragma_token(arg);
+    if normalized.is_empty() {
+        return None;
+    }
+
+    let lower = normalized.to_ascii_lowercase();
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #5777, opened 2026-04-24
fix(code-actions): respect UTF-7 encoding pragmas (#0000)

EffortlessSteven:
### Motivation

- The enhanced code action provider suggested adding UTF-8 pragmas even when the source already declared an encoding such as `use encoding 'UTF-7';`.

### Description

- Stop offering the "Add UTF-8 support" quick fix when a source encoding declaration exists by introducing `has_source_encoding_declaration` and using it in the decision logic.
- Recognize both `use utf8` and `use encoding '...'`/`use encoding "..."` forms (covers `UTF-7`).
- Add unit tests that assert `use encoding 'UTF-7';` is detected and that the UTF-8 quick fix is suppressed in that case.
- Affected file: `crates/perl-lsp-rs-core/src/providers/code_actions/enhanced/mod.rs`.

### Testing

- Ran `cargo test -p perl-lsp-rs-core test_utf8_code_action_skipped_when_encoding_pragma_exists` and the test passed.
- Ran `cargo test -p perl-lsp-rs-core test_source_encoding_declaration_utf7_is_recognized` and the test passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaa0a0572883339fd7424729c22ea0)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/enhanced/mod.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/enhanced/mod.rs
@@ -407,8 +407,8 @@ impl EnhancedCodeActionsProvider {
             });
         }
 
-        // Add utf8 support if missing
-        if !self.source.contains("use utf8") && helpers.has_non_ascii_content() {
+        // Add utf8 support if no source encoding declaration exists.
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

Issue #4246, opened 2026-04-11
[umbrella] import/symbol visibility parity — #3472 + #3475 + #3476


EffortlessSteven:
## Overview

Issues #3472, #3475, and #3476 are one coordinated engineering lane, not three separate fixes. They all concern the same question: *which symbols become visible in a consuming file's scope when it uses a `use` or `require` statement?*

Tracking them separately causes duplicated analysis work and makes the shared abstraction invisible. This umbrella makes the lane explicit.

## Shared abstraction

The unifying concept is an **`ImportSurface` / `VisibleImports` layer** — a per-file record of which bare names were imported from which module, with enough source location information for goto-definition to jump to the definition site.

**Produced from:**
- `use Module qw(sym1 sym2);` / `use Module ('sym1');` / `use Module 'sym1';` — qw/parens/single-quote list forms (#3472)
- `use constant { NAME => value };` / `use constant NAME => value;` — constant namespace entries (#3475)
- `require Module; Module->import('sym');` with static names — literal require + explicit named import (#3476)

**Consumed by:**
- goto-definition on bareword function calls
- Completion candidate list for bareword identifiers
- `undefined-symbol` diagnostics (must not fire on resolved imports)
- Workspace index visibility (cross-file symbol queries)

## Shipping order

The issues form a dependency chain from simplest to most complex:

1. **#3472** — `use Module qw(...)` / parens-list / tag imports (closed — landed in PRs #3778, #3828, #4133)
2. **#3475** — `use constant { ... }` hash-ref and scalar forms (closed — landed in PR #4133)
3. **#3476** — `require Module; Module->import('sym')` with *static names only* (open — literal-only slice targeted for v0.13.0)

## Explicit scope boundary

The following are **out of scope** for this lane and should be filed as separate issues if needed:

- `Module::Runtime`-style dynamic module name construction
- String-built module names (`my $m = "Foo"; $m->import(...)`)
- Framework-specific magic imports (`Moose`, `Moo` import side effects beyond what `@ISA` already handles)

## Acceptance criteria

- `use Carp qw(croak); croak();` — goto-def on `croak` resolves to `Carp::croak` (done via #3472)
- `use constant { PI => 3.14 }; print PI;` — goto-def on `PI` resolves to the `use constant` line (done via #3475)
- `require Foo; Foo->import('bar'); bar();` — goto-def on `bar` resolves to `Foo::bar` when `Foo` is a static string literal (tracked in #3476)

[...]

PR #5788, opened 2026-04-24
refactor(semantic): introduce ImportSurface / VisibleImports record (#4246)

EffortlessSteven:
### Motivation

- Imported bareword visibility and import-origin logic were scattered across ad-hoc checks in scope and declaration code, making reasoning and future extension harder.

### Description

- Add a new `ImportSurface` abstraction and `VisibleImport` / `VisibleImportKind` types in `crates/perl-semantic-analyzer/src/analysis/import_surface.rs` to represent per-file visible imported bare names and their origin (bare name, source module/package, kind, origin span, resolved flag).
- Populate `ImportSurface` from static `use` forms already supported: `qw(...)`, parenthesized import lists, single-quoted imports, known export tag expansions, and `use constant` variants (including hash/`qw`/pair forms).
- Thread `ImportSurface` into existing consumers: replace the ad-hoc imported-barewords set in `scope_analyzer` with `ImportSurface::from_ast(...)` and use `has_visible_name`, and consult `ImportSurface` from `declaration::find_import_source` to prefer file-local visible imports before falling back to the existing `require ...; Module->import(...)` logic.
- Re-export the new module via `analysis/mod.rs` and `lib.rs` so other crates can reference `import_surface`.
- Add unit tests exercising `qw`/parenthesized/single-quote/tag/`use constant` forms and extend declaration tests to cover parenthesized/single-quote/tag imports and that `use constant` entries do not incorrectly resolve to the `constant` pragma module.

### Testing

- Ran formatting: `cargo xtask fmt` / `cargo fmt -p perl-semantic-analyzer` (completed).
- Ran unit tests for the changed crate: `cargo test -p perl-semantic-analyzer` — all tests passed.
- Ran checks for the changed crate: `cargo check --all-targets -p perl-semantic-analyzer` — succeeded.
- Full workspace check: `cargo check --workspace --all-targets` surfaced an unrelated pre-existing syntax issue in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` (not caused by this change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead312e7d483339c7ab101ae3f71c6)

PR #5791, opened 2026-04-24
refactor(semantic): introduce ImportSurface / VisibleImports record (#4246)

EffortlessSteven:
### Motivation
- Replace scattered ad-hoc tracking of visible imported bare names with a single per-file abstraction so consumers (diagnostics, completion, declaration lookup) share one consistent view.

### Description
- Add `crates/perl-semantic-analyzer/src/analysis/import_surface.rs` implementing `ImportSurface`, `VisibleImport`, and `VisibleImportKind` with origin span and resolved/candidate flags, and tests covering `qw`/parenthesized/single-quote/tag/`use constant` forms.
- Wire the new module into the public API by registering `analysis::import_surface` and re-exporting it from `perl-semantic-analyzer` crate root.
- Replace the ad-hoc `collect_imported_barewords` usage in scope analysis with `ImportSurface::from_ast(...)` and call sites now query `ImportSurface` (preserving previous semantics for visible names).
- Route declaration/`find_import_source` lookup to consult `ImportSurface` first for static `use`-based visibility while keeping the existing `require` + `Module->import(...)` scan and export-tag resolution to avoid regressions.

### Testing
- Ran `cargo test -p perl-semantic-analyzer` and the crate's unit tests (including the new `import_surface` tests) passed.
- Ran `cargo check --all-targets -p perl-semantic-analyzer` and it succeeded for the crate under change.
- Ran `cargo check --workspace --all-targets` and it surfaced an unrelated pre-existing compile error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` (unterminated string); this failure is not caused by the changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead312e7d483339c7ab101ae3f71c6)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/declaration.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/declaration.rs
@@ -6,6 +6,7 @@
 use crate::ast::{Node, NodeKind};
 use crate::symbol::is_universal_method;
 use crate::workspace_index::{SymKind, SymbolKey};
+use perl_module::import::resolve_known_export_tag;
 use rustc_hash::FxHashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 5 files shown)


EffortlessSteven:
Superseded by #5788 (cluster C18, issue #4246).

Triage: 4 sibling Codex PRs targeted the same task. Comparison:
- **#5788 (keeper)**: threads `&ImportSurface` as a parameter — caller builds the surface once per file then passes it down (efficient façade pattern). Adds dedicated `import_surface_tests.rs` test file and updates `lib.rs` re-export.
- **This PR (#5791)**: more aggressive deletion (removes the legacy `Use`-node scanning block from the `find` helper entirely), but (a) rebuilds `ImportSurface::from_ast(ast)` inside `find_import_source` on every invocation rather than threading, and (b) the aggressive removal of the legacy scanner risks regressions if `ImportSurface` coverage is incomplete for any edge case the old scanner handled. #5788's conservative "add fast-path, keep legacy fallback" approach is safer for a pure refactor.

If #5788 lands and we later want to remove the legacy scanner as a follow-up, that should be a dedicated dead-code removal PR with its own coverage proof.

Closing in favor of #5788.

PR #5792, opened 2026-04-24
feat(semantic): extract per-file Exporter metadata (#3416)

EffortlessSteven:
### Motivation

- Provide per-file metadata for Exporter-style modules so downstream phases can resolve exported names to same-file subroutines. 
- Capture only conservative, statically-readable forms to avoid guessing about dynamic/runtime export construction.

### Description

- Added `ExportedSymbol` and `ExporterMetadata` types and exposed an `exporter_metadata: Option<ExporterMetadata>` field on `ClassModel` to carry per-file Exporter data. 
- Extended `ClassModelBuilder` to collect `@EXPORT`, `@EXPORT_OK`, and `%EXPORT_TAGS` from static `our`/bare assignments and to merge static `%EXPORT_TAGS` hash literals into a tag map. 
- Implemented conservative resolution that maps exported names to same-package subroutine declarations (populating `declaration: Option<SourceLocation>`), leaving unresolved symbols as `None` rather than panicking. 
- Added unit tests exercising `@EXPORT`, `@EXPORT_OK`, `%EXPORT_TAGS`, and unresolved-symbol behavior and small helper functions `collect_export_tag_map` and `merge_export_tags` to support static extraction.

### Testing

- Ran `cargo test -p perl-semantic-analyzer` which passed (existing and new tests in `analysis::class_model` succeeded).
- Ran `cargo check --workspace --all-targets` which currently fails due to a pre-existing, unrelated syntax error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` (unterminated string), not caused by these changes. 
- Ran `cargo xtask fmt` formatting step but it was blocked by the same unrelated compile error above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead317d3e08333a4459344d323b35f)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/class_model.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/class_model.rs
@@ -6,7 +6,7 @@
 
 use crate::SourceLocation;
 use crate::ast::{Node, NodeKind};
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 /// Which OO framework a package uses.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Duplicate of #5795 (Codex ensemble burst, same issue #3416, both touch class_model.rs). Keeping #5795 (313 additions vs 183 here). Note: #5793 is also kept — it implements a separate surface (exporter_metadata.rs in semantic/). Closing as ensemble dupe.

Issue #3415, opened 2026-04-10
Import/Export Gap: Module::Runtime (use_module, require_module) support missing


EffortlessSteven:
# Import/Export Gap: Module::Runtime (use_module, require_module) support missing

## Problem

Module::Runtime provides `use_module()` and `require_module()` for loading modules at runtime with better error handling than bare `require`. The LSP does not recognize these patterns as module imports.

## Current State

- Parsing: ✗ No specific handling for `use_module()` calls
- Semantic analysis: ✗ Runtime module loading not tracked
- Navigation: ✗ Can't jump to modules loaded via Module::Runtime

## Example Patterns Not Supported

```perl
use Module::Runtime qw(use_module require_module);

# Runtime module loading - LSP doesn't track this as an import
my $module = use_module('Some::Module');
$module->new();

# Conditional loading
if ($condition) {
    require_module('Optional::Dependency');
}

# Dynamic module names
my $backend = 'JSON::XS';
use_module($backend);  # Can't resolve statically
```

## Technical Challenge

- Module names are often dynamic (variables, concatenated strings)
- Runtime loading means the module might not always be loaded
- `use_module` returns the module name, creating indirect usage patterns

## Suggested Implementation

1. **Static cases**: When `use_module('Literal::Name')` is called, treat as `use Literal::Name`
2. **Pattern detection**: Recognize common patterns like `use_module($backend)` where $backend is assigned from a constant
3. **Conservative tracking**: Mark files using Module::Runtime as having "dynamic dependencies"

## Files to Modify

- `crates/perl-module-import/src/lib.rs` - Module import parsing
- `crates/perl-semantic-analyzer/src/analysis/symbol.rs` - Symbol tracking

## Priority

Low - Runtime module loading is less common than compile-time `use`, but important for complete analysis.

## Related

- Issue #3409 (Exporter support)
- crates/perl-module-import

EffortlessSteven:
# Deep Research: Module::Runtime Support for perl-lsp (Issue #3415)

## Executive Summary

Module::Runtime provides `use_module()` and `require_module()` for runtime module loading, presenting unique challenges for Language Server Protocol (LSP) implementations. Unlike static `use` and `require` statements, these runtime calls create dynamic dependencies that are difficult to trace statically. This research analyzes prior art from other LSPs and proposes a four-phase implementation plan for perl-lsp.

---

## 1. Prior Art Analysis

### 1.1 Python LSP (pylsp/pyright)

[...]

PR #5808, opened 2026-04-24
feat(module-resolution): support static Module::Runtime imports (#3415)

EffortlessSteven:
### Motivation
- Enable resolving literal `use_module`/`require_module` calls when `Module::Runtime` explicitly imports those helpers while remaining conservative for dynamic or unimported cases.
- Fix incorrect alias detection that treated unqualified helper calls as module-runtime aliases even when `Module::Runtime` wasn't imported.

### Description
- Add `ModuleRuntimeImports` and `module_runtime_imports(stmts: &[Node])` to detect `use Module::Runtime qw(...)` helper imports in a statement list.
- Gate unqualified `use_module`/`require_module` alias detection on the presence of an explicit `Module::Runtime` import while still accepting fully-qualified `Module::Runtime::use_module`/`require_module` calls by updating `module_runtime_alias`.
- Thread the detected `module_runtime_imports` into `scan_statements_for_require_import` so the existing `require`+`->import(...)` resolution path is reused and remains literal-only for module names.
- Add tests in `tests/require_import_resolution.rs` covering static `use_module` and `require_module` with manual `Module::Runtime` import, the conservative behavior when the helper is not imported, and the conservative behavior for dynamic module-name arguments.

### Testing
- Ran `cargo test -p perl-semantic-analyzer` and all tests passed (121 + crate test suites ran; modified tests included and passed).
- Ran `cargo test -p perl-workspace` and the workspace-index tests passed.
- Attempted `cargo check --workspace --all-targets` which failed due to a pre-existing unrelated compile error (`E0765` unterminated string) in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` and not caused by these changes.
- The change was committed with message `feat(module-resolution): support static Module::Runtime imports (#3415)`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead34d1a108333af72b6d8b44f0222)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/declaration.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/declaration.rs
@@ -1544,7 +1544,67 @@ pub fn symbol_at_cursor(ast: &Node, offset: usize, current_pkg: &str) -> Option<
             }
         }
 
-        fn module_runtime_alias(expr: &Node) -> Option<(String, String)> {
+        #[derive(Default)]
+        struct ModuleRuntimeImports {
+            use_module: bool,
+            require_module: bool,
+        }
+
+        impl ModuleRuntimeImports {
+            fn tracks(&self, call_name: &str) -> bool {
+                matches!(
+                    (call_name, self.use_module, self.require_module),
+                    ("use_module", true, _) | ("require_module", _, true)
+                )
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #5810, opened 2026-04-24
feat(module-resolution): support static Module::Runtime imports (#3415)

EffortlessSteven:
### Motivation

- Module::Runtime's `use_module`/`require_module` calls were not recognized through the existing manual-import resolution path, preventing literal static calls from resolving to their target modules.

### Description

- Teach the existing require+import scan to read `use Module::Runtime ...` import lists and collect `use_module`/`require_module` as imported names. 
- Accept unqualified `use_module(...)` / `require_module(...)` only when those names were actually imported and always accept fully-qualified `Module::Runtime::use_module` / `Module::Runtime::require_module` calls. 
- Resolve only literal string module-name arguments and fold literal runtime calls into the same `required_modules`/`aliases` scan instead of adding a parallel resolution path. 
- Add unit tests covering literal `use_module`/`require_module` cases, the qualified `Module::Runtime::use_module` form, the dynamic-name conservative path, and the unqualified-without-import conservative path.

### Testing

- Ran `cargo test -p perl-semantic-analyzer` which passed (all tests OK). 
- Ran `cargo test -p perl-workspace` which passed (all tests OK). 
- Ran `cargo check --workspace --all-targets` which failed due to a pre-existing unrelated compile error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` (E0765: unterminated string) and not due to these changes. 
- Ran `cargo xtask fmt` which could not complete because of the same pre-existing unrelated compile error.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead34d1a108333af72b6d8b44f0222)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/declaration.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/declaration.rs
@@ -1544,7 +1544,71 @@ pub fn symbol_at_cursor(ast: &Node, offset: usize, current_pkg: &str) -> Option<
             }
         }
 
-        fn module_runtime_alias(expr: &Node) -> Option<(String, String)> {
+        fn module_runtime_imports(stmts: &[Node]) -> std::collections::HashSet<String> {
+            let mut imported = std::collections::HashSet::new();
+            for stmt in stmts {
+                let NodeKind::Use { module, args, .. } = &inner_expr(stmt).kind else {
+                    continue;
+                };
+                if module != "Module::Runtime" {
+                    continue;
+                }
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Superseded by #5811 (Codex ensemble dedup). Same issue (#3415) as #5811; #5810 only touches 2 files (declaration.rs + tests) while #5811 also indexes literal Module::Runtime targets in the workspace index. Closing as duplicate; cluster keeper is #5811. See cluster-triage skill.

Issue #3482, opened 2026-04-10
[module-resolution] Parent class methods not resolved for inherited method calls


EffortlessSteven:
## Problem

Perl's `use parent` and `use base` establish inheritance relationships, but perl-lsp does not:
1. Track which methods come from parent classes
2. Resolve bareword method calls on parent classes
3. Enable go-to-definition on inherited methods
4. Suggest parent methods in completion

## Current Behavior

```perl
# Parent.pm
package Parent;
sub inherited_method { "from parent" }
1;

# Child.pm
package Child;
use parent 'Parent';

sub child_method { }

inherited_method();  # CTRL+Click fails — method not found locally
```

Evidence:
- `crates/perl-lsp-type-hierarchy/src/` exists and handles `use parent` for type hierarchy
- But does NOT use that information for method resolution in scope analysis
- `perl-semantic-analyzer/src/analysis/symbol.rs` only tracks local subroutines
- Parent methods are not added to consumer scope's visible symbols

## Why This Matters

1. **Object-oriented Perl code** — Inheritance is fundamental in OO code
2. **Method calls break** — Go-to-def and completion fail for inherited methods
3. **Scope diagnostics wrong** — Marks inherited methods as undefined
4. **Common code pattern** — Any Moose/Moo class or inheritance-based code affected

### Case 1: Direct method call (should work but doesn't)
```perl
package Base;
sub base_method { "base" }
1;

package Child;
use parent 'Base';

base_method();  # ERROR: undefined subroutine
```

### Case 2: Object method (works, but chain-resolution is weak)
```perl
package Base;
sub method { "result" }
1;

package Child;
use parent 'Base';
1;

my  = Child->new();
->method();  # Currently: may not resolve to Base::method properly
```

### Case 3: Multiple inheritance
```perl
package Child;
use parent qw(Base1 Base2);

method_from_base1();  # Should check Base1 then Base2
method_from_base2();  # Should find it
```

### Type Hierarchy Already Exists
`crates/perl-lsp-type-hierarchy/src/` already builds parent relationships:
- Parses `use parent` and `use base` statements
- Finds parent class definitions
- Builds hierarchy for textDocument/typeHierarchy requests

**Gap**: This hierarchy is NOT used for symbol resolution or method discovery.

### Needed: Scope Extension
When analyzing a file with `use parent 'ParentClass'`:
1. Load Parent::Class definition (already done for type hierarchy)
2. Extract all subroutines from Parent::Class
[...]

PR #5811, opened 2026-04-24
feat(module-resolution): support static Module::Runtime imports (#3415)

EffortlessSteven:
### Motivation
- Module::Runtime loader calls (`use_module` / `require_module`) used to be untracked or only partially handled, preventing literal static imports from participating in import-resolution and workspace dependency indexing.
- The goal is to recognize only literal module-name usages for `Module::Runtime` when the loader functions are explicitly imported (or fully-qualified) and keep variable-driven cases conservative.

### Description
- Reused the existing require/import resolution path in `crates/perl-semantic-analyzer/src/analysis/declaration.rs` by adding `extract_module_runtime_imports`, `module_runtime_call_module`, and wiring them into the existing `scan_statements_for_require_import` logic so literal `use_module`/`require_module` calls are treated like `require` + `->import` pairs.
- Added indexing support in `crates/perl-workspace-index/src/workspace/workspace_index.rs` by tracking `Module::Runtime` imports per file and exposing `extract_module_runtime_literal_dependency` / `extract_module_runtime_import_names` so literal runtime loader calls register dependencies via the normal path.
- Added tests: positive literal cases and a negative dynamic-name case in `crates/perl-semantic-analyzer/tests/require_import_resolution.rs`, and workspace-index dependency tests in `crates/perl-workspace-index/tests/comprehensive_unit_tests.rs` to assert literal resolution and conservative handling of dynamic names.
- Kept behavior conservative: variable-driven module names remain unresolved and no new parallel resolution system was introduced.

### Testing
- Ran `cargo test -p perl-semantic-analyzer` and all tests passed.
- Ran `cargo test -p perl-workspace` and all tests passed (workspace crate is `perl-workspace`).
- Ran `cargo check --all-targets -p perl-semantic-analyzer` and `cargo check --all-targets -p perl-workspace` and both succeeded.
- `cargo check --workspace --all-targets` / formatting runs are blocked by a pre-existing syntax error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` unrelated to these changes, so full workspace check/format could not complete in this session.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead34d1a108333af72b6d8b44f0222)

PR #5812, opened 2026-04-24
feat(module-resolution): resolve inherited bareword methods via parent chain (#3482)

EffortlessSteven:
### Motivation

- Strict `subs` mode flagged bareword function calls that were actually inherited from parent packages; the analyzer needs to treat inherited methods as visible symbols to avoid false-positive diagnostics.

### Description

- Add `ClassModelBuilder` usage and compute per-package inherited method names with `collect_inherited_barewords` using the existing class model and recursive parent traversal with memoization and cycle protection.
- Store inherited symbols in `AnalysisContext` as `inherited_barewords_by_package` and expose `has_inherited_bareword` to the analyzer.
- Suppress `UnquotedBareword` diagnostics when `has_inherited_bareword` returns true so inherited bareword calls (via `use parent` / `use base` / `@ISA`) are accepted, respecting child-local shadowing and parent ordering.
- Add tests in `scope_and_symbol_tests.rs` covering single parent, grandparent chain, multiple inheritance ordering, and child override shadowing.

### Testing

- Ran `cargo test -p perl-semantic-analyzer` and all tests passed.
- `cargo test -p perl-lsp-type-hierarchy` did not run because that package is not present in this workspace (expected mapping to `perl-lsp-rs-core` providers in this repo).
- `cargo test -p perl-lsp-rs-core type_hierarchy` and `cargo check --workspace --all-targets` were blocked by a pre-existing syntax error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs`, which is unrelated to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead3504b148333900b4a2c26c1196a)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
@@ -50,6 +50,7 @@
 //! ```
 
 use crate::ast::{Node, NodeKind};
+use crate::class_model::ClassModelBuilder;
 use crate::pragma_tracker::{PragmaState, PragmaTracker};
 use perl_module::import::resolve_known_export_tag;
 use rustc_hash::FxHashMap;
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #5813, opened 2026-04-24
feat(module-resolution): resolve inherited bareword methods via parent chain (#3482)

EffortlessSteven:
### Motivation

- `use parent` / `use base` parent relationships were collected but inherited methods were not exposed to scope analysis, causing `use strict 'subs'` to flag valid inherited bareword method calls as unquoted barewords.

### Description

- Reuse the existing `ClassModelBuilder` to build per-package inherited-method maps and expose them in `AnalysisContext` via `collect_inherited_barewords` and `has_inherited_bareword`.
- Add DFS/C3 ancestor-order helpers (`dfs_ancestor_order`, `c3_ancestor_order`) to walk parent chains and determine inherited methods while preserving child-local shadowing.
- Integrate the inherited-method visibility check into the strict-subs bareword logic so inherited methods are treated like imported barewords during `NodeKind::Identifier` analysis.
- Add tests covering single parent, grandparent chain, multiple inheritance ordering, and child override shadowing to ensure strict diagnostics no longer fire for supported inherited calls.

### Testing

- Ran `cargo test -p perl-semantic-analyzer` and the crate's full test suite: succeeded (all tests passed in that crate).
- Ran `cargo test -p perl-semantic-analyzer --test scope_and_symbol_tests`: succeeded (scope-related tests including new inheritance tests passed).
- Attempted `cargo test -p perl-lsp-type-hierarchy`: failed because package `perl-lsp-type-hierarchy` is not present in this workspace (the type-hierarchy provider is now inside `perl-lsp-rs-core`).
- Ran `cargo check --workspace --all-targets`: failed due to a pre-existing parse error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs`, which is unrelated to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead3504b148333900b4a2c26c1196a)

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/scope_analyzer.rs
@@ -49,11 +49,13 @@
 //! # }
 //! ```
 
+use crate::analysis::class_model::{ClassModel, ClassModelBuilder, MethodResolutionOrder};
 use crate::ast::{Node, NodeKind};
 use crate::pragma_tracker::{PragmaState, PragmaTracker};
 use perl_module::import::resolve_known_export_tag;
 use rustc_hash::FxHashMap;
 use std::cell::{Cell, RefCell};
+use std::collections::{HashMap, HashSet};
 use std::ops::Range;
 use std::rc::Rc;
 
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

PR #5822, opened 2026-04-24
refactor(pragma): expose explicit PragmaState query API (#0000)

EffortlessSteven:
### Motivation
- Provide an explicit, queryable compile-time pragma environment so downstream consumers can ask “what pragma/feature state is in effect at this file position?” and to make save/restore semantics explicit for lexical scopes.

### Description
- Added `CompileTimeEnvironment` with `from_ast`, `query`, `state_at`, and `ranges` to expose pragma state by file offset.
- Introduced `PragmaStateQuery` (with `strict_enabled`, `warnings_enabled`, `has_feature`) and `PragmaSnapshot` plus `PragmaState::snapshot()` / `PragmaSnapshot::restore_into()` to make save/restore operations explicit and immutable.
- Updated `PragmaTracker` to use the new snapshot/restore API when entering and leaving lexical scopes (`build_scoped_body` and `Block` handling).
- Added behavior-spec tests that exercise the new API and acceptance scenarios: inner `no strict` does not leak, `eval { ... }` restores outer state, and `eval` with a string argument is treated conservatively.

### Testing
- Ran `cargo test -p perl-pragma` and the crate tests passed (all new and existing tests succeeded).
- Ran `cargo check --workspace --all-targets` which failed due to a pre-existing, unrelated syntax error in `crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs` (unterminated string); this failure is not caused by this change.
- Ran per-crate formatting (`cargo fmt -p perl-pragma`) successfully; the workspace `cargo xtask fmt` / full-format step was impeded by the same unrelated parse error.

Problem: Compile-time pragma tracking was only exposed as ad hoc range lookups instead of an explicit queryable environment API.
Fix: Added `CompileTimeEnvironment` + `PragmaStateQuery` API and explicit `PragmaSnapshot` save/restore semantics, and updated tracker usage and tests.
Verification: `cargo test -p perl-pragma` passes; `cargo check --workspace --all-targets` currently fails due to an unrelated pre-existing syntax error.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead35a5b3883339078b9b2a0a697e1)

### DIFF
--- a/crates/perl-pragma/src/lib.rs
+++ b/crates/perl-pragma/src/lib.rs
@@ -110,6 +110,100 @@ impl PragmaState {
     pub fn has_builtin_import(&self, name: &str) -> bool {
         self.builtin_imports.iter().any(|import| import == name)
     }
+
+    /// Creates an immutable snapshot of this compile-time pragma state.
+    #[must_use]
+    pub fn snapshot(&self) -> PragmaSnapshot {
+        PragmaSnapshot { state: self.clone() }
+    }
+}
+
+/// Immutable saved pragma state for explicit save/restore operations.
+#[derive(Debug, Clone, PartialEq)]
+pub struct PragmaSnapshot {
+    state: PragmaState,
+}
+
+impl PragmaSnapshot {
+    /// Returns the captured pragma state.
+    #[must_use]
+    pub fn state(&self) -> &PragmaState {
+        &self.state
+    }
+
+    /// Restores the captured state into a mutable target.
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as Codex multi-copy artifact from generation burst. Superseded by #5823. Thanks for the contribution — the selected variant covers the same ground with cleaner implementation per haiku triage.

PR #5848, opened 2026-04-24
fix(incremental-document): harden batch edits for bounds and UTF-8 boundaries (#0000)

EffortlessSteven:
### Motivation
- Batch application of incremental edits could perform unsafe slicing when edits were out-of-bounds, inverted, or not aligned to UTF-8 char boundaries, risking panics or partial source corruption.
- Prefer a conservative fallback (full parse) over performing partial mutations when any edit in a batch is unmappable or malformed.

### Description
- Validate each edit in `apply_edits()` before mutating `source` and collect affected ranges only for successfully-applied edits, otherwise fall back to a full parse of the unchanged source; preserves reverse-order semantics. (`crates/perl-parser/src/incremental/incremental_document.rs`)
- Change `apply_edit_in_place()` to return `Option<(usize, usize)>` and enforce strict checks for bounds, inverted ranges (`start > end`), and UTF-8 char-boundaries before calling `replace_range`.
- When a batch contains any unmappable/malformed edit, abort the batch mutation and rebuild the document from the original `source` (conservative fallback) to avoid partial corruption.
- Add regression tests (inline in `incremental_document.rs`) that exercise out-of-bounds, inverted-range, mid-codepoint UTF-8 edits, mixed batches with one unmappable edit, and a supported-batch vs fresh-parse equivalence case.

### Testing
- Ran `cargo xtask fmt` / `rustfmt` on the edited file successfully.
- `cargo check --all-targets -p perl-parser` completed successfully.
- `cargo test -p perl-parser` was executed; the test run exposed a single unrelated existing failure (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`), so the suite did not fully pass end-to-end.
- The newly-added incremental batch regression tests were run individually and passed (the targeted tests for the new behavior succeed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eadde20d288333b61dcbd07e72df84)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_document.rs
+++ b/crates/perl-parser/src/incremental/incremental_document.rs
@@ -136,13 +136,23 @@ impl IncrementalDocument {
         // Reverse ordering keeps byte offsets stable while avoiding repeated
         // full-string allocations for each edit.
         let mut new_source = self.source.clone();
+        let mut affected_ranges = Vec::with_capacity(sorted_edits.len());
         for edit in &sorted_edits {
-            self.apply_edit_in_place(&mut new_source, edit);
+            if let Some((start, end)) = self.apply_edit_in_place(&mut new_source, edit) {
+                affected_ranges.push((start, end));
+            } else {
+                debug!(
+                    "Falling back to full parse for edit batch due to invalid edit: start={}, end={}",
+                    edit.start_byte, edit.old_end_byte
+                );
+                let mut parser = Parser::new(&self.source);
+                let new_root = parser.parse()?;
+                self.root = Arc::new(new_root);
+                self.cache_subtrees();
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Superseded by #5851: sibling Codex PR with same goal (harden batch edits for bounds + UTF-8). #5851 is the more complete implementation — adds a shared validated_edit_range helper used by apply_edit/apply_edits/apply_edit_to_string/apply_edit_in_place, uses saturating_sub on the single-token fast-path to avoid underflow on malformed ranges, and ships 5 named regression tests for out-of-bounds, start>end, mid-codepoint, mixed-batch fallback, and parity with fresh parse. Closing this in favor of consolidating on #5851.

PR #5850, opened 2026-04-24
fix(incremental-document): harden batch edits for bounds and UTF-8 boundaries (#0000)

EffortlessSteven:
### Motivation

- Batch application of incremental edits could slice the source at out-of-bounds or non-char-boundary offsets, risking panics or silent corruption; a conservative approach is required to keep source and AST consistent.

### Description

- Validate each edit with a shared `normalized_edit_range` helper that ensures `start`/`end` are within `source.len()`, `start <= end`, and both are UTF-8 char boundaries. 
- Make `apply_edit_in_place` return a `bool` so callers can detect unmappable edits and preserve reverse-order application semantics for valid edits. 
- If any edit in a batch cannot be applied safely, abort partial mutation and perform a conservative full reparse fallback to keep `source` and `root` in sync. 
- Replace ad-hoc slicing in `apply_edit_to_string` with the same normalized validation and add focused regression tests exercising out-of-bounds ranges, inverted ranges, mid-codepoint edits, unmappable-batch fallbacks, and a supported-batch vs full-parse equality case.

### Testing

- Ran `cargo test -p perl-parser`; the test run exercised the crate but the suite reported a pre-existing unrelated failure (`test_complex_subroutine_signatures` in `enhanced_edge_case_parsing_tests`), so the overall crate test command failed; the incremental-document tests were added and exercise the new behavior. 
- Ran `cargo xtask fmt` to format changes prior to commit.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eadde20d288333b61dcbd07e72df84)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_document.rs
+++ b/crates/perl-parser/src/incremental/incremental_document.rs
@@ -137,7 +137,24 @@ impl IncrementalDocument {
         // full-string allocations for each edit.
         let mut new_source = self.source.clone();
         for edit in &sorted_edits {
-            self.apply_edit_in_place(&mut new_source, edit);
+            if !self.apply_edit_in_place(&mut new_source, edit) {
+                debug!(
+                    "Falling back to full reparse due to unmappable batch edit: start={}, end={}, len={}",
+                    edit.start_byte,
+                    edit.old_end_byte,
+                    self.source.len()
+                );
+
+                // Conservative fallback:
+                // if any edit in a batch cannot be mapped safely, avoid partial
+                // source mutation and keep the existing source/tree in sync.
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Superseded by #5851: sibling Codex PR with same goal (harden batch edits for bounds + UTF-8). #5851 includes the same normalized_edit_range concept as a validated_edit_range helper plus an additional saturating_sub fast-path fix that this PR lacks. #5851's verification log also references the correct unrelated failing test (test_cleanup_respects_retention_count), while this PR's log references a different failing test, suggesting weaker verification. Closing this in favor of consolidating on #5851.

PR #5857, opened 2026-04-24

EffortlessSteven:
### Motivation
- Reduce over-conservative re-lex windows and improve bookkeeping so checkpoint-window incremental reparses can reuse more cached tokens without changing the cache architecture.
- Improve observability of how many bytes are re-lexed (including tail fallbacks) and the size of the checkpoint relex window for diagnostics.

### Description
- Preserve provably-unchanged token slices when invalidating overlapping segments by splitting overlapping segments into left/right retained slices instead of dropping the whole segment (`invalidate_range` in `TokenCache`).
- Make `get_tokens_before` and `get_tokens_from` operate at token granularity across segments that may straddle lookup boundaries to avoid spurious cache misses on prefix/suffix lookups.
- Tighten relex window computation so `relex_end` is never below the edited end, record `last_relex_window_bytes`, and include fallback tail-lex bytes in `bytes_relexed` while keeping the conservative full-reparse fallback when a safe prefix cannot be proven.
- Add `tail_bytes_relexed` and `last_relex_window_bytes` to `IncrementalStats` and surface them in the `Display` output for diagnostics.
- Add a checkpoint-focused test `test_interior_edit_reuses_prefix_and_tracks_relex_bytes` that asserts prefix/suffix reuse signals, byte accounting, window bounds, and AST equivalence with a fresh full parse.

### Testing
- `cargo xtask fmt` ran successfully.
- `cargo test -p perl-parser` was executed but the suite failed on a pre-existing unrelated test (`enhanced_edge_case_parsing_tests::test_complex_subroutine_signatures`); this failure is not caused by the changes in this file.
- Focused incremental tests (including the new `test_interior_edit_reuses_prefix_and_tracks_relex_bytes`) were exercised and the targeted checks succeeded locally.

------

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -215,7 +215,29 @@ impl TokenCache {
     /// * `start` - Start byte position of the range to invalidate.
     /// * `end` - End byte position of the range to invalidate.
     fn invalidate_range(&mut self, start: usize, end: usize) {
-        self.segments.retain(|seg| !seg.overlaps(start, end));
+        let mut updated = Vec::new();
+
+        for segment in self.segments.drain(..) {
+            if !segment.overlaps(start, end) {
+                updated.push(segment);
+                continue;
+            }
+
+            let left_tokens: Vec<Token> =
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5858, opened 2026-04-24

EffortlessSteven:
### Motivation

- Reduce over-reporting and stale stats for checkpoint-window reparsing by tightening the computed relex window and improving byte accounting.
- Preserve the existing conservative fallback behavior for interior edits while eliminating obviously-empty-tail fallback counting.

### Description

- Add `relex_window_bytes` to `IncrementalStats` and include it in the `Display` output to surface requested checkpoint-window width.
- Clamp and normalize `relex_start` / `relex_end` so bounds stay within `source.len()`, ensure `relex_end >= edit_end`, and update `left_checkpoint_distance` / `right_checkpoint_distance` using saturating arithmetic to avoid stale values.
- Count only the in-window bytes during Phase 2 re-lexing (`bytes_in_window = min(token_end, relex_end) - token_start`) and accumulate those to `bytes_relexed`.
- In the tail fallback path, add real tail bytes to `bytes_relexed` and only increment `full_tail_fallbacks` when there is an actual tail region beyond `relex_end` (skip empty-tail fallbacks at EOF).
- Add a focused regression test `test_relex_window_metrics_for_eof_edit_do_not_report_tail_fallback` that verifies EOF-near edits produce window metrics and do not falsely report a tail fallback.

### Testing

- Ran `cargo test -p perl-parser`; the test run executed and produced 132 passed, 1 failed, where the failure is an existing unrelated test `refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count` and not related to the checkpoint-window changes, so the new checkpoint-focused test and related logic were exercised as part of the suite.

------

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -347,6 +347,8 @@ pub struct IncrementalStats {
     pub right_checkpoint_distance: usize,
     /// Total bytes relexed during incremental parsing
     pub bytes_relexed: usize,
+    /// Total byte width of requested checkpoint windows (`relex_end - relex_start`)
+    pub relex_window_bytes: usize,
     /// Count of segments reused before the edit (cache efficiency metric)
     pub segments_reused_before: usize,
     /// Count of segments reused after the edit (cache efficiency metric)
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5860, opened 2026-04-24

EffortlessSteven:
### Motivation

- Checkpoint-window metrics under-reported relexed bytes and the effective relex window bounds were not consistently recorded, making it hard to diagnose overly-wide relex regions.
- The two-sided checkpoint window needs tighter, clamped bounds while preserving the existing conservative fallback when prefix correctness cannot be proven.

### Description

- Clamp and tighten checkpoint window bounds in `reparse_from_checkpoint_two_sided` by bounding `relex_start`/`relex_end` to the source length and ensuring the right bound is at least the post-edit end, while preserving the conservative fallback `relex_start > 0 && cached_before.is_none()`.
- Improve stats in `IncrementalStats` by adding cumulative/max checkpoint-distance fields, last-window start/end, and max relex window, and include them in the `Display` output for diagnostics.
- Correct `bytes_relexed` accounting to record the actual phase-2 relex span and also include bytes from the fallback tail relex path so totals reflect all relex activity.
- Add unit assertions and a new focused test `test_relex_window_and_byte_metrics_are_bounded` in the checkpoint tests to validate window bounds, byte accounting, and cumulative metrics.

### Testing

- Ran the full suite via `cargo test -p perl-parser`; compilation and most tests passed but the run reported one failure in a pre-existing flaky/unrelated test `refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count` (failure is not caused by changes in `incremental_checkpoint.rs`).
- Ran the checkpoint-focused and new tests (including `test_relex_window_and_byte_metrics_are_bounded`) which exercised the updated code paths and passed.
- Ran formatting with `cargo xtask fmt` as part of verification (completed successfully).

------

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -345,8 +345,22 @@ pub struct IncrementalStats {
     pub left_checkpoint_distance: usize,
     /// Distance from edit end to the right checkpoint (in bytes)
     pub right_checkpoint_distance: usize,
+    /// Running sum of left checkpoint distances across incremental parses.
+    pub total_left_checkpoint_distance: usize,
+    /// Running sum of right checkpoint distances across incremental parses.
+    pub total_right_checkpoint_distance: usize,
+    /// Largest observed left checkpoint distance.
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5861, opened 2026-04-24

EffortlessSteven:
### Motivation
- Reduce needless re-lexing and over-counting of relexed bytes for two-sided checkpoint-window reparses while preserving the existing conservative fallback semantics. 
- Make checkpoint-distance stats refresh deterministically on each edit to avoid stale metrics carrying across parses.

### Description
- Use saturating arithmetic to refresh `left_checkpoint_distance` and `right_checkpoint_distance` on every edit via `saturating_sub` instead of conditional assignments. 
- Pre-resolve the suffix cache (`cached_after`) and `segments_after` before phase-2 so the re-lex loop can be bounded when a reusable suffix exists. 
- Bound `bytes_relexed` accounting to the planned checkpoint window by clipping token contribution with `min(relex_end)` and stop re-lexing early when tokens start at or beyond `relex_end`. 
- Add two focused unit tests: `test_checkpoint_distances_refresh_across_repeated_edits` and `test_relexed_bytes_bounded_by_checkpoint_window` to validate distance refresh and that bytes re-lexed do not exceed the computed checkpoint window.

### Testing
- Ran `cargo check --all-targets -p perl-parser`, which passed. 
- Ran `cargo test -p perl-parser`, which completed but surfaced an existing unrelated flaky test (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`) causing a single failure while the rest passed (132 passed; 1 failed). 
- Executed targeted incremental-checkpoint test runs which exercised the new tests and assertions and did not fail locally. 
- Ran `cargo xtask fmt` successfully to ensure formatting compliance.

------

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -591,12 +591,8 @@ impl CheckpointedIncrementalParser {
 
         // Track checkpoint distances for statistics
         let edit_end = edit.start + edit.new_text.len();
-        if edit.start >= relex_start {
-            self.stats.left_checkpoint_distance = edit.start - relex_start;
-        }
-        if relex_end >= edit_end {
-            self.stats.right_checkpoint_distance = relex_end - edit_end;
-        }
+        self.stats.left_checkpoint_distance = edit.start.saturating_sub(relex_start);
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5862, opened 2026-04-24

EffortlessSteven:
### Motivation
- Two-sided checkpoint reparses could carry an unnecessarily wide right relex bound even when no cached suffix existed to reuse, inflating metrics and lex work without improving cache reuse.
- `bytes_relexed` previously omitted bytes consumed by the tail fallback path, and there was no explicit metric for when we tightened the right bound.

### Description
- Tighten the effective `relex_end` in `reparse_from_checkpoint_two_sided` to `edit_end` when a right checkpoint exists but there are no cached suffix segments available, and increment `tightened_relex_windows` in that case.
- Add `tightened_relex_windows` to `IncrementalStats` and include it in the `Display` output for diagnostics.
- Count `bytes_relexed` during the tail fallback lexing path so the metric reflects all bytes re-lexed during incremental reparses.
- Add a focused test `test_interior_edit_tightens_window_without_suffix_cache` that asserts the tightened-window bookkeeping is exercised, the right checkpoint distance reflects the tightened bound, and the incremental AST matches a fresh full parse.

### Testing
- Ran the crate test suite with `cargo test -p perl-parser`, which executed 133 tests where 132 passed and 1 failed; the failing test is unrelated to these changes: `refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`.
- Verified the checkpoint-focused behavior via the new unit test `test_interior_edit_tightens_window_without_suffix_cache`, and ran formatting via `cargo xtask fmt`.

------

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -355,6 +355,8 @@ pub struct IncrementalStats {
     pub segments_invalidated: usize,
     /// Count of times we had to relex the entire tail (cache coverage gaps)
     pub full_tail_fallbacks: usize,
+    /// Count of times the right relex boundary was tightened because no suffix cache could be reused.
+    pub tightened_relex_windows: usize,
 }
 
 impl std::fmt::Display for IncrementalStats {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5863, opened 2026-04-24

EffortlessSteven:
### Motivation

- Reduce overly-wide re-lex windows in the checkpointed incremental parser while preserving the conservative fallback semantics for interior edits.
- Improve accuracy of checkpoint-distance and bytes-relexed reporting so metrics reflect the actual work performed.

### Description

- Tighten right-bound computation in `reparse_from_checkpoint_two_sided` so `relex_end` is never smaller than the post-edit end and normalize checkpoint distances with `saturating_sub` for safe reporting (file: `crates/perl-parser/src/incremental/incremental_checkpoint.rs`).
- Stop phase-2 re-lexing when a token begins at or after the right checkpoint so suffix tokens that start exactly at the right boundary are reused from cache instead of re-lexed, and clamp per-token byte accounting to the active window.
- Ensure bytes relexed from the tail/fallback lex path are included in `bytes_relexed` so stats reflect total re-lex work across both the windowed phase and fallback path.
- Add a focused unit test `test_checkpoint_window_does_not_relex_token_starting_at_right_boundary` in the same file to assert AST equivalence with a full parse and to validate tightened bytes/relex and checkpoint-distance bookkeeping.

### Testing

- Ran the package test suite with `cargo test -p perl-parser`; the run shows the repository has an existing, unrelated failing test (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`), while the incremental-checkpoint-focused tests (including the new test) exercised the changed logic and did not indicate regressions for the checkpoint-window cases.

------

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -586,17 +586,16 @@ impl CheckpointedIncrementalParser {
     ) -> ParseResult<Node> {
         // Calculate relex bounds using checkpoint positions
         let relex_start = left_checkpoint.as_ref().map(|cp| cp.position).unwrap_or(0);
-        let relex_end =
-            right_checkpoint.as_ref().map(|cp| cp.position).unwrap_or(self.source.len());
+        let relex_end = right_checkpoint
+            .as_ref()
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

PR #5883, opened 2026-04-24
fix(perl-module): improve multiline @INC pragma parsing (#0000)

EffortlessSteven:
### Motivation

- The lexical @INC extraction scanned source line-by-line which failed to recognize multiline or parenthesized `use lib` / `no lib` statements. 
- The change aims to extract operations in true lexical order across statement boundaries while preserving existing single-line behavior.

### Description

- Switched `extract_use_lib_paths` and `extract_use_lib_operations` to iterate over whole statements produced by a new `split_perl_statements` helper instead of iterating over raw lines (files changed: `crates/perl-module/src/resolution/use_lib.rs`, `crates/perl-module/tests/use_lib_resolution_unit_tests.rs`).
- Added `split_perl_statements` which splits on semicolons but does not split inside single- or double-quoted strings and honors escaped quotes, enabling multiline and parenthesized `use lib (...)` and `no lib (...)` forms.
- Kept the existing argument-parsing logic (`extract_paths_from_args`, `extract_qw_paths`, `extract_quoted_list`, etc.) so simple single-line forms remain unchanged.
- Added focused unit tests `multiline_use_lib_is_extracted`, `multiline_use_and_no_lib_are_ordered`, and `quoted_semicolon_does_not_split_statement` to cover multiline extraction, add/remove ordering, and quoted-semicolon handling.

### Testing

- Ran `cargo test -p perl-module --test use_lib_resolution_unit_tests` and all tests passed, including the three new tests and existing coverage.
- Ran `cargo check --all-targets -p perl-module` which completed successfully.
- Ran `cargo clippy -p perl-module -- -D warnings` which passed with no warnings.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead9eed5788333a6f4d9da3c5377c4)

PR #5882, opened 2026-04-24
fix(perl-module): improve multiline @INC pragma parsing (#0000)

EffortlessSteven:
### Motivation
- Line-by-line scanning missed or misordered parenthesized and multiline `use lib` / `no lib` statements, causing incorrect lexical @INC extraction and cancellation semantics.

### Description
- Add a small quote-aware statement splitter `split_perl_statements` that splits on semicolons but does not split inside single- or double-quoted strings, and avoids breaking escaped quote sequences.
- Update `extract_use_lib_paths` and `extract_use_lib_operations` to iterate over whole statements from `split_perl_statements` instead of raw `source.lines()` so multiline/parenthesized forms are recognized and ordering is preserved.
- Add focused unit tests and import to exercise the new behavior: `multiline_use_lib_is_extracted`, `multiline_use_and_no_lib_are_ordered`, and `quoted_semicolon_does_not_split_statement` in `crates/perl-module/tests/use_lib_resolution_unit_tests.rs`.
- Changes are confined to `crates/perl-module/src/resolution/use_lib.rs` and `crates/perl-module/tests/use_lib_resolution_unit_tests.rs`.

### Testing
- Ran `cargo test -p perl-module --test use_lib_resolution_unit_tests`, which passed (all added tests succeeded).
- Ran `cargo check --all-targets -p perl-module`, which completed successfully.
- Ran `cargo clippy -p perl-module -- -D warnings`, which completed successfully.
- Ran `cargo fmt --package perl-module`, which completed; `cargo xtask fmt` was attempted but failed due to an unrelated compile error in another crate (`crates/perl-lsp-rs-core/src/providers/formatting/formatting.rs`) and is not caused by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead9eed5788333a6f4d9da3c5377c4)

PR #5884, opened 2026-04-24
fix(perl-module): improve multiline @INC pragma parsing (#0000)

EffortlessSteven:
### Motivation
- The extractor previously scanned source line-by-line which missed multiline/parenthesized `use lib` and `no lib` statements and produced incorrect lexical ordering and cancellation semantics.
- The goal is to recognize whole Perl statements (without full parsing) and avoid splitting on semicolons inside quoted strings so common multiline forms are handled correctly.

### Description
- Added a small quote-aware statement splitter `split_perl_statements` that splits on semicolons while ignoring semicolons inside single- and double-quoted strings and respecting escaped characters.
- Switched `extract_use_lib_paths` and `extract_use_lib_operations` to operate over statements produced by `split_perl_statements` instead of raw lines.
- Preserved existing argument parsing logic; unchanged behavior for simple single-line `use lib` forms is retained.
- Modified files: `crates/perl-module/src/resolution/use_lib.rs` and `crates/perl-module/tests/use_lib_resolution_unit_tests.rs`, and added focused regression tests for multiline forms and quoted-semicolon handling.

### Testing
- Ran `cargo test -p perl-module --test use_lib_resolution_unit_tests` and all tests passed.
- Ran `cargo check --all-targets -p perl-module` and the check passed.
- Ran `cargo clippy -p perl-module -- -D warnings` and no lints were reported.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ead9eed5788333a6f4d9da3c5377c4)

### DIFF
--- a/crates/perl-module/src/resolution/use_lib.rs
+++ b/crates/perl-module/src/resolution/use_lib.rs
@@ -38,8 +38,8 @@ pub enum UseLibAction {
 pub fn extract_use_lib_paths(source: &str) -> Vec<UseLibPath> {
     let mut paths = Vec::new();
 
-    for line in source.lines() {
-        let trimmed = line.trim();
+    for statement in split_perl_statements(source) {
+        let trimmed = statement.trim();
         if let Some(rest) = strip_use_lib_prefix(trimmed) {
             extract_paths_from_args(rest, &mut paths);
         }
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as REDUNDANT — #5882 implements the multiline @INC statement-splitter fix with identical scope and test coverage. Winner selected for simplicity (bool-flag state tracking). Your [enum-based approach](https://github.com/.../pull/5884/files#diff-...) in #5883 was evaluated as architecturally sound but less minimal; the bool-flag approach in #5882 was preferred for maintenance.

PR #5902, opened 2026-04-24
fix(workspace): apply set_root_uri for initializationOptions.workspaceFolders (#0000)

EffortlessSteven:
### Motivation

- Initialization requests that supply `workspaceFolders` inside `initializationOptions` populated folder state but did not drive the same root-uri update path used by the normal initialize flow, leaving the server `root_path` and per-folder config inconsistent.

### Description

- Read `workspaceFolders` from `params.initializationOptions` when `params.workspaceFolders` is absent and route them through the same folder-registration path that calls `self.set_root_uri()` on the first folder.
- Use the existing `extract_workspace_folder_uris(...)` and existing folder creation logic so behavior is identical to the normal initialization flow.
- Add a regression unit test `initialize_with_init_options_workspace_folders_sets_root_and_loads_project_config` that initializes via `initializationOptions.workspaceFolders` and verifies root path, folder path, and that `.perl-lsp.toml` per-folder `include_paths` are loaded.

### Testing

- Ran the new regression unit test with `cargo test -p perl-lsp-rs --lib initialize_with_init_options_workspace_folders_sets_root_and_loads_project_config` and it passed.
- Verified the existing workspace-folder test with `cargo test -p perl-lsp-rs --lib initialize_with_workspace_folders_sets_root_path_from_first_folder` and it passed.
- Ran `cargo check --workspace --all-targets` to ensure no regressions and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eae704712883339ea29e7349000e0e)

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/lifecycle/capabilities.rs
+++ b/crates/perl-lsp-rs/src/runtime/lifecycle/capabilities.rs
@@ -206,9 +206,17 @@ impl LspServer {
             }
 
             // Initialize workspace folders
-            if let Some(workspace_folders) =
-                params.get("workspaceFolders").and_then(|f| f.as_array())
-            {
+            let workspace_folders = params
+                .get("workspaceFolders")
+                .and_then(|f| f.as_array())
+                .or_else(|| {
+                    params
+                        .get("initializationOptions")
+                        .and_then(|opts| opts.get("workspaceFolders"))
+                        .and_then(|f| f.as_array())
+                });
+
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as REDUNDANT — #5904 implements the same workspace folder initialization with the correct `or_else` chaining logic. #5904's approach safely tries the nested initializationOptions path even if the top-level workspaceFolders exists but is malformed.

Your implementation and test name variations were noted; #5904 was selected for the cleaner logical flow.

PR #5918, opened 2026-04-24
refactor(completion): plumb effective @INC paths into module completion (#4314)

EffortlessSteven:
### Motivation
- Module-name completion needed document-specific include-path context and opted-in system `@INC` visibility to enable future include-path-aware suggestions without implementing filesystem scanning yet.

### Description
- Thread document-scoped config and effective include paths into completion entrypoints by calling `config_for_doc(uri)`, `include_paths_for_doc(uri)`, and collecting system `@INC` only when `use_system_inc` is enabled. 
- Extend `CompletionProvider` with `include_paths: Vec<PathBuf>` and `system_inc_paths: Vec<PathBuf>` and add `new_with_index_source_paths(...)` that accepts those vectors while preserving the old `new_with_index_and_source(...)` behavior via empty-vector defaults. 
- Pass the new include/system vectors into the `use`/`require` module completion path by adjusting the call site and plumbing `add_use_module_completions(...)` signature; no scanning, ranking, or dedupe behavior was changed. 
- Add targeted tests verifying provider construction with provided include/system vectors and that existing `use` completion behavior is preserved when include/system vectors are empty.

### Testing
- Ran `cargo test -p perl-lsp-rs-core completion::` which passed (`148` tests in that suite ran and succeeded). 
- Ran `cargo check --workspace --all-targets` which completed successfully. 
- Ran `cargo test -p perl-lsp-completion` which failed because the package name does not exist in this workspace (command returned an error).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eae60d84c08333a48253285caa90a4)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
@@ -122,6 +122,7 @@ use perl_semantic_analyzer::symbol::{SymbolExtractor, SymbolKind, SymbolTable};
 use perl_semantic_analyzer::type_inference::TypeInferenceEngine;
 use perl_workspace::workspace_index::WorkspaceIndex;
 use std::collections::{HashMap, HashSet};
+use std::path::PathBuf;
 use std::sync::Arc;
 
 /// Maps module_name -> Set of explicitly imported symbol names.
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT in favor of #5921.

Winner #5921 uses clearer semantic naming: new_with_index_and_source_and_inc_paths makes it explicit these are @INC paths, not arbitrary search paths. Your method name (new_with_index_source_paths) breaks the codebase's established naming pattern.

All four PRs in this cluster are functionally equivalent. #5921's semantic clarity is the tiebreaker. Thank you!

PR #5919, opened 2026-04-24
refactor(completion): plumb effective @INC paths into module completion (#4314)

EffortlessSteven:
### Motivation
- Module-name completion lacked document-scoped include-path context and optional system `@INC` data, blocking later work to scan include paths for `.pm` files.
- Provide a minimal, safe plumbing layer to expose effective per-document include paths and opt-in system `@INC` to the completion provider without changing completion ranking or scanning behavior.

### Description
- Added `include_paths: Vec<PathBuf>` and `system_inc_paths: Vec<PathBuf>` fields to `CompletionProvider` and a new constructor `new_with_index_and_source_and_paths(...)` that accepts them while preserving existing constructors (which default to empty vectors).
- Runtime completion handlers now compute document-specific include paths and only probe system `@INC` when `use_system_inc` is enabled via `completion_paths_for_doc(...)`, and pass those vectors into the provider constructors in both normal and cancellable completion flows (`crates/perl-lsp-rs/src/runtime/language/completion.rs`).
- Threaded the new fields through the `use`/`require` module completion path by extending the `add_use_module_completions` signature (currently accepted but not yet used for filesystem scanning) in `workspace.rs`.
- Added targeted unit tests that verify a provider can be constructed with doc-specific include/system paths and that behavior is unchanged when empty vectors are provided (`crates/perl-lsp-rs-core/src/providers/completion/completion.rs`).

### Testing
- Attempted `cargo test -p perl-lsp-completion` which failed because that package name does not exist in the workspace (expected workspace package is `perl-lsp-rs-core`).
- Ran `cargo test -p perl-lsp-rs-core completion`, which succeeded (completion-related unit tests passed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eae60d84c08333a48253285caa90a4)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
@@ -122,6 +122,7 @@ use perl_semantic_analyzer::symbol::{SymbolExtractor, SymbolKind, SymbolTable};
 use perl_semantic_analyzer::type_inference::TypeInferenceEngine;
 use perl_workspace::workspace_index::WorkspaceIndex;
 use std::collections::{HashMap, HashSet};
+use std::path::PathBuf;
 use std::sync::Arc;
 
 /// Maps module_name -> Set of explicitly imported symbol names.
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT in favor of #5921.

All four PRs in this cluster are functionally equivalent implementations. #5921 wins on semantic clarity: new_with_index_and_source_and_inc_paths makes @INC intent explicit, while #5919/#5920 use generic 'paths'.

All preserve backward compat with delegating wrappers. Thank you!

PR #5964, opened 2026-04-24
fix(parser): correct transliteration part extraction and expand regressions (#6706)

EffortlessSteven:
### Motivation

- Fix correctness gaps in transliteration parsing (both `tr///` and `y///`) that caused replacement/modifier mis-attribution and brittle behavior for various delimiters.
- Handle both non-paired and paired delimiter forms, proper modifier separation, and obvious malformed-but-nonpanicking inputs. 
- Expand the regression coverage so the original fuzz-discovered repro is asserted and similar edge cases are prevented from regressing.

### Description

- Tighten `extract_transliteration_parts` in `crates/perl-parser-core/src/syntax/quote.rs` to allow optional whitespace after the operator, reject invalid delimiters (alphanumeric/whitespace), and determine paired vs non-paired delimiter semantics up-front. 
- Use strict extraction helpers (`extract_delimited_content_strict`, `extract_unpaired_body_skip_strings`, `extract_delimited_content_strict`) to track whether the search and replacement sections were properly closed and to avoid leaking characters into the modifier string. 
- Only parse and validate transliteration modifiers after the replacement section is confirmed closed; otherwise return empty modifiers to avoid misclassification. 
- Replace the print-based repro test with a focused regression bank in `crates/perl-parser/tests/fuzz_transliteration_crash_repro.rs` (renamed test `transliteration_regression_bank`) that asserts a matrix of non-paired, paired, mixed, edge-delimiter, and malformed-but-nonpanicking inputs.

### Testing

- Ran `cargo test -p perl-parser --test fuzz_transliteration_crash_repro`, which executed the minimal repro and the regression bank and reported all tests passed. 
- Ran formatting via `cargo xtask fmt` as part of the verification step and completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffa3b31c8333ac509b6c5961125f)

### DIFF
--- a/crates/perl-parser-core/src/syntax/quote.rs
+++ b/crates/perl-parser-core/src/syntax/quote.rs
@@ -302,82 +302,62 @@ pub fn extract_substitution_parts(text: &str) -> (String, String, String) {
 /// Extract search, replace, and modifiers from a transliteration token
 pub fn extract_transliteration_parts(text: &str) -> (String, String, String) {
     // Skip 'tr' or 'y' prefix
-    let content = if let Some(stripped) = text.strip_prefix("tr") {
+    let after_op = if let Some(stripped) = text.strip_prefix("tr") {
         stripped
     } else if let Some(stripped) = text.strip_prefix('y') {
         stripped
     } else {
         text
     };
+    // Perl allows optional whitespace between operator and delimiter.
+    let content = after_op.trim_start();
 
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Superseded by merged #6424 (parser hardening) and #6249 (qr/tr edge forms). The transliteration parsing improvements have landed via those PRs. Closing as redundant.

PR #5972, opened 2026-04-24
perf(incremental-checkpoint): tighten checkpoint window and segment invalidation

EffortlessSteven:
### Motivation
- Reduce unnecessary cache invalidation for common interior edits so incremental reparses reuse more cached work. 
- Make left/right checkpoint selection and relex bounds tighter to shrink re-lex windows. 
- Improve observability of tail-fallback cost so we can distinguish true cache gaps from conservative fallbacks.

### Description
- Change `TokenCache::cache_tokens` to emit fine-grained per-token segments instead of one whole-file segment to limit invalidation blast radius. 
- Capture lexer checkpoints when the lexer position crosses configured targets (`>=` check) and clamp `relex_end` if a malformed right-bound is observed to avoid inverted windows. 
- Add tail-fallback counters (`tail_fallback_tokens_relexed`, `tail_fallback_bytes_relexed`, `tail_fallbacks_with_segments`) and wire them into fallback accounting and `Display` output. 
- Add checkpoint-focused tests that assert the relex window remains bounded and that tail-fallback metrics are consistent while preserving AST equivalence with a fresh parse.

### Testing
- Ran the crate test suite with `cargo test -p perl-parser`; the run completed but a pre-existing unrelated test (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`) failed intermittently and is not caused by these changes. 
- Exercised the checkpoint-focused tests (`incremental`/`incremental_checkpoint` scenarios) using targeted test filters and observed no failures for the added assertions (no regressions in the incremental-checkpoint tests were observed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffb24cf8833385c03e70823ddd2c)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -322,12 +322,17 @@ impl TokenCache {
     /// * `end` - End byte position of the segment.
     /// * `tokens` - Parser tokens for this segment.
     fn cache_tokens(&mut self, start: usize, end: usize, tokens: Vec<Token>) {
-        if tokens.is_empty() {
+        if tokens.is_empty() || start >= end {
             return;
         }
 
-        let segment = TokenSegment::new(start, end, tokens);
-        self.add_segment(segment);
+        // Store fine-grained per-token segments so interior edits can invalidate
+        // and reuse cache with tighter boundaries.
+        for token in tokens {
+            if token.start < token.end {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as REDUNDANT - #6001 is the cluster winner. Winner adds segment-granular splitting (preserving non-overlapping prefix/suffix tokens instead of dropping whole segments), tighter tail-fallback accounting, and per-segment metrics. Your implementation has coarser invalidation and conflates metric paths. Thank you.

PR #5973, opened 2026-04-24

EffortlessSteven:
### Motivation
- Batch edits that only shift unaffected subtrees were not being matched reliably, so safe subtree reuse was undercounted for non-overlapping edits and separate whitespace/comment or identifier/value edits.
- The advanced reuse analyzer already computes edit- and position-aware information but its shifted-node matching wasn't projecting old positions through the exact edit sequence before scoring.

### Description
- Pass the `EditSet` into `find_position_shifted_matches` and use `edits.apply_to_position(...)` to project old node positions into the new document so shifted candidates are preferred when appropriate.
- Improve shifted-node scoring by adding a projection proximity bonus and selecting the best candidate per old node (confidence + projected-distance tie-breaker), while respecting `max_position_shift` and `min_confidence` limits.
- Relax the advanced-reuse acceptance in `incremental_v2` so multi-edit batches with actionable reuse (non-zero reused nodes) will use the analyzer even if the single-edit quality gate would not pass; single-edit quality checks remain unchanged.
- Add regression tests to `incremental_v2` that exercise (a) non-overlapping whitespace/comment edits and (b) separated identifier + value edits, and assert AST equivalence with a fresh full parse to ensure correctness.

### Testing
- Ran the new incremental-v2 regression tests individually and they passed, including the fresh-parse equivalence assertions for the generated trees.
- Ran targeted existing tests in isolation (the refactor cleanup test that had been flaky) and they passed when run alone; the full `cargo test -p perl-parser` run encountered unrelated pre-existing failures in other test modules while the modified incremental code compiled and behaved as expected.
- Ran `cargo xtask fmt` to format changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffad48548333b1fa7c7a5297e07f)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
+++ b/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
@@ -158,7 +158,13 @@ impl AdvancedReuseAnalyzer {
         self.find_direct_structural_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
 
         // Strategy 2: Position-shifted matching
-        self.find_position_shifted_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
+        self.find_position_shifted_matches(
+            &old_analysis,
+            &new_analysis,
+            &mut reuse_map,
+            edits,
+            config,
+        );
 
         // Strategy 3: Content-updated matching
         if config.enable_content_reuse {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as REDUNDANT - #5985 implements the same scope more completely with more targeted regression tests for multi-edit, whitespace/comment, and mixed identifier/value scenarios. 

Your implementation approach (position-shifted matching with confidence scoring) is the same; winner's regression test suite is more comprehensive, testing both AST equivalence and reuse metrics across edge cases.

Thank you for contributing to the ensemble.

PR #5978, opened 2026-04-24
fix(incremental): harden batch edit normalization and UTF-8 boundary fallback (#6703)

EffortlessSteven:
### Motivation
- Batch incremental edits relied on optimistic byte-range and UTF-8 assumptions which could produce nondeterministic or unsafe outcomes for overlapping, backwards, out-of-bounds, or mid-codepoint edits.

### Description
- Added `IncrementalEditSet::normalized_reverse_non_overlapping` to validate and normalize batches, rejecting backwards ranges, out-of-bounds ranges, and overlapping edits before fast-path application (`crates/perl-parser/src/incremental/incremental_edit.rs`).
- Implemented a mappable-batch gate in `IncrementalDocument::apply_edits` to require normalized, non-overlapping, and UTF-8-aligned edits; unmappable batches now fall back to conservative full-text application plus a full reparse (`crates/perl-parser/src/incremental/incremental_document.rs`).
- Hardened single-edit and in-place application with explicit backwards/out-of-bounds checks and UTF-8 char-boundary checks to avoid corrupting `source` on malformed ranges (`crates/perl-parser/src/incremental/incremental_document.rs`).
- Added targeted regression tests for boundary conditions covering overlapping edits, backwards ranges, mid-codepoint edits, one-bad-edit batch fallback, and an incremental-vs-fresh-parse parity case (`crates/perl-parser/src/incremental/incremental_boundary_regressions.rs`) and wired the test module into the incremental tree (`crates/perl-parser/src/incremental/mod.rs`).

### Testing
- Ran `cargo xtask fmt` to format changes successfully.
- Ran `cargo check --all-targets -p perl-parser` and `cargo check --all-targets -p perl-parser --features incremental`, both succeeded.
- Ran targeted tests with `cargo test -p perl-parser --features incremental --lib incremental::incremental_boundary_regressions::tests` and the new boundary regression suite passed (5/5 tests).
- Note: a pre-existing unrelated test (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`) fails in the full `cargo test -p perl-parser`; this failure predates and is unrelated to these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaff94025083338c5ee4779cd0492d)

PR #5977, opened 2026-04-24
fix(incremental): harden batch edit normalization and UTF-8 boundary fallback (#6704)

EffortlessSteven:
### Motivation
- Batch incremental edits assumed well-formed byte ranges and UTF-8 boundaries which could lead to optimistic/partial application for malformed edit sets.
- Make batch application deterministic and safe by validating ranges and falling back conservatively when any edit is unmappable.

### Description
- Add deterministic normalization and validation in `IncrementalEditSet::normalized_reverse()` to reject backwards ranges and overlapping edits while preserving valid zero-width insertions, and expose `has_ordered_range()` on `IncrementalEdit` (`crates/perl-parser/src/incremental/incremental_edit.rs`).
- Harden `IncrementalDocument` batch/single-edit application by adding strict bounds and UTF-8 boundary checks, `try_apply_batch_to_source()`, and `fallback_full_reparse()` so a malformed batch or unmappable edit triggers a conservative full reparse; preserve existing reverse-order semantics for valid batches (`crates/perl-parser/src/incremental/incremental_document.rs`).
- Replace optimistic multi-edit reuse with a conservative full parse after successful batch mapping to ensure incremental state equals a fresh parse for supported cases.
- Add targeted regression tests exercising overlapping batches, backwards ranges, mid-codepoint edits, one-bad-edit batch fallback, and successful batch parity with a fresh parse (`crates/perl-parser/tests/incremental_batch_boundary_regressions.rs`).

### Testing
- Ran `cargo test -p perl-parser --features incremental --test incremental_batch_boundary_regressions` and the new regression suite passed (5/5 tests).
- Ran `cargo check --all-targets -p perl-parser` which completed successfully.
- Ran the full `cargo test -p perl-parser`; the run failed due to an unrelated existing test in `refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`, not due to changes in incremental code.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaff94025083338c5ee4779cd0492d)

PR #5983, opened 2026-04-24
fix(incremental): harden batch edit normalization and UTF-8 boundary fallback (#6698)

EffortlessSteven:
### Motivation

- Batch edits could be applied with overlapping or backwards ranges or at non-UTF-8 byte boundaries which risks producing corrupted source or inconsistent incremental trees. 
- The incremental code relied on optimistic in-place application and optimistic fast-path reuse, so a single malformed edit could subtly break byte/char invariants or incremental parity with a fresh parse. 
- Make the batch path conservative and deterministic so malformed batches degrade safely to a full parse rather than corrupting state.

### Description

- Added `IncrementalEditSet::normalize_for_batch` to validate and deterministically sort edits, rejecting backwards ranges and overlapping edits while preserving zero-width insertions. (file: `crates/perl-parser/src/incremental/incremental_edit.rs`) 
- Updated `IncrementalDocument::apply_edits` to use normalization, preserve reverse-order application semantics for valid batches, and fall back to a conservative full-parse when normalization fails or any in-place edit is unmappable. (file: `crates/perl-parser/src/incremental/incremental_document.rs`) 
- Strengthened bounds and UTF-8 char-boundary checks in `apply_edit_to_string` and `apply_edit_in_place`, and made in-place application return a boolean success so the batch can detect unmappable edits. (file: `crates/perl-parser/src/incremental/incremental_document.rs`) 
- Added a safety parity check that compares the incremental/with-reuse parse result against a fresh parse and uses the fresh parse when they diverge, and added a focused regression test file covering overlapping edits, backwards ranges, mid-codepoint edits, one-bad-edit batch fallback, and incremental == fresh parse parity. (tests: `crates/perl-parser/tests/incremental_batch_boundary_regressions.rs`) 

### Testing

- Ran `cargo test -p perl-parser --test incremental_batch_boundary_regressions --features incremental` and the new regression suite passed (`5 passed`).
- Ran full `cargo test -p perl-parser` which uncovered a pre-existing unrelated test failure in refactor retention tests (`test_cleanup_respects_retention_count`), so the incremental-focused tests and checks introduced by this change are green while a separate unrelated test remains failing in the wider suite.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaff94025083338c5ee4779cd0492d)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_document.rs
+++ b/crates/perl-parser/src/incremental/incremental_document.rs
@@ -128,40 +128,56 @@ impl IncrementalDocument {
         // Reset metrics for this batch of edits
         self.metrics = ParseMetrics::default();
 
-        // Sort edits by position (reverse order for correct application)
-        let mut sorted_edits = edits.edits.clone();
-        sorted_edits.sort_by(|a, b| b.start_byte.cmp(&a.start_byte));
+        let mut normalized_edits = match edits.normalize_for_batch() {
+            Some(normalized) => normalized,
+            None => {
+                debug!(
+                    "Batch normalization failed (overlap/backwards); using conservative fallback"
+                );
+                return self.reparse_from_source(edits.apply_to_string(&self.source), start);
+            }
+        };
 
         // Apply all edits to source in-place.
         // Reverse ordering keeps byte offsets stable while avoiding repeated
         // full-string allocations for each edit.
         let mut new_source = self.source.clone();
-        for edit in &sorted_edits {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Duplicate of #5977 (Codex ensemble burst, same issue family — batch edit normalization). Note: #5984 and #5983 produce byte-identical diffs. Keeping #5977 (102 test additions vs 93 here) and #5978 (uses a separate boundary regressions module — distinct approach worth keeping). Closing as ensemble dupe.

PR #5985, opened 2026-04-24
perf(incremental-v2): improve safe shifted-node reuse across batch edits (#6696)

EffortlessSteven:
### Motivation
- Batch edits previously missed safe subtree reuse because shifted matching skipped expected targets and allowed multiple old nodes to claim the same new node, reducing reuse for non-overlapping and non-structural edits.
- Keep AST equivalence with a full parse as the hard constraint while increasing safe reuse for identifier/value and whitespace/comment edits.

### Description
- Feed the `EditSet` into position-shift matching and predict expected shifted targets with `map_old_position_to_new`, improving hit-rate for edits that only change positions. (`crates/perl-parser/src/incremental/incremental_advanced_reuse.rs`)
- Enforce one-to-one old→new mapping across strategies via a central `try_register_match` to prevent multiple old nodes from mapping to the same new node. (`incremental_advanced_reuse.rs`)
- Add a fallback nearby-candidate search when the edit-derived expected location misses, with confidence penalties for shifts and position-shift thresholds honored. (`incremental_advanced_reuse.rs`)
- Use the centralized registration helper from direct/content/aggressive strategies to reduce duplicated logic and make reuse accounting safer. (`incremental_advanced_reuse.rs`)
- Add targeted regression tests exercising multiple non-overlapping edits, multi-region whitespace/comment edits, and mixed identifier/value edits with incremental-vs-full-parse AST equivalence checks. Tests added as a standalone test file and as in-module tests for `incremental_v2`. (`crates/perl-parser/tests/incremental_v2_shifted_reuse_regression.rs`, `crates/perl-parser/src/incremental/incremental_v2.rs`)

### Testing
- Ran `cargo test -p perl-parser --features incremental --test incremental_v2_shifted_reuse_regression -- --nocapture` and the new incremental-v2 regression tests passed. ✅
- Ran `cargo check --all-targets -p perl-parser` and it completed successfully. ✅
- Ran `cargo xtask fmt` to format code. ✅
- Ran `cargo test -p perl-parser` and observed an existing unrelated failure in refactoring validation (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`) that predates these changes; the failure is not caused by this reuse work. ❌

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffad48548333b1fa7c7a5297e07f)

PR #5992, opened 2026-04-24

EffortlessSteven:
### Motivation

- Batch edits with non-overlapping changes or separated whitespace/comment edits under-report safe subtree reuse because shifted nodes were not matched relative to pending edits.
- We want to increase safe subtree reuse for multi-edit scenarios and simple identifier/value edits while keeping the returned AST identical to a fresh full parse.

### Description

- Thread `EditSet` into position-shifted matching and add `project_old_position_to_new` to project old byte positions through pending edits to predict expected target positions for unaffected nodes (file: `crates/perl-parser/src/incremental/incremental_advanced_reuse.rs`).
- Select best shifted candidate per old node, skip nodes marked as affected, prefer exact projected positions by bumping confidence, and avoid reusing the same new node twice (file: `crates/perl-parser/src/incremental/incremental_advanced_reuse.rs`).
- Accept advanced reuse analysis when it reports shifted-position matches (`position_adjustments > 0`) while still returning the freshly parsed `new_tree` so AST equivalence with a full parse remains the hard correctness constraint (file: `crates/perl-parser/src/incremental/incremental_v2.rs`).
- Add feature-gated regression tests that assert incremental output equals a fresh parse and that shifted matching yields at least as much reuse as a conservative no-shift configuration (file: `crates/perl-parser/tests/incremental_v2_shifted_reuse_regression.rs`).

### Testing

- Ran `cargo test -p perl-parser --features incremental --test incremental_v2_shifted_reuse_regression` and both new regression tests passed.
- Ran `cargo test -p perl-parser` which hit an existing unrelated test failure (`test_complex_subroutine_signatures`); this failure is not caused by these changes and is pre-existing in the suite.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffad48548333b1fa7c7a5297e07f)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
+++ b/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
@@ -158,7 +158,13 @@ impl AdvancedReuseAnalyzer {
         self.find_direct_structural_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
 
         // Strategy 2: Position-shifted matching
-        self.find_position_shifted_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
+        self.find_position_shifted_matches(
+            &old_analysis,
+            &new_analysis,
+            edits,
+            &mut reuse_map,
+            config,
+        );
 
         // Strategy 3: Content-updated matching
         if config.enable_content_reuse {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT - #5985 is the winner. Both PRs share the same approach (EditSet-aware position-shift matching, projection of old positions through pending edits, one-to-one mapping enforcement). 

#5985 edge case: adds fallback nearby-candidate search with distance-based confidence penalties when edit-derived expected location misses. This edge case is a novel extraction from your PR that improves robustness—will be incorporated into #5985.

Thank you.

PR #5993, opened 2026-04-24
perf(incremental-checkpoint): tighten checkpoint window and segment invalidation (#0000)

EffortlessSteven:
### Motivation

- Reduce unnecessary token-cache invalidation and large tail re-lex windows for common edits by making checkpoint-window and segment invalidation more precise while preserving correctness-first fallback behavior.

### Description

- Make `TokenCache::invalidate_range` segment-aware so overlapping segments are split into reusable prefix/suffix segments instead of dropping the whole segment, and return the count of invalidated segments.
- Change incremental reparse caching to only store the region actually re-lexed (`relexed_parser_tokens`) so unaffected prefix/suffix segments remain cached for future edits.
- Add `tail_fallback_with_cached_segments` to `IncrementalStats`, wire it into reporting, and improve bytes/accounting for tail re-lexing when a suffix exists but cannot be reused.
- Add checkpoint-focused tests that assert (1) invalidation preserves prefix/suffix segments and (2) segmented-tail fallback is reported.

### Testing

- `cargo test -p perl-parser` was run: the test run produced `132 passed; 1 failed` where the single failure is a pre-existing, unrelated test (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`).
- Formatting was applied via `cargo xtask fmt` / `rustfmt` on the modified file.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffb24cf8833385c03e70823ddd2c)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_checkpoint.rs
+++ b/crates/perl-parser/src/incremental/incremental_checkpoint.rs
@@ -168,23 +168,6 @@ impl TokenCache {
         self.segments.iter().filter(|seg| seg.is_after(position)).cloned().collect()
     }
 
-    /// Return all segments that overlap with the given range.
-    ///
-    /// This is useful for determining which segments would be affected by
-    /// an edit or for finding cached tokens in a specific region.
-    ///
-    /// # Arguments
-    ///
-    /// * `start` - Start byte position of the range.
-    /// * `end` - End byte position of the range.
-    ///
-    /// # Returns
-    ///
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as REDUNDANT - #6001 is the checkpoint-tightening cluster winner. Same scope, but winner's implementation is more complete (segment-granular splitting, enhanced tail-fallback metrics). Thank you.

PR #6003, opened 2026-04-24

EffortlessSteven:
### Motivation
- Batch edits that only shift unaffected subtrees (non-overlapping edits, separated whitespace/comment edits, simple identifier/value changes) were not being reused as aggressively as they could be without weakening correctness.
- Use the existing advanced reuse analyzer to recognise safe position-shifted reuse rather than duplicating heuristics in incremental_v2.

### Description
- Added an edit-aware shifted-position matching pass `find_edit_shifted_matches` to `AdvancedReuseAnalyzer` which predicts each unaffected node's expected post-edit byte position from cumulative edit deltas and matches by structural/content hashes when safe.
- Added `expected_shifted_position` helper to compute cumulative shifts across an `EditSet` and used it to drive edit-aware matching.
- Tightened generic `find_position_shifted_matches` to skip nodes marked as affected by edits so only safe candidates are considered.
- Exposed analysis metrics via `position_adjustments` increments when shifted matches are chosen so tests can assert shifted-match activity.
- Added unit tests: a focused analyzer test for cumulative shift math (`test_expected_shifted_position_with_multiple_edits`) and two incremental-v2 regression tests that assert improved reuse and AST equivalence for non-overlapping multi-edit and separated whitespace/comment edit scenarios; added a `replacement_edit` test helper to build consistent `Edit` values.

### Testing
- Ran `cargo xtask fmt` successfully.
- Ran `cargo check --all-targets -p perl-parser` successfully.
- Ran full test suite `cargo test -p perl-parser`; result: most tests passed but one unrelated pre-existing test failed: `refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count` (132 passed, 1 failed) and is not caused by these changes.
- Ran the new/targeted incremental tests individually and via filtered runs; the new analyzer and incremental-v2 regression tests passed and the incremental-v2 outputs matched fresh full-parse ASTs.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb009eed908333badab3c2a8950ad1)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
+++ b/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
@@ -157,15 +157,18 @@ impl AdvancedReuseAnalyzer {
         // Strategy 1: Direct structural matching
         self.find_direct_structural_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
 
-        // Strategy 2: Position-shifted matching
+        // Strategy 2: Edit-aware position-shifted matching
+        self.find_edit_shifted_matches(&old_analysis, &new_analysis, edits, &mut reuse_map, config);
+
+        // Strategy 3: Generic position-shifted matching
         self.find_position_shifted_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
 
-        // Strategy 3: Content-updated matching
+        // Strategy 4: Content-updated matching
         if config.enable_content_reuse {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as REDUNDANT - #5985 wins this cluster. Both implement the same position-shift matching with EditSet awareness. #5985 is the more complete implementation. Thank you.

PR #6005, opened 2026-04-24
refactor(incremental-v2): improve shifted-node reuse across batch edits (#0000)

EffortlessSteven:
### Motivation

- Incremental v2 under-counted safe shifted-node reuse for multi-edit batches, hurting reuse for non-overlapping edits and separated whitespace/comment edits.
- Improve reuse for simple identifier/value edits where only node positions shift while keeping AST equivalence with a fresh full parse as the correctness constraint.

### Description

- Make position-shifted matching edit-aware by passing the incoming `EditSet` into `find_position_shifted_matches` and skipping nodes marked as affected by edits.
- Add `estimated_shift_at` to compute expected post-edit positions and prefer position-shift matches close to the expected shifted location, applying a distance-based confidence penalty.
- Score candidates by combined structural/content confidence and distance-from-expected-position, inserting chosen `PositionShift` strategies into the reuse map and incrementing `analysis_stats.position_adjustments`.
- Add focused regression tests (feature-gated `incremental`) that assert fresh-parse AST equivalence and improved reuse for: non-overlapping identifier/value edits and multi-region whitespace/comment edits, plus small helpers in the in-module tests.

### Testing

- Ran `cargo test -p perl-parser --features incremental --test incremental_v2_reuse_regression` and the new regression tests passed.
- Ran `cargo check --all-targets -p perl-parser` and `cargo check --all-targets -p perl-parser --features incremental` and both succeeded.
- Ran broader `cargo test -p perl-parser` during development and observed an intermittent, unrelated failure in an external refactor test when run as part of the full suite; that failing test passes when run in isolation and is not caused by these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb009eed908333badab3c2a8950ad1)

### DIFF
--- a/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
+++ b/crates/perl-parser/src/incremental/incremental_advanced_reuse.rs
@@ -158,7 +158,13 @@ impl AdvancedReuseAnalyzer {
         self.find_direct_structural_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
 
         // Strategy 2: Position-shifted matching
-        self.find_position_shifted_matches(&old_analysis, &new_analysis, &mut reuse_map, config);
+        self.find_position_shifted_matches(
+            &old_analysis,
+            &new_analysis,
+            edits,
+            &mut reuse_map,
+            config,
+        );
 
         // Strategy 3: Content-updated matching
         if config.enable_content_reuse {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT - #5985 is the cluster winner. Both implement EditSet-aware position-shift matching. #5985 adds more detailed fallback search with distance-based confidence penalties and more extensive test coverage of edge cases (non-overlapping edits, multi-region comments, identifier/value mixes). Thank you.

PR #6007, opened 2026-04-24
fix(parser-core): handle package v-string versions (#6688)

EffortlessSteven:
### Motivation

- Package declarations that include v-string versions (for example `package My::App v5.38;`) were previously parsed with the version dropped, producing an incorrect package name and AST shape.

### Description

- Treat `TokenKind::VString` as a valid package version token in `parse_package` and add a regression test `parse_package_declaration_with_vstring_version` verifying the `Package` node `name` includes the v-string; changes are in `crates/perl-parser-core/src/engine/parser/declarations.rs` and `crates/perl-parser-core/tests/parser_tests.rs`.

### Testing

- Ran `cargo test -p perl-parser-core --test parser_tests parse_package_declaration_with_vstring_version` which passes.
- Ran `cargo check --workspace --all-targets` which passes, and note that full `cargo test -p perl-parser-core` and `cargo test -p perl-parser` were exercised but encountered unrelated pre-existing test failures in this environment.

------


EffortlessSteven:
### Review: handle package v-string versions

**Real Perl-syntax gap closed.** `package My::App v5.38;` is canonical Perl 5.36+ syntax (introduced when `use feature 'class'` started recommending v-strings for module versions). Master previously only accepted `package My::App 1.234;` -- a bare numeric version.

**The fix is two lines**:

```rust
- let version = if self.peek_kind() == Some(TokenKind::Number) {
+ let version = if self.peek_kind() == Some(TokenKind::Number)
+     || self.peek_kind() == Some(TokenKind::VString)
+ {
       Some(self.tokens.next()?.text.to_string())
   }
```

Minimal, targeted, correct. The lexer already emits `TokenKind::VString` for `v5.38`-style tokens; the parser just wasn't accepting it in the version slot.

**Test pins the contract**:

```rust
let mut parser = Parser::new("package My::App v5.38;");
// ...
V1NodeKind::Package { name, block, .. } => {
    assert_eq!(name, "My::App v5.38");
    assert!(block.is_none(), "package statement form should not have a block");
}
```

Worth noting: the test uses `assert_eq!(name, "My::App v5.38")` -- so the version is being **concatenated into the name string** rather than stored in a separate `version` field. That's worth checking. Looking at the diff, `version` is captured locally:

```rust
let version = if self.peek_kind() == Some(TokenKind::Number) || ... {
    Some(self.tokens.next()?.text.to_string())
}
```

But the test asserts `name == "My::App v5.38"`, suggesting the version gets folded into the name downstream. If `Package { name, version, block, .. }` carries `version` as a separate field, the test should ideally also assert on it:

```rust
V1NodeKind::Package { name, version, block, .. } => {
    assert_eq!(name, "My::App");
    assert_eq!(version.as_deref(), Some("v5.38"));
    assert!(block.is_none());
}
```

If version is *not* a separate field on the AST node, that's a separate (probably long-standing) gap -- the version is being thrown away or stringified into the name. Worth a follow-up issue.

**Concerns**:

1. **Test assertion ambiguity** (above): unclear whether `version` makes it into the AST. The test asserts on the concatenated string, suggesting the `version` local is *unused* after capture. Look at what happens after:
   ```rust
   let version = if ... { Some(...) } else if let Some(TokenKind::Identifier) = ... { /* check v-string */ };
   ```
   The diff doesn't show what's done with `version` afterward. If it's discarded (only used for parser-state advancement, not stored), then this fix only allows the version *token* to be consumed but doesn't surface it. The Package node's `name` containing `"My::App v5.38"` is suspicious -- that's likely the bug, not the fix.

2. **The `else if let Some(TokenKind::Identifier)` branch** is the existing v-string-as-identifier fallback (e.g., `v5.38` if the lexer happens to tokenize it as Identifier rather than VString). This branch was the *workaround* for the bug this PR is fixing. With proper `VString` support, the `Identifier` branch may now be dead code -- worth checking if any test exercises it.

3. **Edge cases not tested**:
   - `package My::App v5;` (single-component v-string)
   - `package My::App v5.38.0;` (three-component v-string)
   - `package My::App v5.38 { ... }` (v-string with block form)
   - `package My::App;` regression (no version, must still work)

4. **`use My::App v5.38;`** also accepts v-strings on `use` lines. The parser code shown in the diff is only for `package`, so `use` may have had this support already (from PR #6009's larger patch which I'm reviewing separately, the `use` parser does check both `Number` and `VString`). Worth confirming `use My::App v5.38;` already works.

1. **Verify `version` reaches the AST.** If the `Package` node has a `version` field, assert on it directly:
   ```rust
   V1NodeKind::Package { name, version, .. } => {
       assert_eq!(name, "My::App");
       assert_eq!(version.as_deref(), Some("v5.38"));
   }
   ```
   If `version` is being thrown away, that's a follow-up bug to file separately.

2. **Add the block-form test** (`package My::App v5.38 { ... }`):
   ```rust
   let mut parser = Parser::new("package My::App v5.38 { my $x = 1; }");
   ```
   This is the form Perl 5.38+ recommends and is the most common use of package-with-version in modern code.

3. **Add a regression test for the bareword identifier fallback**:
   ```rust
   // Lexer might produce `Identifier("v5")` rather than `VString` in some contexts
   let mut parser = Parser::new("package My::App v5;");
   ```

**Verdict**: small, correct fix that closes a real Perl-5.36+ gap. The implementation is right; the test coverage could be tightened (multi-component v-strings, block form, version field on AST). Two-line fix, two-line review.

EffortlessSteven:

EffortlessSteven:

PR #6008, opened 2026-04-24
fix(parser-core): recover scoped sub declaration name errors (#6687)

EffortlessSteven:
### Motivation

- Address recovery and AST-shape loss for scoped sub declarations (`my|our|state sub`) when the sub name is missing, which previously flowed into anonymous-sub expression handling and could drop declaration structure.

### Description

- When parsing a `my|our|state sub` head, detect the case where the parsed `Subroutine` has no `name` and emit a targeted `ParseError::unexpected("subroutine name", ...)` diagnostic while returning the `Subroutine` node to preserve statement shape (`crates/perl-parser-core/src/engine/parser/statements.rs`).
- Added regression tests that assert a forward `my sub helper;` remains a `Subroutine` and that `my sub ;` recovers into a `Subroutine` node without losing the following declaration and emits the expected diagnostic (`crates/perl-parser-core/tests/fix_scoped_sub_declarations.rs`).
- No changes to public APIs; change is narrow and scoped to statement parsing and recovery behavior.

### Testing

- Ran `cargo test -p perl-parser-core --test fix_scoped_sub_declarations`, which passed.
- Ran `cargo check --workspace --all-targets`, which passed.
- Note: running the full test suites (`cargo test -p perl-parser-core` and `cargo test -p perl-parser`) exercised unrelated existing failures in other tests in this workspace; those failures are pre-existing and not caused by this focused change.

------

### DIFF
--- a/crates/perl-symbol/src/surface/mod.rs
+++ b/crates/perl-symbol/src/surface/mod.rs
@@ -30,6 +30,7 @@
 pub mod decl;
 pub mod facts;
 pub mod r#ref;
+pub mod ref_facts;
 
 pub use decl::{SymbolDecl, extract_symbol_decls};
 pub use facts::{
@@ -37,3 +38,4 @@ pub use facts::{
     symbol_refs_to_semantic_facts,
 };
 pub use r#ref::{SymbolRef, SymbolRefKind, extract_symbol_refs};
+pub use ref_facts::{SymbolRefSemanticFacts, symbol_refs_to_semantic_facts};
--- a/crates/perl-symbol/src/surface/ref_facts.rs
+++ b/crates/perl-symbol/src/surface/ref_facts.rs
@@ -1,0 +1,136 @@
+use crate::surface::r#ref::{SymbolRef, SymbolRefKind};
+use perl_semantic_facts::{
+    AnchorFact, AnchorId, Confidence, EdgeFact, EdgeId, EdgeKind, EntityId, FileId, OccurrenceFact,
+    OccurrenceId, OccurrenceKind, Provenance,
+};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SymbolRefSemanticFacts {
+    pub anchors: Vec<AnchorFact>,
+    pub occurrences: Vec<OccurrenceFact>,
+    pub references_edges: Vec<EdgeFact>,
+}
+
+pub fn symbol_refs_to_semantic_facts(
+    refs: &[SymbolRef],
+    file_id: FileId,
+    resolve_entity_id: impl Fn(&SymbolRef) -> Option<EntityId>,
+) -> SymbolRefSemanticFacts {
+    let mut anchors = Vec::with_capacity(refs.len());
+    let mut occurrences = Vec::with_capacity(refs.len());
+    let mut references_edges = Vec::new();
+
+    for symbol_ref in refs {
+        let anchor_span = symbol_ref.anchor_span.unwrap_or(symbol_ref.full_span);
+        let anchor_id = AnchorId(stable_id("ref_anchor", &symbol_ref.qualified_name, anchor_span.0, anchor_span.1));
+        anchors.push(AnchorFact {
+            id: anchor_id,
+            file_id,
+            span_start_byte: anchor_span.0 as u32,
+            span_end_byte: anchor_span.1 as u32,
+            scope_id: None,
+            provenance: Provenance::ExactAst,
+            confidence: Confidence::High,
+        });
+
+        let entity_id = resolve_entity_id(symbol_ref);
+        let occurrence_id = OccurrenceId(stable_id(
+            "ref_occurrence",
+            &symbol_ref.qualified_name,
+            symbol_ref.full_span.0,
+            symbol_ref.full_span.1,
+        ));
+        occurrences.push(OccurrenceFact {
+            id: occurrence_id,
+            kind: occurrence_kind(symbol_ref),
+            entity_id,
+            anchor_id,
+            scope_id: None,
+            provenance: Provenance::ExactAst,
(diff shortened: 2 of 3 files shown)


EffortlessSteven:
### Review: recover scoped sub declaration name errors

**Real parser-recovery improvement.** Before this fix, malformed `my sub ;` (scoped sub declaration with missing name) would silently parse as a nameless `Subroutine` node *without* recording a diagnostic. Editors would see no red squiggle on the malformed declaration -- the user just saw a function with no name and got no LSP help.

Worse: looking at the test `recovers_scoped_sub_missing_name_without_losing_following_declaration`, the *previous* behavior likely lost the following `my $x = 1;` declaration entirely (since the parser may have consumed tokens trying to find a sub name). The fix preserves both the malformed sub *and* the following declaration as recoverable AST nodes.

**The implementation is targeted**:

```rust
let mut sub_node = self.parse_subroutine()?;
sub_node.location.start = decl_token.start;
if let NodeKind::Subroutine { name, .. } = &sub_node.kind
    && name.is_none()
{
    let error_pos = self.current_position();
    let found = self
        .peek_kind()
        .map_or_else(|| "EOF".to_string(), |kind| kind.display_name().to_string());
    self.record_error(ParseError::unexpected(
        "subroutine name",
        found,
        error_pos,
    ));
    Ok(sub_node)
} else {
    self.finish_subroutine_statement(sub_node)
}
```

Three things are right:

1. **Detect the bug condition cleanly**: `name.is_none()` after `parse_subroutine()` returned successfully is the canonical signal. `parse_subroutine` is permissive (tolerates missing names because anonymous subs are valid in expression contexts -- `my $f = sub { ... };`). The caller has the context "scoped sub declarations require a name" that `parse_subroutine` doesn't.

2. **Use `unexpected("subroutine name", found, error_pos)`** -- the structured error variant. Editors and LSP can surface this as "expected: subroutine name; found: ;". Better than a freeform message.

3. **Skip `finish_subroutine_statement`** for the malformed case. `finish_subroutine_statement` likely expects a fully-formed sub (block, possibly attributes); calling it on a half-parsed node could produce cascading errors. Returning `Ok(sub_node)` directly preserves the partial AST and lets the *next* statement be parsed cleanly.

**Tests pin both the success path and the recovery path**:

| Test | Input | Expected |
|---|---|---|
| `parses_scoped_sub_forward_declaration` | `my sub helper; my $x = 1;` | both statements parse cleanly, no errors |
| `recovers_scoped_sub_missing_name_without_losing_following_declaration` | `my sub ; my $x = 1;` | both statements parse, diagnostic mentions "subroutine name" |

The forward-declaration test is the regression guard -- without it, this fix could over-trigger on `my sub helper;` (a *legal* forward declaration with no body but a name) and surface a spurious "missing name" error. The test confirms `name.is_none()` (not `body.is_none()`) is the right discriminator.

**Concerns**:

1. **The `if let ... && ...` chain** uses Rust 2024 let-chain syntax:
   ```rust
   if let NodeKind::Subroutine { name, .. } = &sub_node.kind
       && name.is_none()
   ```
   This requires Rust 2024 edition (or `let_chains` nightly feature). Confirm `crates/perl-parser-core/Cargo.toml` is on edition 2024 or has the feature enabled. If not, it needs to be:
   ```rust
   if let NodeKind::Subroutine { name, .. } = &sub_node.kind {
       if name.is_none() {
           // ...
       }
   }
   ```
   The other parser-core code likely already uses let-chains, so probably fine, but worth a `cargo check` confirmation.

2. **`error_pos = self.current_position()`** is captured *after* `parse_subroutine` returned. That position is the token *after* whatever `parse_subroutine` consumed -- typically the `;` or the next statement. The diagnostic would point at the `;`, not at the `sub` keyword or the missing-name slot. Consider:
   ```rust
   let error_pos = sub_node.location.end;  // end of the partial sub, where the name should have been
   ```
   This points the diagnostic at the boundary where the name was expected, which is more user-friendly. Minor; the current behavior is "good enough" but not perfect.

3. **Error variant is `unexpected("subroutine name", found, error_pos)`** -- the test asserts:
   ```rust
   assert!(diagnostics.contains("subroutine name"), ...);
   ```
   That's a `Debug`-format substring match. Brittle if the error type's `Debug` formatting changes (e.g., adds prefix like "ParseError::Unexpected(..."). Consider asserting on a structured field:
   ```rust
   let errors = parser.get_errors();
   assert!(errors.iter().any(|e| matches!(e, ParseError::Unexpected { expected, .. } if expected == "subroutine name")));
   ```
   Stronger contract. Non-blocking.

4. **`sub_node.location.start = decl_token.start;`** is preserved from existing code -- this line predates the fix. It correctly anchors the sub's source span to the `my`/`our`/`state` keyword, not to the `sub` keyword. Good.

5. **No test for `our sub ;` and `state sub ;`** -- the recovery should work for all three (my/our/state), but only `my sub ;` is tested. Worth parameterizing or adding two more tests.

1. **Use `sub_node.location.end` for the error position** (concern #2). Cleaner UX.
2. **Strengthen test assertions to structured matching** (concern #3).
3. **Add `our sub ;` and `state sub ;` recovery tests** (concern #5).
4. **Consider also handling `my sub` followed by `(` (signature) without a name**:
   ```rust
   my sub ($x, $y) { ... }    # malformed: anonymous scoped sub doesn't make sense
   ```
   Currently this would parse as anon-sub-as-statement, likely causing weird downstream issues. Out of scope for this PR.

**Verdict**: real parser-recovery improvement that turns silent corruption into a structured diagnostic. The discriminator (`name.is_none()` after `parse_subroutine` returns) is the right one; tests pin both success and recovery paths. Worth a couple polish items (error position, structured assertion, our/state coverage).

EffortlessSteven:

EffortlessSteven:

PR #6009, opened 2026-04-24
fix(parser-core): accept leading :: in package declarations (#6686)

EffortlessSteven:
### Motivation
- Package declarations that begin with the main-package qualifier (`::Name`) were being parsed as errors, which broke recovery and AST shape for realistic package declaration edge forms.

### Description
- Teach `parse_package` to accept an optional leading `::` (both `DoubleColon` and `Colon`+`Colon` tokenizations) before calling `parse_qualified_name` and preserve the prefix in the produced `Package` node span.
- Add regression tests exercising `package ::My::App;` and `package ::My::App:: { ... }` to lock in correct AST shape and clean-parse behavior.
- Changes touch `crates/perl-parser-core/src/engine/parser/declarations.rs` and `crates/perl-parser-core/tests/parser_tests.rs` only.

### Testing
- Ran `cargo test -p perl-parser-core --test parser_tests`, and the new parser tests passed.
- Ran the full `perl-parser-core` unit test suite which revealed a pre-existing unrelated failure in `test_edge_cases_deep::test_deref_hash_subscript_regex_key` (tests otherwise ran to completion during verification). 
- Ran `cargo test -p perl-parser --lib` which showed a pre-existing unrelated failure in `refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`.

------

### DIFF
--- a/crates/perl-symbol/src/surface/mod.rs
+++ b/crates/perl-symbol/src/surface/mod.rs
@@ -30,6 +30,7 @@
 pub mod decl;
 pub mod facts;
 pub mod r#ref;
+pub mod ref_facts;
 
 pub use decl::{SymbolDecl, extract_symbol_decls};
 pub use facts::{
@@ -37,3 +38,4 @@ pub use facts::{
     symbol_refs_to_semantic_facts,
 };
 pub use r#ref::{SymbolRef, SymbolRefKind, extract_symbol_refs};
+pub use ref_facts::{SymbolRefSemanticFacts, symbol_refs_to_semantic_facts};
--- a/crates/perl-symbol/src/surface/ref_facts.rs
+++ b/crates/perl-symbol/src/surface/ref_facts.rs
@@ -1,0 +1,136 @@
+use crate::surface::r#ref::{SymbolRef, SymbolRefKind};
+use perl_semantic_facts::{
+    AnchorFact, AnchorId, Confidence, EdgeFact, EdgeId, EdgeKind, EntityId, FileId, OccurrenceFact,
+    OccurrenceId, OccurrenceKind, Provenance,
+};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SymbolRefSemanticFacts {
+    pub anchors: Vec<AnchorFact>,
+    pub occurrences: Vec<OccurrenceFact>,
+    pub references_edges: Vec<EdgeFact>,
+}
+
+pub fn symbol_refs_to_semantic_facts(
+    refs: &[SymbolRef],
+    file_id: FileId,
+    resolve_entity_id: impl Fn(&SymbolRef) -> Option<EntityId>,
+) -> SymbolRefSemanticFacts {
+    let mut anchors = Vec::with_capacity(refs.len());
+    let mut occurrences = Vec::with_capacity(refs.len());
+    let mut references_edges = Vec::new();
+
+    for symbol_ref in refs {
+        let anchor_span = symbol_ref.anchor_span.unwrap_or(symbol_ref.full_span);
+        let anchor_id = AnchorId(stable_id("ref_anchor", &symbol_ref.qualified_name, anchor_span.0, anchor_span.1));
+        anchors.push(AnchorFact {
+            id: anchor_id,
+            file_id,
+            span_start_byte: anchor_span.0 as u32,
+            span_end_byte: anchor_span.1 as u32,
+            scope_id: None,
+            provenance: Provenance::ExactAst,
+            confidence: Confidence::High,
+        });
+
+        let entity_id = resolve_entity_id(symbol_ref);
+        let occurrence_id = OccurrenceId(stable_id(
+            "ref_occurrence",
+            &symbol_ref.qualified_name,
+            symbol_ref.full_span.0,
+            symbol_ref.full_span.1,
+        ));
+        occurrences.push(OccurrenceFact {
+            id: occurrence_id,
+            kind: occurrence_kind(symbol_ref),
+            entity_id,
+            anchor_id,
+            scope_id: None,
+            provenance: Provenance::ExactAst,
+            confidence: Confidence::High,
+        });
+
+        if let Some(to_entity_id) = entity_id {
+            let from_entity_id = EntityId(stable_id(
+                "ref_source_entity",
+                &symbol_ref.qualified_name,
+                symbol_ref.full_span.0,
+                symbol_ref.full_span.1,
+            ));
+            references_edges.push(EdgeFact {
+                id: EdgeId(stable_id("references_edge", &symbol_ref.qualified_name, from_entity_id.0 as usize, to_entity_id.0 as usize)),
+                kind: EdgeKind::References,
+                from_entity_id,
+                to_entity_id,
+                via_occurrence_id: Some(occurrence_id),
+                provenance: Provenance::ExactAst,
+                confidence: Confidence::High,
+            });
+        }
+    }
+
+    SymbolRefSemanticFacts { anchors, occurrences, references_edges }
+}
+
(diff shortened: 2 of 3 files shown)


EffortlessSteven:
Standards review findings:

**Issue found**: Two expect(TokenKind::Number) calls in declarations.rs parse_use_pragma (line 526 and elsewhere). While guarded by prior peek() checks, the coding standard forbids all expect() in production code without exception.

**Fix required**: Replace with consume_token()? since the guard ensures the token is present.

This is a fix-forward candidate — the logic is safe but violates the style standard. Should be corrected before deep-review.

EffortlessSteven:
Fixed: replaced `self.expect(TokenKind::Number)` with `self.consume_token()` at line 544 in declarations.rs. The prior `peek_kind()` guard ensures the token is present and of the correct kind, so `consume_token()` is safe and aligns with codebase pattern. Cargo check and test suite pass. Commit: 00235af34

EffortlessSteven:
Verified: Fix applied in commit 00235af34. Replaced expect(TokenKind::Number) with consume_token() as flagged by standards review. All parser-core lib tests pass (564 passed). Removed needs-builder-fix label. Ready for next stage.

EffortlessSteven:
### Review: accept leading :: in package declarations

**Real Perl-syntax gap closed.** `package ::My::App;` and `package ::Foo;` are valid Perl -- the leading `::` explicitly anchors the package name to the global root namespace, equivalent to `main::My::App` (since `::` at the start of a name resolves to `main::`). This pattern shows up in:

- Defensive code that wants to avoid ambiguity with nested packages.
- Generated code (e.g., `Module::Build`, `Inline::C`) that prefixes names with `::` for safety.
- Older CPAN modules from the Perl 5.6/5.8 era.

Master previously rejected `package ::My::App;` with a parse error.

**The implementation has the right shape**:

```rust
let leading_main_start = if self.peek_kind() == Some(TokenKind::DoubleColon) {
    let token = self.consume_token()?;
    Some(token.start)
} else if self.peek_kind() == Some(TokenKind::Colon)
    && self.tokens.peek_second().map(|t| t.kind) == Ok(TokenKind::Colon)
{
    let first = self.consume_token()?;
    self.consume_token()?;
    Some(first.start)
} else {
    None
};

let (mut name, mut name_span) = self.parse_qualified_name(true)?;
if let Some(prefix_start) = leading_main_start {
    name.insert_str(0, "::");
    name_span.start = prefix_start;
}
```

Three correct moves:

1. **Handle both lexer encodings**: `TokenKind::DoubleColon` (one token) *and* two `TokenKind::Colon` tokens (split). This is necessary because the lexer is context-sensitive and may emit `::` either way depending on surrounding tokens. Future-proof.

2. **Preserve source location**: `name_span.start = prefix_start` captures the position of the leading `::` so go-to-definition and rename point at the right place.

3. **Prepend `::` to the parsed name**: `name.insert_str(0, "::")` -- the AST `Package { name: "::My::App" }` carries the leading marker explicitly. Downstream consumers can decide whether to treat it as `main::My::App` or as a literal `::`-prefixed name. Keeps semantics in the AST instead of in the parser.

**Tests pin both forms**:

| Test | Input | Expected |
|---|---|---|
| `parse_package_declaration_with_leading_separator` | `package ::My::App;` | `name == "::My::App"`, no block |
| `parse_package_declaration_with_leading_and_trailing_separator_block` | `package ::My::App:: { ... }` | `name == "::My::App::"`, has block |

The second test is the boundary case -- both leading *and* trailing `::` plus a block form. Confirms the existing `parse_qualified_name(allow_trailing_separator=true)` interaction works with the new prefix logic. Good.

**Significant concerns**:

1. **Cosmetic re-flow of unrelated code.** The diff reformats large chunks of `parse_use_or_require` (the `use My::App v5.36` parser) into different vertical layouts. None of this is related to the leading-`::` fix:

   ```rust
   // Before:
   let mut module = if matches!(
       self.peek_kind(),
       Some(TokenKind::Number) | Some(TokenKind::VString)
   ) {
       ...
   };

   // After (no logic change):
   let mut module =
       if matches!(self.peek_kind(), Some(TokenKind::Number) | Some(TokenKind::VString)) {
           ...
       };
   ```

   And re-indentation of comment-style trailing comments:
   ```rust
   self.consume_token()?; // consume comma
   - // Continue to parse next argument
   + // Continue to parse next argument        // (now reads as part of the comment chain)
   ```

   Per `AGENTS.md`: *"Do NOT rewrite files you did not need to touch -- creates spurious review surface"*. These reformatting hunks should be:
   - **Reverted** before merge (preferred), or
   - **Split** into a separate `style: cargo-fmt-pass` PR.

   The unrelated reformatting accounts for ~80 of the 141 added/66 removed lines. The actual fix is ~40 lines. Reviewers shouldn't have to skim through `cargo fmt` noise to find the substantive change.

2. **Test renames** in `parser_tests.rs`:
   ```rust
   - fn parse_package_declaration_with_trailing_separator_and_block()
   - -> Result<(), Box<dyn std::error::Error>> {
   + fn parse_package_declaration_with_trailing_separator_and_block(
   + ) -> Result<(), Box<dyn std::error::Error>> {
   ```
   Pure cosmetic re-flow. Same comment as above -- shouldn't be in this PR.

3. **`peek_second().map(|t| t.kind) == Ok(TokenKind::Colon)`** -- this assumes `peek_second` returns `Result<Token>`. If it returns `Result<&Token>` or `Option<Token>`, the comparison would need adjustment. Worth confirming the surrounding lexer API. Pattern suggests `Result`, which is unusual for a `peek_second` (typically `Option`). If `peek_second` errors when at EOF, this branch correctly returns `None` (the `Ok(...) ==` comparison fails for EOF-error case). OK.

4. **`name.insert_str(0, "::")`** allocates if the inner buffer wasn't pre-sized. For a typical "::Foo::Bar" name, the `String` capacity from `parse_qualified_name` is probably 7-15 bytes, prepending "::" requires shifting. Negligible cost. Fine.

5. **`name_span.start = prefix_start`** overwrites a `mut` field. If `name_span` is shared/cloned anywhere downstream, this mutates *only* the local copy. OK, but worth confirming the `parse_qualified_name` return value semantics.

6. **Edge cases not tested**:
   - `package :: ;` (just `::` with no following name -- should error or accept as `name = "::"`?)
   - `package ::Foo v5.38;` (leading `::` *with* version -- the version-token check is downstream of name parsing, but the version token's interaction with `name_span.start` may be subtle)
   - `package ::main::Foo;` (redundant leading `::` to `main::`-prefixed name)

**Suggestions**:

1. **Revert all `cargo fmt`-only hunks** in `declarations.rs` (concern #1). The PR should be ~40 lines of substance, not 207. If the master branch is genuinely out of fmt-compliance, file a separate `style:` PR.

2. **Revert the test rename** in `parse_package_declaration_with_trailing_separator_and_block` (concern #2). Style-only change, doesn't belong here.

3. **Add the `package ::Foo v5.38;` test** (concern #6) -- this is the interaction between this PR (#6009) and #6007 (v-string versions). If both land, this case should work; tests should pin it.

4. **Consider adding a comment** above the `leading_main_start` block:
   ```rust
   // Handle leading `::` in package declarations: `package ::My::App;`
   // resolves to the same package as `main::My::App` but is explicit about
   // anchoring at the global root namespace.
   ```

**Verdict**: real Perl-syntax gap fix with the right approach (capture both lexer encodings, preserve source location, prepend marker to name). However, the PR is **bloated by ~80 lines of unrelated `cargo fmt` reflow** that should be split into a separate style PR. Once that's cleaned up, this is a small, surgical fix worth merging.

**Blocking ask**: split or revert the unrelated reformatting.

CI failures on this branch are stale (2+ days). Recommend a rebase + push to refresh CI after the cleanup.

EffortlessSteven:

EffortlessSteven:

Issue #5593, opened 2026-04-23
[MASTER BREAK] perl-module Windows 8.3 short-path vs long-path canonicalization mismatch


EffortlessSteven:
Windows Guardrails (module-separator-regressions) lane shows 18 test failures in `crates/perl-module/tests/module_resolution_use_require.rs` due to Windows 8.3 short-path (`C:\Users\RUNNER~1\...`) vs long-path (`C:\Users\runneradmin\...`) canonicalization mismatch.

## Root Cause

Windows `fs::canonicalize` may return paths in 8.3 short form (e.g., `C:\Users\RUNNER~1\...`), while test assertions compare against long-form paths constructed from `tempfile::TempDir::path()`.

**Example failure**:
- Expected: `C:\Users\runneradmin\...` (long form from test construction)
- Got: `C:\Users\RUNNER~1\...` (8.3 short form from `resolve_module_path()`)

Both refer to the same file, but byte-for-byte comparison fails.

## Solution Approach

Normalize all returned module paths to long-form equivalents using Windows API `GetLongPathName` before returning from `resolve_module_path()`. This ensures paths are consistently in long form regardless of how the filesystem returns them.

Wrap the Windows-specific call in a helper function in `perl-module/src/path/mod.rs` and call it at the two return points in `resolve_module_path()`.

### Step 1: Add helper function in `crates/perl-module/src/path/mod.rs`

Create a new public function `canonicalize_path_long_form(path: &Path) -> PathBuf`:
- **On Windows**: Use `winapi::um::fileapi::GetLongPathNameW` via unsafe block to expand 8.3 short names to long form
  - Allocate UTF-16 buffer for input path
  - Call GetLongPathNameW with appropriate buffer size
  - Convert result back to Rust PathBuf
  - If the path doesn't exist, return it unchanged (GetLongPathName only works on existing paths on some Windows versions)
- **On non-Windows**: Return path unchanged (no-op on Linux/macOS)

### Step 2: Update `crates/perl-module/src/resolution/path.rs`

Apply normalization at both return points:

**Line 45-46** (return when match found):
```rust
if safe_candidate.exists() {
    return Some(crate::path::canonicalize_path_long_form(&safe_candidate));
}
```

**Line 50** (fallback return):
```rust
Some(crate::path::canonicalize_path_long_form(&root.join("lib").join(relative_path)))
```

### Step 3: Update `crates/perl-module/Cargo.toml`

[...]

PR #6035, opened 2026-04-24

EffortlessSteven:
### Motivation
- Module resolution could return Windows 8.3 (short) paths in some cases and long-form paths in others, causing path mismatches in guardrail checks and downstream consumers.

### Description
- Added `canonicalize_path_long_form(path: &Path) -> PathBuf` to `crates/perl-module/src/path/mod.rs` with a Windows implementation that calls `GetLongPathNameW` and a non-Windows no-op that returns the input path unchanged.
- Normalized all return sites in `resolve_module_path` so both matched include-path results and the `root/lib/...` fallback are returned via `canonicalize_path_long_form` for consistent path form on Windows.
- Added a Windows-target dependency on `windows-sys` with the `Win32_Storage_FileSystem` feature in `crates/perl-module/Cargo.toml`.
- Updated integration/unit tests to compare against canonicalized expected paths and added a small test ensuring nonexistent paths are left unchanged by the helper.

### Testing
- Ran `cargo test -p perl-module`, which completed successfully (all tests passed).
- Ran formatting via `cargo xtask fmt` / `cargo fmt`, which completed successfully.

------

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1724,6 +1724,7 @@ dependencies = [
  "proptest",
  "tempfile",
  "url",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/crates/perl-module/Cargo.toml
+++ b/crates/perl-module/Cargo.toml
@@ -12,6 +12,9 @@ url = { workspace = true }
 perl-parser-core = { workspace = true }
 perl-workspace = { workspace = true }
 
+[target.'cfg(windows)'.dependencies]
+windows-sys = { version = "0.59", features = ["Win32_Storage_FileSystem"] }
+
 [dev-dependencies]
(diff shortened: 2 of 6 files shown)


EffortlessSteven:
Closing as superseded by #5986. That PR (merged earlier today at master SHA d88271b4) fixes the Windows module-separator canonicalization problem from the opposite direction — it REMOVES canonicalization from resolve_module_path returns (returns caller-supplied path) rather than ADDING long-form canonicalization. The original issue was that canonicalize expands 8.3 short names (RUNNER~1 -> runneradmin) making the result inconsistent with caller expectations. Merging #6035 would re-introduce the inconsistency #5986 solved. If specific path normalization gaps remain, please file a new issue that references #5986's current master approach.

PR #6036, opened 2026-04-24

EffortlessSteven:
### Motivation

- Prevent mismatches between 8.3 short-path and long-path Windows names that broke module-resolution guardrails by normalizing returned module filesystem paths.

### Description

- Added `canonicalize_path_long_form(path: &Path) -> PathBuf` in `crates/perl-module/src/path/mod.rs` which is a no-op on non-Windows and uses `GetLongPathNameW` on Windows to return long-form paths.
- Updated resolver return sites in `crates/perl-module/src/resolution/path.rs` to call `canonicalize_path_long_form` for both found include-path candidates and the `root/lib/<module>.pm` fallback so all returned paths are normalized.
- Added a Windows-only dependency under `cfg(windows)` on `windows-sys` (feature `Win32_Storage_FileSystem`) in `crates/perl-module/Cargo.toml` and captured the lockfile change.
- Added/adjusted targeted tests in `crates/perl-module/tests/` to assert the resolver returns the canonicalized path and to verify the helper is a no-op on non-Windows via an added test using `canonicalize_path_long_form`.

### Testing

- Ran `cargo test -p perl-module` and all tests passed.
- Ran `cargo clippy -p perl-module` and it completed without errors.
- Ran formatting via `cargo xtask fmt` locally; `just pr-fast` was not available in this environment so it was not executed.

------

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1724,6 +1724,7 @@ dependencies = [
  "proptest",
  "tempfile",
  "url",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/crates/perl-module/Cargo.toml
+++ b/crates/perl-module/Cargo.toml
@@ -12,6 +12,9 @@ url = { workspace = true }
 perl-parser-core = { workspace = true }
 perl-workspace = { workspace = true }
 
+[target.'cfg(windows)'.dependencies]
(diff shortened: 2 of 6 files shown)


EffortlessSteven:

PR #6037, opened 2026-04-24

EffortlessSteven:
### Motivation

- Windows 8.3 short-path vs long-path mismatches caused inconsistent filesystem paths from module resolution and broke guardrail checks, so resolved module paths must be normalized to a consistent long form.

### Description

- Add `canonicalize_path_long_form(path: &Path) -> PathBuf` in `crates/perl-module/src/path/mod.rs` that is a no-op on non-Windows and calls `GetLongPathNameW` on Windows.
- Use `canonicalize_path_long_form` at all return sites in `resolve_module_path` so resolved candidates and the `root/lib/...` fallback are consistently normalized (`crates/perl-module/src/resolution/path.rs`).
- Wire `windows-sys` as a `cfg(windows)` dependency for `perl-module` and add a workspace-level `windows-sys` entry to the top-level `Cargo.toml` so the Win32 API is available only on Windows (`crates/perl-module/Cargo.toml`, `Cargo.toml`).
- Update existing integration assertions to compare against normalized paths and add a Windows-only test `module_resolution_windows_long_path.rs` that creates a file, obtains its 8.3 short form via `GetShortPathNameW`, and verifies `canonicalize_path_long_form` expands it back.

### Testing

- Ran `cargo test -p perl-module`, which passed (module crate tests all OK).
- Ran `cargo xtask fmt` and formatting completed.
- Ran `cargo clippy -p perl-module --all-targets -- -D warnings`, which failed due to an existing clippy warning (a `collapsible-if`) in an unrelated test file; the failure is pre-existing and unrelated to the Windows helper changes.

------

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1724,6 +1724,7 @@ dependencies = [
  "proptest",
  "tempfile",
  "url",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -321,6 +321,7 @@ clap = { version = "4.5.60", features = ["derive"] }
 tokio = { version = "1.51.1", features = ["net", "rt-multi-thread", "macros", "io-util", "sync", "time"] }
 proptest = "1.11.0"
 criterion = "0.8.1"
+windows-sys = "0.61.2"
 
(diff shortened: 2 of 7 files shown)


EffortlessSteven:

PR #6038, opened 2026-04-24

EffortlessSteven:
### Motivation

- Windows 8.3 short-path vs long-path mismatches caused module-resolution comparisons to diverge and break guardrail checks.
- The intent is to normalize returned module filesystem paths to a stable long-form representation on Windows while leaving other platforms unchanged.

### Description

- Added a Windows-target helper `canonicalize_path_long_form(path: &Path) -> PathBuf` which calls `GetLongPathNameW` and falls back to the original path on failure or when not on Windows.
- Applied `canonicalize_path_long_form` at all return sites in `resolve_module_path` so both include-path matches and the `root/lib/...` fallback are consistently normalized.
- Added `windows-sys` as a `cfg(windows)` dependency with the `Win32_Storage_FileSystem` feature to provide `GetLongPathNameW` only on Windows.
- Added targeted tests: two unit tests that assert no-op/stable behavior on non-Windows and a Windows-only integration test that exercises short-path -> long-path normalization when available.

### Testing

- Ran `cargo test -p perl-module` and all tests in the crate passed.
- Ran `cargo check --all-targets -p perl-module` and it completed successfully.
- Ran `cargo clippy -p perl-module` which completed, and note that running clippy with `-D warnings` surfaced an unrelated collapsible-if lint in a test which is not part of this change; `just pr-fast` was not available in this environment (`just: command not found`).

------

### DIFF
--- a/crates/perl-module/Cargo.toml
+++ b/crates/perl-module/Cargo.toml
@@ -12,6 +12,9 @@ url = { workspace = true }
 perl-parser-core = { workspace = true }
 perl-workspace = { workspace = true }
 
+[target.'cfg(windows)'.dependencies]
+windows-sys = { version = "0.59.0", features = ["Win32_Storage_FileSystem"] }
+
 [dev-dependencies]
 proptest = { workspace = true }
 tempfile = { workspace = true }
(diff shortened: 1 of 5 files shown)


EffortlessSteven:

Issue #3522, opened 2026-04-10
[workspace] Workspace-wide refactoring operations not supported


EffortlessSteven:
# [workspace] workspace-wide refactoring operations not supported

## Problem

perl-lsp supports local refactoring (extract variable, extract subroutine) but lacks workspace-wide refactoring operations that modify multiple files. Users cannot:
1. Rename a symbol across the entire workspace
2. Move a module to a new location with import updates
3. Extract a subroutine to a different module
4. Organize imports across multiple files

## Current State

Local refactoring exists:
- Extract variable (within function)
- Extract subroutine (within file)
- Inline variable (within file)

Missing workspace-wide operations:
- Rename symbol (cross-file, cross-folder)
- Move module with import updates
- Safe delete (check references across workspace)
- Organize imports (multi-file cleanup)

### Rename symbol across workspace
```perl
# In lib/OldName.pm
package OldName;
sub process { ... }

# In bin/script.pl  (different file, maybe different folder)
use OldName;
OldName::process();  # Should update on rename
```

When renaming `OldName` to `NewName`, all references across the workspace should update.

### Move module
```
lib/Internal/Util.pm  →  lib/Public/Utils.pm
```
All `use Internal::Util` statements across the workspace should update to `use Public::Utils`.

## Technical Challenges

1. **Workspace edit scope**: LSP `WorkspaceEdit` supports multiple changes, but server must compute them
2. **Reference finding**: Need accurate cross-file reference detection
3. **Undo support**: Atomic workspace edits or proper undo grouping
4. **Conflict detection**: Handle files that have changed since edit computation
5. **Multi-root**: Works across workspace folders (see cross-folder navigation issue)

### Phase 1: Cross-file rename
- Find all references to symbol across workspace (see cross-folder navigation)
- Generate `WorkspaceEdit` with all file modifications
- Apply atomically via LSP client

### Phase 2: Move refactoring
- Detect imports of moved module
- Compute all import statement updates
- Optionally update fully-qualified references

### Phase 3: Safe delete
- Check for references before allowing deletion
- Warn if references exist in other files

## Code Locations

- `crates/perl-lsp-code-actions/src/` — existing local refactoring
- `crates/perl-lsp-rename/src/` — rename infrastructure
- `crates/perl-workspace-index/src/` — reference finding
- `crates/perl-lsp/src/runtime/workspace.rs` — workspace edit application

## Acceptance Criteria

- [ ] Rename symbol works across multiple files
[...]

PR #6053, opened 2026-04-24
feat(rename): add first workspace-wide multi-file rename slice (#3522)

EffortlessSteven:
### Motivation

- Implement a guarded, workspace-wide rename path that can produce multi-file `WorkspaceEdit`s for statically-resolved symbols and that cleanly refuses when symbol identity is weak or ambiguous.

### Description

- Add `RenameRefusal` enum and change `build_rename_edit` to return `Result<Vec<RenameEdit>, RenameRefusal>` and restrict workspace rename to statically-resolved subs, refusing `main`-package and unresolved definitions; detection logic implemented in `is_ambiguous_sub_reference` (file: `crates/perl-lsp-rs/src/features/workspace_rename.rs`).
- Preserve declaration + all cross-file references via the existing `WorkspaceIndex` search and produce `RenameEdit` -> LSP `WorkspaceEdit` via `to_workspace_edit` (file: `crates/perl-lsp-rs/src/features/workspace_rename.rs`).
- Propagate refusal results through the LSP rename path by mapping `RenameRefusal` into a JSON-RPC invalid-params error (`-32602`) in `handle_rename_workspace`, while preserving same-file fallback behavior for partial/degraded index modes (file: `crates/perl-lsp-rs/src/runtime/language/rename.rs`).
- Add focused tests: unit tests exercising workspace rename behavior and ambiguity refusal (`workspace_rename::tests`), and LSP-level tests `test_workspace_rename_returns_multi_file_workspace_edit` and `test_workspace_rename_refuses_ambiguous_symbol_identity` (file: `crates/perl-lsp-rs/tests/lsp_rename_tests.rs`).

### Testing

- `cargo fmt --all` was run and succeeded.
- Added unit and LSP tests (see `workspace_rename` tests and `lsp_rename_tests`), but a full `cargo test -p perl-lsp-rs` run in this environment was interrupted by an out-of-disk situation and package-cache / build lock contention, so targeted runs could not complete here; the new tests are included and should be executed on CI with `cargo test -p perl-lsp-rs` and `cargo test -p perl-workspace-index` to verify cross-file rename and refusal behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb0045b994833394e29c0fb205c9b3)

### DIFF
--- a/crates/perl-lsp-rs/src/features/workspace_rename.rs
+++ b/crates/perl-lsp-rs/src/features/workspace_rename.rs
@@ -5,6 +5,7 @@
 use perl_parser::workspace_index::{SymKind, SymbolKey, WorkspaceIndex};
 use serde_json::{Value, json};
 use std::collections::BTreeMap;
+use std::fmt;
 
 /// Represents a text edit for a single document
 #[derive(Debug, Clone)]
@@ -26,6 +27,38 @@ pub struct RenameEdit {
     pub edits: Vec<TextEdit>,
 }
 
+/// Reasons why workspace rename is refused.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RenameRefusal {
+    /// Symbol kind is not in the supported static workspace-rename slice.
+    UnsupportedSymbolKind,
+    /// Workspace index could not produce a stable symbol identity.
+    WeakIdentity(String),
+    /// Workspace index produced references that cannot be safely attributed.
+    AmbiguousIdentity(String),
+}
+
+impl fmt::Display for RenameRefusal {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::UnsupportedSymbolKind => {
+                write!(
+                    f,
+                    "Workspace rename currently supports statically resolved subroutines only"
+                )
+            }
+            Self::WeakIdentity(reason) => {
+                write!(f, "Workspace rename refused: weak symbol identity ({reason})")
+            }
+            Self::AmbiguousIdentity(reason) => {
+                write!(f, "Workspace rename refused: ambiguous symbol identity ({reason})")
+            }
+        }
+    }
+}
+
+impl std::error::Error for RenameRefusal {}
+
 /// Build a rename edit across the workspace.
 ///
 /// Finds all references to the given symbol and builds text edits to rename them.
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
## Deep Review — reviewer-deep

**SEND-BACK** — structural logic errors require builder attention

### Issue 1: Ambiguity refusal test will fail at runtime (critical)

The test `rename_refuses_ambiguous_unqualified_cross_package_call` and LSP-level test `test_workspace_rename_refuses_ambiguous_symbol_identity` both expect `Err(AmbiguousIdentity)` for this scenario:

```perl
# Foo.pm
package Foo;
sub process_data { return 1; }

# Bar.pm  
package Bar;
sub run { return process_data(); }  # unqualified call
```

But the scenario cannot trigger the refusal. Here's why:

`find_refs(key = Foo::process_data)` searches for the string `"Foo::process_data"` in the reference map (see `workspace_index.rs:2791`). When Bar.pm indexes the bare call `process_data()`, the indexer stores it as `"process_data"` (bare) AND `"Bar::process_data"` (qualified, using `current_package = Bar`) — see `workspace_index.rs:3127-3134`. Neither key is `"Foo::process_data"`, so `find_refs` returns nothing for Bar.pm.

Result: `build_rename_edit` iterates only over the definition location in Foo.pm. `is_ambiguous_sub_reference` is never called. The function returns `Ok([RenameEdit { uri: "Foo.pm", ... }])`. The test's `expect_err` assertion fails.

**`is_ambiguous_sub_reference` is dead code in the current design** — it can only fire for locations returned by `find_refs`, but those locations always contain `::` (because non-main subs are indexed under qualified names), so the `contains("::") → return false` guard always fires first.

### Issue 2: Secondary logic bug in is_ambiguous_sub_reference

The `original.starts_with('&')` early return is logically wrong. An unqualified `&bare_name` call in a different package is as ambiguous as `bare_name()`. This guard incorrectly exempts it. (Moot given Issue 1, but wrong regardless.)

### Issue 3: IndexAccessMode::Partial refusal propagation inconsistency

In `Partial` mode, a `RenameRefusal` is returned immediately as a JSON-RPC error (`-32602`). In degraded-index scenarios, it would be more appropriate to fall through to same-file rename (matching how `Ok(empty)` is handled in Partial mode). In `Full` mode, immediate error is correct because the index is authoritative.

### Recommended builder fix

The ambiguity refusal tests need to be redesigned around a scenario the indexer actually produces. Options:

**Option A**: Test with a qualified cross-package reference that the indexer doesn't expect:
```perl
# Foo.pm: package Foo; sub process_data { ... }
# Bar.pm: package Bar; sub run { return Foo::process_data(); }  # qualified ref
```
In this case, `find_refs(Foo::process_data)` DOES include the Bar.pm location (stored as `"Foo::process_data"`). The rename should succeed (it's unambiguous, just update both files). So this tests the HAPPY path, not a refusal.

**Option B**: Remove `is_ambiguous_sub_reference` as premature optimization. The rename is unambiguous for subs with a resolved definition. The main guard (refusing main-package subs and requiring `find_def` success) is sufficient for safety. Add refusal logic only when the indexer actually produces ambiguous data.

**Option C**: Redesign `is_ambiguous_sub_reference` to work on the caller site based on package context rather than relying on `find_refs` to pre-filter. But this would require a broader index query (all-packages ref scan).

### What works correctly

- `RenameRefusal::UnsupportedSymbolKind` (non-Sub kinds) — correct
- Refusing `main`-package subs — correct  
- Requiring a resolved definition via `find_def` — correct
- `is_non_target_package_declaration` skipping foreign declarations — correct
- Error propagation in `Full` mode — correct

### Architectural invariant

No dependency on `perl-parser-core` added. `RenameRefusal` type design is clean and correctly implements `Display` + `std::error::Error`.

EffortlessSteven:

EffortlessSteven:
Superseded by #7019 — cherry-pick recovery onto current master. The original branch had no shared ancestry with master (fresh-root divergence). API adaptation required: build_rename_edit now returns Result<Vec<RenameEdit>, RenameRefusal> in the cherry-pick vs Vec<RenameEdit> on master; rename.rs updated to handle the Result.

PR #6065, opened 2026-04-24
refactor(completion): plumb effective include roots into module completion (#4314)

EffortlessSteven:
### Motivation

- Module-name completion (`use`/`require`) needs doc-scoped include path context (workspace config + `PERL5LIB`) and optional system `@INC` to later enable effective module lookups without introducing filesystem scanning yet.
- Threading those vectors through the completion provider ensures the runtime can supply per-document include/system roots while keeping other completion behavior unchanged.

### Description

- Added `LspServer::completion_include_context` which fetches the document-specific `WorkspaceConfig`, computes `effective_include_paths` (respecting `PERL5LIB` precedence), and returns system `@INC` paths only when `use_system_inc` is enabled.
- Extended `CompletionProvider` with `include_paths: Vec<String>` and `system_inc_paths: Vec<String>` and added `new_with_index_source_and_inc(...)` while keeping `new_with_index_and_source(...)` delegating to it with empty vectors to preserve default behavior.
- Updated the LSP completion handler to call `completion_include_context` and construct the provider with the include/system vectors, and threaded those vectors through to the module completion callsite only (`add_use_module_completions`).
- Adjusted `add_use_module_completions` signature to accept the include/system path slices; the implementation is currently a no-op consumer (no filesystem scanning or resolution added yet).

### Testing

- Ran `cargo check -p perl-lsp-rs-core -p perl-lsp-rs` which completed successfully.
- Ran `cargo check --workspace --all-targets` which completed successfully.
- Ran `cargo test -p perl-lsp-rs-core` which executed the suite but one pre-existing test failed (a test expecting `crates/perl-lsp/Cargo.toml` that is not present in this environment), so not caused by these changes.
- Ran `cargo test -p perl-lsp-rs` which did not complete due to environment/linker issues (linker bus error / `No space left on device`) unrelated to the code changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb003617988333afec4189fc3d127f)

PR #6068, opened 2026-04-24
refactor(completion): plumb effective include roots into module completion (#4314)

EffortlessSteven:
### Motivation

- Ensure module-name completion has access to the document-scoped include roots and optional system `@INC` so future module-resolution and filesystem-backed completions can use accurate search scopes.

### Description

- In the LSP completion handler, load the document's effective `WorkspaceConfig`, compute `include_paths` (merging `PERL5LIB`) and collect `system_inc_paths` only when `use_system_inc` is enabled, then pass those vectors into provider creation for both normal and cancellable completion paths.
- Extend `CompletionProvider` with `include_paths` and `system_inc_paths` fields and add `new_with_index_source_and_paths` constructor while keeping existing constructors backward-compatible (empty vectors by default).
- Thread the include/system vectors into the module-name completion path by adding parameters to `add_use_module_completions` and passing the provider-held vectors into that call site (no filesystem scanning or resolution logic added yet).
- Add a focused unit test that constructs a `CompletionProvider` with explicit include and system vectors and verifies module completions still work (plumbing validation for phase 1).

### Testing

- Ran `cargo check --workspace --all-targets` which passed.
- Ran `cargo test -p perl-lsp-rs-core` which could not fully pass in this environment due to a pre-existing test expectation around a missing `crates/perl-lsp/Cargo.toml` (unrelated to this change); the new unit test added for include vectors passed in local test runs that exercised the provider constructor.
- Ran `cargo test -p perl-lsp-rs` which failed in this environment due to linker/disk exhaustion during test builds (environmental), not due to logic changes.
- Performed `cargo clean` to recover disk space so checks could complete after cleanup.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb003617988333afec4189fc3d127f)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion.rs
@@ -170,6 +170,8 @@ pub struct CompletionProvider {
     type_engine: Option<TypeInferenceEngine>,
     workspace_index: Option<Arc<WorkspaceIndex>>,
     import_map: ImportMap,
+    include_paths: Vec<String>,
+    system_inc_paths: Vec<String>,
 }
 
 impl CompletionProvider {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6065 implements the same feature-plumbing scope with cleaner type choices and constructor delegation. All 4 variants (6068, 6067, 6066, 6065) are identical 4-shot outputs from the same Codex task; winner uses `Vec<String>` paths (better for LSP runtime context) and cleaner constructor chaining pattern. No novel edge cases to extract — all implementations are phase-1 plumbing only (no filesystem scanning or resolution logic yet). Thank you for the submission.

PR #6093, opened 2026-04-24
fix(lexer): harden heredoc and data-section scanning under edge cases

EffortlessSteven:
### Motivation

- Improve heredoc correctness for CR / CRLF edge cases, backtick heredocs, `<<~` indentation, quoted vs unquoted labels, and to keep slow-paths bounded for pathological inputs.
- Ensure `__DATA__`/`__END__` transitions are robust across CR-only lines and do not cause runaway scanning while preserving existing time/depth budgets.

### Description

- Add `parse_quoted_heredoc_label` helper to strictly parse quoted/backtick heredoc labels and require the closing quote to appear on the same line, returning `None` for unterminated labels so they fallback to non-heredoc parsing.
- Replace inline quoted-label scanning with the new helper for `"..."`, `'...'`, and `` `...` ``, and store quoted labels as `Arc<str>` in `pending_heredocs` for consistent ownership.
- Simplify `find_line_end` return semantics and make `__DATA__/__END__` marker consumption use the CR/LF-aware `find_line_end` + `consume_newline` path to correctly handle CR, CRLF and LF variants.
- Add targeted tests for unterminated quoted/backtick labels, indented-terminator rules for unquoted heredocs, `<<~` + CRLF terminator behavior, and a bounded CR `__DATA__` transition test.

### Testing

- Ran `cargo test -p perl-lexer heredoc` and `cargo test -p perl-lexer`, both completed successfully with all tests passing.
- Ran `cargo check --workspace --all-targets`, which completed successfully with the workspace checking cleanly.
- Ran formatting via `cargo xtask fmt` successfully; `just pr-fast` was not available in the environment so it was not executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c8d69fc8333938ddccc576defe7)

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -382,13 +382,30 @@ impl<'a> PerlLexer<'a> {
         while end < bytes.len() && bytes[end] != b'\n' && bytes[end] != b'\r' {
             end += 1;
         }
-        // Visible end strips trailing \r if followed by \n
-        let visible_end = if end > start && end > 0 && bytes[end.saturating_sub(1)] == b'\r' {
-            end - 1
-        } else {
-            end
-        };
-        (end, visible_end)
+        (end, end)
+    }
+
+    #[inline]
+    fn parse_quoted_heredoc_label(&mut self, text: &mut String, quote: char) -> Option<Arc<str>> {
+        text.push(quote);
+        self.advance();
+        let mut delim = String::new();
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6095 implements the same heredoc hardening scope. You covered escaped quotes in quoted heredoc labels (a real edge case); extracting that into winner branch as follow-up for escaped-delimiter handling. Thank you.

PR #6095, opened 2026-04-24
fix(lexer): harden heredoc and data-section scanning under edge cases (#6601)

EffortlessSteven:
### Motivation
- Heredoc and data-section scanning had corner cases that could misclassify input or enter slow paths incorrectly for CR-only lines, unterminated quoted/backtick delimiters, and `<<~` indentation variants.

### Description
- Add a shared helper `parse_quoted_heredoc_delimiter` to reject multiline or unterminated quoted heredoc labels so malformed openers degrade gracefully instead of entering heredoc mode. 
- Stop stripping the visible-line-end in `find_line_end` and centralize newline handling via `consume_newline` to make CR / CRLF / LF behavior consistent when scanning markers and bodies. 
- Make `__DATA__`/`__END__` marker consumption robust to `\r`, `\r\n`, and `\n` by checking for both `b'\n'` and `b'\r'` and then calling `consume_newline`, ensuring a clean transition to `InDataSection`. 
- Keep existing safety controls intact (`MAX_HEREDOC_BYTES`, `MAX_HEREDOC_DEPTH`, `HEREDOC_TIMEOUT_MS`) and ensure the slow-path remains bounded. 
- Add focused tests for unterminated backtick-delimiters, CR-only data markers, large CRLF heredoc bodies exceeding the budget, indented-terminator tabs, and exact-match behavior for quoted labels.

### Testing
- Ran formatting with `cargo xtask fmt` (completed successfully). 
- Ran targeted heredoc tests with `cargo test -p perl-lexer heredoc` which passed. 
- Ran the full lexer test suite with `cargo test -p perl-lexer` which passed. 
- Ran `cargo check --workspace --all-targets` which completed successfully (with unrelated warnings elsewhere).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c8d69fc8333938ddccc576defe7)

PR #6096, opened 2026-04-24
fix(lexer): harden heredoc and data-section scanning under edge cases (#6600)

EffortlessSteven:
### Motivation

- Fix heredoc label parsing edge cases that could over-consume input when quoted/backtick delimiters are unterminated or when CR/CRLF line endings occur.
- Keep slow-paths and malformed-input handling bounded (time/bytes/depth) while preserving graceful degradation and existing budget behavior.
- Ensure `__DATA__/__END__` detection and transition to data mode behave correctly across CR, CRLF, and LF line endings.

### Description

- Added a new helper `parse_quoted_heredoc_label` that parses quoted/backtick heredoc labels but is explicitly bounded to the current logical line and returns `None` on unterminated or malformed labels so the lexer falls back cleanly.
- Updated heredoc start handling to use the bounded quoted-label parser and to reset `self.position` and return `None` when a quoted delimiter does not terminate on the same line, preventing runaway scans for malformed quoted labels.
- Changed data-marker consumption to use `find_line_end` + `consume_newline` so marker lines are consumed in a CR/LF-aware way rather than naively scanning to `
`, avoiding incorrect transitions on CR-only or CRLF inputs.
- Added targeted tests covering unterminated quoted/backtick heredoc labels, bounded handling on large malformed inputs, `<<~` tab-indented terminators, and CR-only `__DATA__` transitions (`crates/perl-lexer/src/lib.rs`, updated tests in `crates/perl-lexer/tests/*`).

### Testing

- Ran `cargo test -p perl-lexer heredoc` and the heredoc-focused test suite passed.
- Ran full crate tests with `cargo test -p perl-lexer` and all `perl-lexer` tests passed.
- Ran a workspace-level check with `cargo check --workspace --all-targets` and it completed successfully.
- Formatting via `cargo xtask fmt` was executed as part of verification.

All automated tests mentioned above succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c8d69fc8333938ddccc576defe7)

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -391,6 +391,54 @@ impl<'a> PerlLexer<'a> {
         (end, visible_end)
     }
 
+    /// Parse a quoted heredoc label, bounded to the current logical line.
+    ///
+    /// Returns `Some(label)` only when a matching closing quote is found before
+    /// any newline (`\n`, `\r`, or `\r\n`). Returns `None` for unterminated or
+    /// malformed quoted labels so callers can fall back to non-heredoc parsing.
+    #[inline]
+    fn parse_quoted_heredoc_label(&mut self, quote: char, text: &mut String) -> Option<String> {
+        text.push(quote);
+        self.advance(); // consume opening quote
+
+        let mut delim = String::new();
+        while self.position < self.input.len() {
+            let ch = self.current_char()?;
+
+            // Heredoc labels must terminate on the same line as the `<<` marker.
+            if matches!(ch, '\n' | '\r') {
+                return None;
+            }
+
+            if ch == '\\' {
+                // Keep escaped quote/backslash sequences intact in the raw label.
+                text.push(ch);
+                self.advance();
+
+                if self.position < self.input.len()
+                    && let Some(escaped) = self.current_char()
+                {
+                    delim.push(escaped);
+                    text.push(escaped);
+                    self.advance();
+                }
+                continue;
+            }
+
+            if ch == quote {
+                text.push(quote);
+                self.advance(); // consume closing quote
+                return Some(delim);
+            }
+
+            delim.push(ch);
+            text.push(ch);
+            self.advance();
+        }
+
+        None
+    }
+
     /// Advance the lexer and return the next token.
     ///
     /// Returns `None` only after an `EOF` token has already been emitted.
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
### Meta-review: lexer heredoc cluster (#6094, #6095, #6096)

These three PRs share the title `fix(lexer): harden heredoc and data-section scanning under edge cases`. All touch `crates/perl-lexer/src/lib.rs`, `tests/edge_case_tests.rs`, `tests/heredoc_regression_tests.rs`, and `tests/heredoc_security_tests.rs`. Parallel maintainer attempts.

**Comparison**:

| Aspect | #6094 (+136/-57) | #6095 (+113/-73) | #6096 (+129/-65) |
|---|---|---|---|
| **Helper extraction** | `parse_heredoc_quoted_label(text, quote)` | `parse_quoted_heredoc_delimiter(quote, text)` | `parse_quoted_heredoc_label(quote, text)` |
| **MAX_HEREDOC_LABEL_BYTES (4096) budget** | **Yes** -- both quoted and bareword | No | No |
| **Empty-delimiter rejection** | **Yes** -- `if delimiter.is_empty() { return None; }` | No | No |
| **Backslash-escape handling inside quoted label** | No | No | **Yes** (`<<"E\"D"` keeps `\"` intact) |
| **`__DATA__` marker line-end handling** | `consume_newline()` after manual loop | Same as #6094 | **Cleanest**: `find_line_end()` + `consume_newline()` |
| **Suspicious unrelated change** | None | Removes `visible_end` `\r`-strip in `find_line_end` -- looks like a regression unrelated to the heredoc theme | None |
| **Test additions** | Quoted/unquoted distinction, `<<~` backtick, CR-only malformed, CRLF data marker, 20K-char malformed-label budget test (≤2s) | `<<~` with tabs, case-sensitive terminator (lowercase `eof`), CR-only backtick malformed, CR-only data marker, large CRLF body budget test | `<<~` with tab terminator, CR-only data marker, broader unterminated-quoted-label set, 50K-line unterminated-label timeout test |

**Recommendation: land #6094 as canonical.** Reasoning:

1. **Only #6094 has the `MAX_HEREDOC_LABEL_BYTES` budget.** This is the actual security boundary -- without it, a malicious or accidental input like `<<"AAAA...20Kchars"` causes a heredoc-label scan that, while bounded by `position < input.len()`, still walks every byte and produces a multi-KB heap allocation per call. The 4096-byte budget caps memory and CPU per malformed input. #6094's own security test (`test_malformed_quoted_heredoc_label_is_bounded`) confirms ≤2s on 20K bytes.

2. **Only #6094 rejects empty delimiters.** `<<""` should not become a `HeredocStart` -- it's a syntax error in Perl. Without this check, the lexer downstream tries to find the matching empty terminator on every blank line, which is both wrong semantically and another scan-budget hazard.

3. **Only #6094 applies the budget to bareword delimiters too** (`if delim.len() >= MAX_HEREDOC_LABEL_BYTES { break; }` in the bareword branch). Symmetric defense -- without this, the quoted-label budget is half-protective.

4. **#6095 contains a suspicious unrelated change**: removes the trailing `\r`-stripping in `find_line_end`'s `visible_end` calculation. That function presumably feeds heredoc body extraction; stripping `\r` before `\n` is the right behavior for distinguishing CRLF line content from line terminators, and removing it without a separate justifying test suggests it's accidental scope drift. **This is the strongest single reason not to land #6095**.

5. **#6096's backslash-escape handling and `find_line_end` use are genuine improvements** -- worth porting as follow-ups onto post-#6094 master:
   - The `if ch == '\\'` arm in `parse_quoted_heredoc_label` correctly keeps `\"` and `\\` sequences as literal characters in the label, matching Perl's actual quoted-heredoc-label semantics.
   - Replacing the manual `while ... != b'\n'` loop in the `__DATA__` marker path with `find_line_end()` + `consume_newline()` is cleaner and naturally handles CR/CRLF/LF.

**Suggested merge plan**:

1. **Land #6094** as the canonical security-leading fix.
2. **Port #6096's escape-handling block and `find_line_end()` use** as a follow-up PR onto post-#6094 master. Both are additive, both improve correctness, and both compose cleanly with #6094's budget.
3. **Port #6095's interesting tests** (especially `heredoc_quoted_label_requires_exact_terminator` -- case-sensitive terminator distinction is a real Perl semantic) onto post-#6094 master, but **discard the `visible_end` change** in #6095.
4. **Close #6095 and #6096** as superseded by #6094 + the planned follow-ups.

EffortlessSteven:
Closing as REDUNDANT — #6095 implements the same heredoc hardening scope and is further along in pipeline (review-reviewed). Your escaped-quote handling in parse_quoted_heredoc_delimiter is solid; extracting into winner for full delimiter quoting semantics. Thank you.

PR #6097, opened 2026-04-24
fix(lexer): harden Unicode, BOM, and multibyte span handling (#6599)

EffortlessSteven:
### Motivation
- Prevent panics and invalid span slicing when the lexer encounters Unicode-heavy prefixes or intermediate non‑char byte offsets while preserving the public token model.
- Ensure BOM normalization, emoji/ZWJ/variation-selector cases, and supplementary Unicode sequences do not break heredoc/quote detection or produce incorrect byte spans.

### Description
- Added a defensive `normalize_char_boundary()` helper and call sites in `PerlLexer::next_token` plus guards in `current_char`, `peek_char`, and `advance` to ensure `position` always lands on UTF-8 `char` boundaries before slicing or `chars()` use.
- Extended `is_perl_identifier_continue` ranges to include supplementary variation selectors and emoji tag sequence code points so ZWJ/VS/skin-tone/tag sequences are treated as identifier continuations where appropriate.
- Reworked `unicode_fix_test.rs` to assert span invariants (char boundaries, in-range starts/ends, and `token.text` == slice by span) and added regression cases that include BOM at start, emoji+ZWJ prefixes, and the original failing `¡<<'` shape.
- Added edge-case tests in `edge_case_tests.rs` that verify emoji ZWJ identifiers are single tokens and that a file-start BOM is skipped without shifting identifier spans.

### Testing
- Ran `cargo test -p perl-lexer unicode` and the unicode-focused tests passed.
- Ran `cargo test -p perl-lexer` and the crate test suite passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c910edc8333844c601f2cc64fe5)

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -338,6 +338,17 @@ impl<'a> PerlLexer<'a> {
         }
     }
 
+    /// Ensure the internal byte offset points at a UTF-8 char boundary.
+    ///
+    /// This is a defensive guard against malformed intermediate offsets from
+    /// complex lookahead/backtracking paths so downstream slicing never panics.
+    #[inline]
+    fn normalize_char_boundary(&mut self) {
+        while self.position < self.input.len() && !self.input.is_char_boundary(self.position) {
+            self.position += 1;
+        }
+    }
+
     /// Set the lexer mode (for resetting state at statement boundaries)
     pub fn set_mode(&mut self, mode: LexerMode) {
         self.mode = mode;
@@ -400,6 +411,7 @@ impl<'a> PerlLexer<'a> {
         if self.position == 0 {
             self.normalize_file_start();
         }
+        self.normalize_char_boundary();
 
         // Loop to avoid recursion when processing heredocs
         loop {
@@ -814,6 +826,9 @@ impl<'a> PerlLexer<'a> {
     #[inline(always)]
     fn current_char(&self) -> Option<char> {
         if self.position < self.input_bytes.len() {
+            if !self.input.is_char_boundary(self.position) {
+                return None;
+            }
             // For ASCII, direct access is safe
             let byte = Self::byte_at(self.input_bytes, self.position);
             if byte < 128 {
@@ -830,6 +845,9 @@ impl<'a> PerlLexer<'a> {
     #[inline(always)]
     fn peek_char(&self, offset: usize) -> Option<char> {
         if offset > self.config.max_lookahead {
+            return None;
+        }
+        if !self.input.is_char_boundary(self.position) {
             return None;
         }
 
@@ -852,6 +870,10 @@ impl<'a> PerlLexer<'a> {
     #[inline(always)]
     fn advance(&mut self) {
         if self.position < self.input_bytes.len() {
+            if !self.input.is_char_boundary(self.position) {
+                self.normalize_char_boundary();
+                return;
+            }
             let byte = Self::byte_at(self.input_bytes, self.position);
             if byte < 128 {
                 // ASCII fast path
--- a/crates/perl-lexer/src/unicode.rs
+++ b/crates/perl-lexer/src/unicode.rs
@@ -75,8 +75,12 @@ pub fn is_perl_identifier_continue(ch: char) -> bool {
             0x200C | 0x200D |
             // Standard variation selectors (e.g. U+FE0F) used to keep emoji presentation.
             0xFE00..=0xFE0F |
+            // Supplementary variation selectors.
+            0xE0100..=0xE01EF |
             // Fitzpatrick skin-tone modifiers.
-            0x1F3FB..=0x1F3FF
+            0x1F3FB..=0x1F3FF |
+            // Emoji tag sequence characters (used by some flag-like emoji sequences).
+            0xE0020..=0xE007F
         )
 }
 
@@ -139,8 +143,10 @@ mod tests {
         assert!(is_perl_identifier_continue('\u{200C}'));
         assert!(is_perl_identifier_continue('\u{200D}'));
         assert!(is_perl_identifier_continue('\u{FE0F}'));
+        assert!(is_perl_identifier_continue('\u{E0100}'));
         assert!(is_perl_identifier_continue('\u{1F3FB}'));
         assert!(is_perl_identifier_continue('\u{1F3FD}'));
+        assert!(is_perl_identifier_continue('\u{E0067}'));
         assert!(!is_perl_identifier_continue(' '));
         assert!(!is_perl_identifier_continue('-'));
     }
--- a/crates/perl-lexer/tests/edge_case_tests.rs
+++ b/crates/perl-lexer/tests/edge_case_tests.rs
@@ -1019,6 +1019,41 @@ fn unicode_subroutine_name() -> R {
     Ok(())
 }
 
+#[test]
+fn unicode_emoji_identifier_with_joiners_is_single_token() -> R {
+    let input = "my $👩‍💻 = 1;";
+    assert_terminates(input);
+    let sig = significant(input);
+    let emoji_var = sig
+        .iter()
+        .find(|t| matches!(&t.token_type, TokenType::Identifier(id) if id.as_ref() == "$👩‍💻"))
+        .ok_or("expected emoji identifier token")?;
+
+    assert_eq!(emoji_var.text.as_ref(), "$👩‍💻");
+    assert_eq!(emoji_var.text.len(), emoji_var.end - emoji_var.start);
+    Ok(())
+}
+
+#[test]
+fn file_start_bom_is_skipped_and_does_not_shift_identifier_span() -> R {
+    let input = "\u{FEFF}my $name = 1;";
+    assert_terminates(input);
+    let sig = significant(input);
+    let my_kw = sig
+        .iter()
+        .find(|t| matches!(&t.token_type, TokenType::Keyword(k) if k.as_ref() == "my"))
+        .ok_or("expected my keyword token")?;
+    let id = sig
+        .iter()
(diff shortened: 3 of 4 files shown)


EffortlessSteven:
### Meta-review: lexer Unicode/BOM cluster (#6097, #6098, #6099)

These three PRs share the title `fix(lexer): harden Unicode, BOM, and multibyte span handling`. All touch `crates/perl-lexer/src/{lib.rs,unicode.rs}`, `tests/edge_case_tests.rs`, and `tests/unicode_fix_test.rs`. Parallel maintainer attempts at the same goal.

**Comparison**:

| Aspect | #6097 (+104/-32) | #6098 (+125/-62) | #6099 (+141/-46) |
|---|---|---|---|
| **Defensive runtime guard** | `normalize_char_boundary` actively skips bad-boundary positions in `current_char`/`advance`/`peek` | None | `debug_assert!(input.is_char_boundary(position))` -- debug-only |
| **Emoji helper extraction** | No | `is_emoji_identifier_start` + `is_perl_identifier_continuation_extension` | `is_emoji_identifier_start` + `is_emoji_identifier_continue` |
| **BOM detection** | Bytes only (`[0xEF, 0xBB, 0xBF]`) | Bytes **and** `'\u{FEFF}'` codepoint variant | Bytes only, extracted to `UTF8_BOM_BYTES` const |
| **Supp variation selectors** (`0xE0100..=0xE01EF`) | Yes | Yes | Yes |
| **Emoji tag sequences** (`0xE0020..=0xE007F`) | Yes | Yes | **No** |
| **Test additions** | Span boundary asserts in edge_case_tests | `is_char_boundary` + round-trip assertions in *both* test files | Boundary asserts + broadest new test cases (BOM, multibyte spans, weird Unicode prefixes) |

**Recommendation**: there's no single "complete" PR -- each has something the others lack:

1. **Production code: #6098 is most complete.** Its `is_emoji_identifier_start` / `is_perl_identifier_continuation_extension` helpers cleanly split the two roles, both new Unicode ranges (supp variation selectors AND emoji tag sequences), and BOM detection accepts both raw bytes *and* the `\u{FEFF}` codepoint (which matters when `position != 0` somehow but the codepoint is at the start, e.g. after a defensive reset).

2. **#6099's tests are broadest.** Specifically:
   - `test_bom_is_skipped_and_spans_stay_byte_exact` -- pinned BOM offset of 3 bytes, round-trips text against input slice.
   - `test_multibyte_spans_and_emoji_joiner_continuations` -- the family-of-four ZWJ sequence `$👨‍👩‍👧‍👦️` exercises a 25-byte identifier with multiple ZWJ + variation selectors. Strong regression net.
   - `test_weird_unicode_prefixes_do_not_panic` -- catches a class of "Unicode-prefix-tricks-the-lexer-into-thinking-this-is-a-quote-op" bugs.

3. **#6097's `normalize_char_boundary`** is the most *defensive* of the three. While #6098/#6099 fix the actual bug paths, #6097 adds a belt-and-suspenders guard that auto-recovers from any future mid-char position regardless of where it came from. That's a tradeoff:
   - **Pro**: catches future regressions silently (no panic).
   - **Con**: silently skipping past mid-char bytes can mask real bugs upstream and produce subtly-wrong tokens. Would prefer panicking in dev / silent in release, which is essentially what #6099's `debug_assert!` does.

**Suggested merge plan**:

1. **Land #6098** as the canonical production fix. Its helper extraction is the right code-quality improvement, both new Unicode ranges are needed, and the dual BOM detection (bytes + codepoint) is genuinely more robust than either alternative.

2. **Port #6099's three test functions onto post-#6098 master** as a follow-up PR. They're additive and orthogonal to #6098's changes.

3. **Consider #6097's `normalize_char_boundary` as a separate decision**: either close as "we prefer #6099's debug_assert! contract", or land it on top with the contract clarified as "intentional silent skip with `tracing::warn!` to surface upstream regression candidates."

4. **Close #6099 and #6097 in favor of #6098** with explicit notes pointing to the merge plan above.

EffortlessSteven:
## Plan-review verdict: cluster C21 — corrective analysis

**TL;DR:** The wrong PR survived. The maintainer's meta-review (posted 2026-04-25T06:32) recommended keeping #6098 and closing #6097 + #6099, but the closure actions reversed this: #6098 and #6099 were closed, leaving #6097 open. This comment documents the corrective path.

---

### Cluster state as of 2026-04-25T20:05

| PR | State | Issue |
|----|-------|-------|
| #6097 (this PR) | OPEN | Should be closed per maintainer recommendation |
| #6098 | CLOSED | Should be the winner — was preferred and then mistakenly closed |
| #6099 | CLOSED | Correct: missing tag range `0xE0020..=0xE007F` is a regression vs both others |

---

### Technical comparison (my independent analysis)

#### lib.rs changes

**#6097 (this PR)** adds `normalize_char_boundary()` — a defensive helper that actively skips non-boundary byte positions by advancing until `is_char_boundary()` is true. It calls this helper at 4 sites: entry to `next_token`, and guards inside `current_char`, `peek_char`, and `advance`.

The problem with this approach:
- Silently skipping continuation bytes to the next char boundary means the **current character is dropped from the token being lexed**. A token that was mid-formation will silently lose the leading bytes of whatever multibyte char landed there.
- This masks bugs upstream (e.g., a future lookahead that accidentally lands between bytes) rather than fixing them. It is belt-and-suspenders without suspenders.
- In practice, the current `advance()` already uses `ch.len_utf8()` which always lands on boundaries. There is no legitimate code path in master that creates a mid-char `self.position`. The defensive guards in #6097 are guarding against a condition that should never occur — and if it did, the right behavior is `debug_assert!` (catch it in dev) not silently skip (corrupt tokens in prod).

**#6098** addresses BOM handling correctly: it adds a `|| self.current_char() == Some('\u{FEFF}')` branch and uses `advance()` for BOM skipping, which is UTF-8-safe and also updates `line_start_offset` from the post-advance position rather than hardcoding `3`. This is more robust than the hardcoded `self.position = 3; self.line_start_offset = 3` in master (which is correct only because U+FEFF is always 3 bytes in UTF-8, but the intent is clearer with `advance()`).

**#6097's BOM handling** keeps the raw-bytes approach unchanged. Not wrong, but less clear.

#### unicode.rs changes

Both #6097 and #6098 add:
- Supplementary variation selectors `0xE0100..=0xE01EF`
- Emoji tag sequences `0xE0020..=0xE007F`

But #6098 additionally:
- Extracts `is_emoji_identifier_start()` as a named `#[inline]` helper — cleaner, testable in isolation
- Extracts `is_perl_identifier_continuation_extension()` as a named `#[inline]` helper — the name documents the semantic intent

**#6097** keeps both classification blocks inline as `matches!()` arms inside `is_perl_identifier_start()` and `is_perl_identifier_continue()`. Functionally identical but harder to audit.

#### Test quality

- **#6097**: adds `assert_spans_are_valid()` helper with char-boundary assertions and round-trip text checks. Good. BOM test asserts `my_kw.start == 3` (hardcoded offset, fragile to BOM encoding changes but correct for UTF-8).
- **#6098**: adds `assert_token_spans_within_input()` and tests BOM skip, emoji ZWJ identifiers with supplementary variation selectors and tag chars directly in the identifier body (`\u{E0100}\u{E0067}`). The `emoji_joiner_variation_selector_identifier_has_valid_multibyte_spans` test directly exercises the new ranges added in unicode.rs, creating a behavioral linkage between the classification logic and its test.
- **#6099 (closed)**: had the broadest test set — `test_multibyte_spans_and_emoji_joiner_continuations` with `$👨‍👩‍👧‍👦️` (25-byte ZWJ family), and `test_weird_unicode_prefixes_do_not_panic` exercising heredoc/quote-op confusion patterns. These test cases are worth porting.

---

### Recommended resolution

**Option A (preferred):** Reopen #6098 and close this PR (#6097).

#6098 is the cleaner fix: proper helper extraction, `advance()`-based BOM handling, and the tag range. Its only gap vs #6099 is the missing `test_weird_unicode_prefixes_do_not_panic` and `test_multibyte_spans_and_emoji_joiner_continuations` tests, which can be added as a follow-up or cherry-picked onto the #6098 branch before merging.

**Option B:** Keep this PR (#6097) but make these changes before merging:
1. Remove `normalize_char_boundary()` entirely — replace the guard in `next_token` with `debug_assert!(self.input.is_char_boundary(self.position))` (following #6099's pattern).
2. Remove the runtime guards added to `current_char`, `peek_char`, and `advance` — they defend against impossible conditions and would corrupt tokens silently if they ever fired.
3. Extract `is_emoji_identifier_start` and `is_perl_identifier_continuation_extension` as named helpers (following #6098's refactor).
4. Port the three test cases from #6099: `test_bom_is_skipped_and_spans_stay_byte_exact`, `test_multibyte_spans_and_emoji_joiner_continuations`, `test_weird_unicode_prefixes_do_not_panic`.

Option B requires substantial rework of this PR's production code. Option A is faster.

---

### #6099 closure: confirmed correct

#6099 misses `0xE0020..=0xE007F` (emoji tag sequences) in `is_emoji_identifier_continue`, which is a regression against both #6097 and #6098. Closing #6099 was correct.

---

### Verify commands (for builder executing Option B, or for reviewer validating Option A after reopen)

```bash
cargo test -p perl-lexer -- unicode
cargo test -p perl-lexer
cargo clippy -p perl-lexer --tests
cargo xtask fmt
```

Expected: all tests pass, no clippy warnings on the touched functions.

EffortlessSteven:
Closing per plan-review verdict (see comment above): this PR uses `normalize_char_boundary()` which silently drops bytes rather than fixing root causes. The maintainer's own analysis (2026-04-25T06:32) recommended keeping #6098 instead. #6098 should be reopened; if not, a builder should port Option B changes onto a new branch. See the plan-review comment above for exact spec.

PR #6101, opened 2026-04-24
perf(lexer): optimize measured hot path in token classification/scanning (#6595)

EffortlessSteven:
### Motivation

- Reduce allocation overhead in lexer hot paths where `TokenType` payloads and `Token.text` were each allocating separate `Arc<str>` instances, which inflated work in number/identifier/operator-heavy workloads.

### Description

- Reuse a single `Arc<str>` per token for both the `TokenType` payload and `Token.text` in hot scanner paths by creating `let text_arc: Arc<str> = Arc::from(text)` and cloning the `Arc` where the token enum also needs ownership.
- Applied the allocation reuse in `try_number`/`parse_decimal_number` (number scanning), `try_identifier_or_keyword` (identifier/keyword classification), and `try_operator` (operator scanning) to keep the change local and measurable.
- No semantic changes to token kinds or lexer behavior were introduced; tokens still carry identical text and kinds but reduce duplicate allocation work.

### Testing

- Ran `cargo bench -p perl-lexer --bench lexer_benchmarks -- --noplot` and observed statistically significant improvements post-change, for example: `keyword_heavy` improved from ~457.00–461.26 µs to ~383.41–391.32 µs (~17.8% faster), `number_parsing` improved from ~1.0718–1.0833 µs to ~895.90–936.60 ns (~18.3% faster), and `large_file` improved from ~2.6188–2.6989 ms to ~2.2802–2.3032 ms (~13.8% faster).
- Ran `cargo test -p perl-lexer` and all lexer tests passed.
- Ran `cargo check --workspace --all-targets` and the workspace type-check succeeded.
- Ran formatting (`cargo xtask fmt`) successfully; `just pr-fast` was not available in this environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c9503888333a841035ad331bba8)

PR #6100, opened 2026-04-24
perf(lexer): optimize measured hot path in token classification/scanning (#6596)

EffortlessSteven:
### Motivation
- Benchmarks showed identifier-heavy tokenization as a hot path; repeated char-level continuation checks in `try_identifier_or_keyword` were creating measurable overhead that could be optimized locally.

### Description
- Add an ASCII byte-scan fast path inside `try_identifier_or_keyword` to advance through common ASCII identifier characters without repeated `char` decoding, falling back to the existing Unicode `is_perl_identifier_continue` checks when needed.
- Preserve quote-operator semantics by adding a small helper `is_quote_op_word_prefix` and by stopping the fast-path before an apostrophe when the leading bytes match `m`, `q`, `qq`, `qw`, `qx`, or `qr`.
- Restore lookahead-sensitive behavior for `::` package separators by gating multi-byte `::` scanning with `config.max_lookahead` to avoid changing semantics for small-lookahead configurations.
- Keep Unicode behavior, error handling, and public API unchanged; changes are local to `crates/perl-lexer/src/lib.rs` in the identifier scanning hot path.

### Testing
- Ran `cargo bench -p perl-lexer --bench lexer_benchmarks -- --noplot` and observed measurable improvements (baseline -> optimized): `simple_tokens` ~1.82 µs -> ~1.72 µs (~6.5% faster median), `string_interpolation` ~4.50 µs -> ~4.02 µs (~11.5% faster median), `whitespace_heavy` ~2.10 µs -> ~1.98 µs (~7% faster median), and `keyword_heavy` ~580 µs -> ~500 µs (~13.5% faster median).
- Ran `cargo test -p perl-lexer` which initially exposed a failure (`with_config_small_lookahead_limits_namespace_parsing`) that was fixed by restoring the lookahead-sensitive `::` behavior, after which `cargo test -p perl-lexer` completed successfully (all tests passed).
- Ran `cargo check --workspace --all-targets` and `cargo check --all-targets -p perl-lexer` successfully.
- Ran `cargo clippy -p perl-lexer --all-targets -- -D warnings` and addressed lints so the check completed cleanly.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c9503888333a841035ad331bba8)

PR #6102, opened 2026-04-24
perf(lexer): optimize measured hot path in token classification/scanning (#6594)

EffortlessSteven:
### Motivation
- Reduce hot-path allocation and branch overhead in the lexer where benchmarks showed most time spent in number/identifier token creation and quote-operator boundary checks.

### Description
- Reuse a single `Arc<str>` per token in number scanning and identifier return paths to avoid creating duplicate `Arc::from` allocations for `token_type` and `text` (changes in `crates/perl-lexer/src/lib.rs`).
- Replace repeated slice-pattern matching for quote-operator words with a byte-level helper `is_quote_operator_word(bytes, start, end)` to avoid allocating/slicing and heavy branch work during identifier scanning.
- Construct `TokenType` variants using the reused `Arc<str>` to ensure a single allocation per token where possible.

### Testing
- Ran benchmarks: `cargo bench -p perl-lexer --bench lexer_benchmarks` and verified measurable wins and regressions reported by Criterion (see before/after below).
- Ran unit/integration tests: `cargo test -p perl-lexer`, all tests passed.
- Type/target checks: `cargo check --workspace --all-targets` passed.
- Lints/format: `cargo clippy -p perl-lexer --all-targets` and `cargo xtask fmt` ran without errors for changed crate.

Before/after (criterion summary):
- `number_parsing`: 1.1472–1.1683 µs -> 832.29–868.96 ns (~25% faster)
- `keyword_heavy`: 438.86–444.34 µs -> 390.09–397.70 µs (~11.5% faster)
- `simple_tokens`: 1.4425–1.4512 µs -> 1.4151–1.4232 µs (~1.7% faster)
- `large_file`: no statistically significant change

Notes: `just pr-fast` is not available in this environment and was not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c9503888333a841035ad331bba8)

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -1290,10 +1290,11 @@ impl<'a> PerlLexer<'a> {
                 if pos > digit_start && saw_digit {
                     self.position = pos;
                     let text = &self.input[start..self.position];
+                    let text_arc: Arc<str> = Arc::from(text);
                     self.mode = LexerMode::ExpectOperator;
                     return Some(Token {
-                        token_type: TokenType::Number(Arc::from(text)),
-                        text: Arc::from(text),
+                        token_type: TokenType::Number(text_arc.clone()),
+                        text: text_arc,
                         start,
                         end: self.position,
                     });
@@ -1313,10 +1314,11 @@ impl<'a> PerlLexer<'a> {
                 if pos > digit_start && saw_digit {
                     self.position = pos;
                     let text = &self.input[start..self.position];
+                    let text_arc: Arc<str> = Arc::from(text);
                     self.mode = LexerMode::ExpectOperator;
                     return Some(Token {
-                        token_type: TokenType::Number(Arc::from(text)),
-                        text: Arc::from(text),
+                        token_type: TokenType::Number(text_arc.clone()),
+                        text: text_arc,
                         start,
                         end: self.position,
                     });
@@ -1336,10 +1338,11 @@ impl<'a> PerlLexer<'a> {
                 if pos > digit_start && saw_digit {
                     self.position = pos;
                     let text = &self.input[start..self.position];
+                    let text_arc: Arc<str> = Arc::from(text);
                     self.mode = LexerMode::ExpectOperator;
                     return Some(Token {
-                        token_type: TokenType::Number(Arc::from(text)),
-                        text: Arc::from(text),
+                        token_type: TokenType::Number(text_arc.clone()),
+                        text: text_arc,
                         start,
                         end: self.position,
                     });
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
### Meta-review: lexer hot-path optimization cluster (#6100, #6101, #6102, #6103)

These four PRs share the same title (`perf(lexer): optimize measured hot path in token classification/scanning`) and all touch only `crates/perl-lexer/src/lib.rs`. They are parallel maintainer attempts.

**My read of the four**:

| PR | Lines | Scope | Notes |
|----|------|-------|-------|
| **#6100** | +55/-27 | **Superset**: Arc-clone dedup for number tokens **AND** new fast ASCII byte-level path for identifier continuation (`is_quote_op_word_prefix`, byte-indexed scan, falls through to UTF-8 path on non-ASCII byte) | Strongest perf payoff |
| #6101 | +27/-15 | Arc-clone dedup for number tokens (single allocation, cloned for `token_type` and `text`) | Subset of #6100 |
| #6102 | +45/-23 | (couldn't fetch diff -- API timeout) | Need to confirm scope |
| #6103 | +24/-13 | Arc-clone dedup for number tokens, near-identical to #6101 | Subset of #6100 |

**Recommendation**: **land #6100** as the canonical version. Its fast-ASCII identifier-continuation path is the bigger win for a hot loop that runs once per identifier character; the Arc-clone dedup is a strict subset of what #6100 already does.

**Specifically about #6100's fast-ASCII path**:

```rust
while self.position < len {
    let byte = bytes[self.position];
    if byte == b'\'' && is_quote_op_word_prefix(&bytes[start..self.position]) {
        break;
    }
    if byte.is_ascii_alphanumeric() || byte == b'_' || byte == b'\'' {
        self.position += 1;
        continue;
    }
    if byte < 128 { break; }
    // ... falls through to UTF-8 path for non-ASCII bytes
}
```

This is the right shape: the *vast majority* of Perl identifier characters are ASCII alphanumeric/underscore. Bypassing UTF-8 char decoding for those cases is a straightforward perf win without changing semantics. The `is_quote_op_word_prefix` predicate replaces the `matches!(... "m" | "q" | "qq" | "qw" | "qx" | "qr")` pattern -- presumably operates on byte slices for the same reason.

The fall-through to UTF-8 path on `byte >= 128` preserves correctness for non-ASCII identifiers (which Perl 5.16+ supports under `use utf8`).

**For the maintainers**:

- Merge **#6100**.
- Close #6101 and #6103 as superseded (their Arc-clone dedup is included in #6100).
- Compare #6102 against #6100 once the API stabilizes; if its scope is also a subset, close. If it adds something orthogonal, recommend rebasing onto post-#6100 master.

EffortlessSteven:
Superseded by #6100: same scope (lexer perf hot-path optimization in token classification/scanning); duplicate Codex variant. Consolidating to oldest first-mover.

PR #6105, opened 2026-04-24
fix(lexer-checkpoint): harden checkpoint edit application and cache invariants (#6591)

EffortlessSteven:
### Motivation

- Incremental edits could leave checkpoints with stale line/column data or trigger overflow/ordering edge-cases after cache updates, risking incorrect incremental lexing state.
- The cache lacked a single canonical rebalance step after mutations, making ordering/uniqueness invariants fragile after edits.

### Description

- Add `reset_state_to` and make `LexerCheckpoint::apply_edit` conservative: use saturating math, byte-shift checkpoints after edits when safe, and invalidate (reset) checkpoints that fall inside replaced ranges while marking line/column as unknown (`line=0,column=0`).
- Preserve byte-accurate `position` updates but explicitly clear `current_pos.line`/`column` when newline effects cannot be safely computed.
- Centralize cache maintenance in `CheckpointCache::rebalance`, invoked from `add` and `apply_edit`, to sort by position, deduplicate collapsed positions, and trim to `max_checkpoints` while preserving broad coverage.
- Add focused tests covering edit-before-checkpoint, edit-spanning-checkpoint (overlap reset), newline-related line/column invalidation, overflow/malformed edit inputs stability, and cache ordering/uniqueness after edits.

### Testing

- Ran unit tests with `cargo test -p perl-lexer` which passed (new and existing tests all OK).
- Ran workspace checks with `cargo check --workspace --all-targets` which passed.
- Ran formatting `cargo xtask fmt` as part of verification.
- New tests specifically exercise: edit before checkpoint, edit spanning checkpoint, line/column invalidation for newline edits, cache invariants after edits, and no-panics for extreme/malformed edit parameters.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb1c72c6948333bda77b89506e810b)

### DIFF
--- a/crates/perl-lexer/src/checkpoint.rs
+++ b/crates/perl-lexer/src/checkpoint.rs
@@ -98,30 +98,46 @@ impl LexerCheckpoint {
         }
     }
 
+    fn reset_state_to(&mut self, new_position: usize) {
+        self.position = new_position;
+        self.mode = LexerMode::ExpectTerm;
+        self.delimiter_stack.clear();
+        self.in_prototype = false;
+        self.prototype_depth = 0;
+        self.after_sub = false;
+        self.after_arrow = false;
+        self.hash_brace_depth = 0;
+        self.after_var_subscript = false;
+        self.paren_depth = 0;
+        self.context = CheckpointContext::Normal;
+
+        // We know byte offsets, but not line/column deltas from edit text.
+        // `0:0` explicitly means "line/column invalidated; recompute if needed".
+        self.current_pos = Position::new(new_position, 0, 0);
+    }
+
     /// Apply an edit to this checkpoint
     pub fn apply_edit(&mut self, start: usize, old_len: usize, new_len: usize) {
-        if self.position > start {
-            if self.position >= start + old_len {
-                // Checkpoint is after the edit
-                self.position = self.position - old_len + new_len;
-            } else {
-                // Checkpoint is inside the edit - invalidate
-                self.position = start;
-                self.mode = LexerMode::ExpectTerm;
-                self.delimiter_stack.clear();
-                self.in_prototype = false;
-                self.prototype_depth = 0;
-                self.after_sub = false;
-                self.after_arrow = false;
-                self.hash_brace_depth = 0;
-                self.after_var_subscript = false;
-                self.paren_depth = 0;
-                self.context = CheckpointContext::Normal;
-            }
+        let edit_end = start.saturating_add(old_len);
+
+        if self.position < start {
+            // Edit is strictly after this checkpoint.
+            return;
         }
 
-        // Update position tracking
-        // In a real implementation, we'd update line/column based on the edit
+        if self.position >= edit_end {
+            // Checkpoint is at/after the edited range and can be byte-shifted.
+            self.position = self.position.saturating_sub(old_len).saturating_add(new_len);
+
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
### Review: harden checkpoint edit application and cache invariants

**Has a near-duplicate in #6107 (same title, same crate, slightly different stats: this PR is +104/-40 vs #6107 +103/-40)**. Likely parallel maintainer attempts at the same fix; one of these should land and the other should rebase / close.

**The substantive change** in this PR is correct: extracts the "invalidate checkpoint state" logic into a `reset_state_to(new_position)` helper, replacing inline duplication in `apply_edit`. Then `apply_edit` becomes:

```rust
pub fn apply_edit(&mut self, start: usize, old_len: usize, new_len: usize) {
    let edit_end = start.saturating_add(old_len);

    if self.position < start {
        // Edit is strictly after this checkpoint -- nothing to do.
        return;
    }
    // ... (rest of the dispatch in elided diff)
}
```

**The `saturating_add` is the right defensive choice** -- a malformed `(start, old_len)` pair where `old_len = usize::MAX` shouldn't wrap and produce a `edit_end < start`. Saturating to `usize::MAX` gives the safe "edit covers everything to the end" interpretation, which results in invalidation -- the correct conservative behavior.

**The extracted `reset_state_to` is the right abstraction**: every field of `LexerCheckpoint` that depends on lexer position (mode, delimiter stack, prototype state, paren depth, context) gets reset to its initial state. The doc-comment explicitly notes:

```rust
// We know byte offsets, but not line/column deltas from edit text.
// `0:0` explicitly means "line/column invalidated; recompute if needed".
self.current_pos = Position::new(new_position, 0, 0);
```

This is the right call -- propagating bogus line/column from the pre-edit state would produce wrong diagnostics; flagging them as `0:0` is a clean sentinel for "needs recomputation."

**Concerns / things to verify**:

1. **`Position::new(new_position, 0, 0)` as the "invalid" sentinel** -- is `0:0` actually invalid, or could it legitimately occur (line 0, column 0 = start of file)? If a checkpoint at the very first byte gets reset, `0:0` is its real value, not a sentinel. This works in practice because the next `next_token()` will recompute, but a more explicit `Position::INVALID` const or `Option<Position>` would make the contract self-documenting.

2. **`self.position < start` check uses `<`, not `<=`** -- if the edit starts *exactly at* the checkpoint position, the checkpoint is at the edge. Worth a one-line comment confirming the boundary semantics: an edit at exactly `position` either touches or doesn't touch the checkpoint state, and the chosen semantics need to be explicit. The diff's first branch (`< start`) treats `start == position` as "edit affects this checkpoint" -- probably correct (the checkpoint sits at the leading edge of the edit), but worth confirming against the original semantics.

**Coordination with #6107**: I'd recommend whichever maintainer has been involved with the lexer-checkpoint work picks one PR to land; the other should rebase to a no-op delta. Without that, the merge order between them matters and the loser produces a dirty conflict.

**Verdict**: clean refactor + correct saturating_add + well-named helper. The duplicate-PR coordination is the only blocker. Pre-existing master CI failures are unrelated; #6495 fixes both halves of the master block.

EffortlessSteven:
Superseded by #6107 (same lexer-checkpoint hardening scope, picked as keeper for cluster).

PR #6117, opened 2026-04-24
fix(tree-sitter-perl-c): make parse_c byte-safe and triage-friendly (#6583)

EffortlessSteven:
### Motivation

- The existing `parse_c` CLI forced UTF-8 decoding and produced limited output, which made triaging non-UTF-8 fixtures and corpus debugging difficult.
- The crate API already supports raw bytes and non-UTF-8 Perl files, so the CLI should follow the same byte-oriented semantics and provide targeted triage output.

### Description

- Switched `parse_c` to read files as raw bytes (`fs::read`) and call `parser.parse(&[u8], ...)` so non-UTF-8 inputs are supported without UTF-8 decoding.
- Added a small CLI option parser with flags `--root-kind`, `--has-error`, and `--sexp` and a usage message to enable optional triage output while keeping the default invocation minimal.
- Improved error reporting for syntax-failing parses by printing the first error node with byte-range and byte/row positions and exiting with status `1` on parse errors.
- Updated `crates/tree-sitter-perl-c/README.md` with `parse_c` examples and notes about byte-safe behavior and triage flags.

### Testing

- Ran `cargo test -p tree-sitter-perl-c` and all tests passed (unit tests, BDD workflows, and doctests).
- Ran `cargo check --all-targets -p tree-sitter-perl-c` and it completed successfully.
- Ran `cargo clippy -p tree-sitter-perl-c --all-targets` and the lint pass completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb276fa1ac8333b3175327847986e0)

PR #6118, opened 2026-04-24
fix(tree-sitter-perl-c): make parse_c byte-safe and more useful for parse triage (#6582)

EffortlessSteven:
### Motivation

- The existing `parse_c` CLI read input as UTF-8 text which made triaging non-UTF-8 Perl fixtures and corpus debugging unreliable. 
- The crate API already supports raw bytes and non-UTF-8 Perl files, so the CLI should expose the same capability for accurate fixture triage. 
- Provide quick, actionable parse diagnostics (root kind, error presence, optional s-expression) while keeping the default exit semantics simple for scripting.

### Description

- Switch `parse_c` to read raw bytes with `fs::read` and parse via `parse_perl_bytes` so non-UTF-8 input is accepted without prior decoding. 
- Add a small CLI argument parser and optional flags `--root-kind`, `--has-error`, and `--sexp` to provide targeted output for triage. 
- When error nodes are present `parse_c` prints an actionable diagnostic (file name, root kind, input byte length, parsed span) and exits `1`; IO/parse failures also exit `1`, while a clean parse exits `0`. 
- Update `crates/tree-sitter-perl-c/README.md` with examples and flag documentation showing the new byte-safe behavior and usage examples.

### Testing

- Ran `cargo test -p tree-sitter-perl-c` and all unit, BDD and doctests passed. 
- Ran `cargo clippy -p tree-sitter-perl-c` with no blocking lints. 
- Ran `cargo xtask fmt` to ensure formatting consistency and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb276fa1ac8333b3175327847986e0)

### DIFF
--- a/crates/tree-sitter-perl-c/README.md
+++ b/crates/tree-sitter-perl-c/README.md
@@ -84,8 +84,32 @@ for snippet in &["my $x = 1;", "print $x;"] {
 
 ## Binaries
 
-- `parse_c` — parse a Perl file and exit with 0 (success) or 1 (error)
-- `bench_parser_c` — parse a Perl file and print timing (requires `--features test-utils`)
+### `parse_c`
+
+`parse_c` reads source as raw bytes (non-UTF-8 safe), parses with the C grammar,
+and exits:
+
+- `0` when parse succeeds without tree-sitter error nodes
+- `1` when IO fails, parsing fails, or the resulting tree contains error nodes
+
+```bash
+cargo run -p tree-sitter-perl-c --bin parse_c -- path/to/file.pl
+cargo run -p tree-sitter-perl-c --bin parse_c -- --root-kind --has-error path/to/file.pl
+cargo run -p tree-sitter-perl-c --bin parse_c -- --sexp path/to/file.pl
+```
+
+Optional output flags:
+
+- `--root-kind` prints `root_kind=<kind>`
+- `--has-error` prints `has_error=<true|false>`
+- `--sexp` prints the root node s-expression
+
+When error nodes are present, `parse_c` emits an actionable diagnostic to stderr
+including file name, root kind, input byte length, and parsed span.
+
+### `bench_parser_c`
+
+- parse a Perl file and print timing (requires `--features test-utils`)
 
 ## Build Requirements
 
--- a/crates/tree-sitter-perl-c/src/bin/parse_c.rs
+++ b/crates/tree-sitter-perl-c/src/bin/parse_c.rs
@@ -2,37 +2,111 @@ use std::env;
 use std::fs;
 use std::process;
 
-fn main() {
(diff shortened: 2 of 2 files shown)


EffortlessSteven:
### Review: make parse_c byte-safe and more useful for parse triage

**Two coordinated improvements** to the `parse_c` binary in tree-sitter-perl-c.

1. **Byte-safe input reading**: previously `parse_c` likely read input as a UTF-8 string (per the README diff calling out "raw bytes (non-UTF-8 safe)"). For parse-triage on real-world Perl, where files may be Latin-1, mixed-encoding, or contain non-UTF-8 bytes mid-string, the parser should not refuse to even start. Switching to byte-level read makes the binary actually useful for the corpus-sweep / triage workflows that need to parse "whatever's on disk."

2. **Triage-friendly diagnostics**: new flags `--root-kind`, `--has-error`, `--sexp` for inspecting the parsed tree, plus an automatic stderr diagnostic on error nodes ("file name, root kind, input byte length, parsed span"). This is the right ergonomics for a triage tool -- when a parse fails on a corpus file, you want to know *where* and *what* without firing up a debugger.

**Correctness checks**:

- Reading as bytes (`fs::read` -> `Vec<u8>`) and passing to `tree_sitter::Parser::parse(&bytes, None)` is the right pattern. tree-sitter's C grammar consumes raw bytes; the language definition handles its own utf8/latin-1 disambiguation per scanner state.

- The `--has-error` / `--root-kind` / `--sexp` flags expose tree-sitter's existing tree-inspection API. Standard CLI ergonomics; no obvious correctness concern.

- Error-node diagnostic to stderr (rather than stdout) is the right choice -- piping `parse_c --sexp file.pl` for further analysis won't get noise mixed into the s-expression output.

1. The diagnostic format in the README ("file name, root kind, input byte length, parsed span") would be more grep-friendly with a fixed prefix like `parse_c: error: file=... root_kind=... bytes=... span=...`. Standard structured-log shape lets corpus-sweep scripts extract by field name instead of column position.

2. Adding a `--exit-zero-on-errors` flag would let triage scripts collect diagnostics across many files without short-circuiting on the first error. Currently exit 1 on error nodes is correct for CI gating, but for "audit my corpus" use cases the script has to swallow exit codes per file. Optional follow-up.

**Verdict**: small but high-leverage tooling improvement. Makes `parse_c` actually usable for the triage workflows the corpus-baseline machinery depends on. Pre-existing master CI failures are unrelated; #6495 fixes both halves of the master block.

EffortlessSteven:
Duplicate of #6117 (Codex ensemble burst, same surface — tree-sitter-perl-c parse_c.rs + README). Keeping #6117 (136 additions vs 123 here). Closing as ensemble dupe.

PR #6122, opened 2026-04-24
test(tree-sitter-perl-c): expand upstream query conformance coverage (#0000)

EffortlessSteven:
### Motivation

- Query-conformance testing only covered injections which left highlights/folds behaviors untested and made regressions hard to pinpoint.
- Using the vendored upstream query files directly ensures the crate’s queries remain compatible with the canonical `tree-sitter-perl` expectations.

### Description

- Add a focused integration test file `crates/tree-sitter-perl-c/tests/query_conformance.rs` that collects query captures and asserts presence of key captures rather than snapshotting full outputs.
- Implement `collect_captures`, `assert_capture_contains`, and `focused_query_from_upstream` helpers to run upstream query fragments and produce readable failure output.
- Add focused coverage for `highlights.scm` (core constructs and quote/heredoc captures), `folds.scm` (comments, blocks, heredoc, POD), and `injections.scm` (Inline::C / Inline::CPP heredocs, POD/comment/substitution evals).
- Tests assert specific capture names/contents so failures are easy to read and pinpoint without coupling to full query dumps.

### Testing

- Ran `cargo test -p tree-sitter-perl-c` and the new `query_conformance` tests passed (all tests in the crate succeeded).
- Ran `cargo xtask fmt` which completed successfully and formatted the new test file.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb277313408333b6f1e38300efd7ed)


EffortlessSteven:
## Maintainer Review: ALIGNED

**Verdict**: Implementation fits project direction and quality bar.

**Scope discipline**: Single focused file adding query conformance tests. No scope drift — spec matches delivery (highlights, folds, injections coverage).

**Pattern introduction**: Introduces test helpers (collect_captures, assert_capture_contains, focused_query_from_upstream) — procedural test utilities, not infrastructure patterns. No new test framework or config surface.

**Complexity budget**: 220 LOC of focused test coverage for upstream query behavior. Well-justified: prior coverage was injections-only, leaving highlights/folds behavior untested. Tests enable regression detection on vendored query files.

**Consistency**: Follows perl-lsp conventions:
- Uses idiomatic BTreeMap for deterministic key ordering (good catch on determinism)
- Tests return Result<(), Box<dyn Error>> (matches repo style)
- Helper functions are readable and concise
- Assertions include context output on failure (good UX for debugging)

**Test quality**: Tests are assertion-based (not snapshots), covering:
- Highlights: shebang, use/no/require, function keywords, scalars, comments, POD
- Folds: structural blocks, comments, heredoc, POD
- Injections: Inline::C/CPP, POD/comment/substitution evals

Assertions verify presence of specific capture names/contents rather than full output coupling (good design).

**One watch point**: Tests call include_str!() with relative paths to tree-sitter-perl/queries/. Current structure is stable; no action needed unless that crate undergoes major reorganization.

EffortlessSteven:
## Deep Review — APPROVED

**Verdict**: APPROVED. No correctness bugs found. No fix-forward needed.

### What I verified

**1. `include_str!` path robustness**
`../../../tree-sitter-perl/queries/` resolves correctly from `crates/tree-sitter-perl-c/tests/`. Path is identical in pattern to the existing `bdd_workflows.rs` usage (`include_str!("../../../tree-sitter-perl/queries/injections.scm")`). Since `tests/**` is excluded from the crate's `include` manifest, there is no publish-time breakage risk.

**2. `focused_query_from_upstream` — non-vacuous**
The helper does two independent jobs: (a) panics if the fragment is absent from upstream (catches query drift), (b) the joined query is executed by the tree-sitter engine against real Perl source and captures must appear. Neither assertion passes if the code under test is missing — the tests are not vacuous.

**3. Fixture representativeness**
All five test functions cover representative real-world Perl patterns:
- Shebang / `use` / `sub` / `$scalar` / POD / comment — core Perl constructs
- `qr/foo.*/` + `<<'SQL'` heredoc — quote-like operators
- `s/x/uc($value)/e` — eval-substitution injection (correctly exercises the `#match? @_modifiers "e"` path)
- `use Inline C` / `use Inline CPP` heredocs — language injection

**4. `cursor.captures()` usage**
The `StreamingIterator`-based loop correctly uses `(query_match, capture_index)` to access each individual capture via `query_match.captures.get(*capture_index)`. This is the idiomatic tree-sitter 0.26 Rust pattern and processes every capture for every pattern.

**5. Folds `.contains()` semantics**
`(comment)+ @fold` captures two consecutive comment lines as a single multi-line node. `assert_capture_contains` uses `.any(|v| v.contains(needle))`, so both `"# heading comment"` and `"# second comment"` correctly match as substrings of the combined capture text. The subroutine and heredoc assertions follow the same logic and are correct.

**6. Assertion strictness**
Every assertion requires a specific capture *name* AND specific *text content*. A regression that renames a capture (e.g., `@variable.scalar` → `@variable`) would cause `assert_capture_contains(&captures, "variable.scalar", "$value")` to fail on the missing key. A regression that drops a query rule entirely would similarly cause the assertion to fail.

### Minor gap noted (not blocking)
`matchup.scm` has no test coverage in either `bdd_workflows.rs` or `query_conformance.rs`. Recommend a follow-up issue — out of scope for this PR.

### Research verification
No external Perl semantics, LSP/DAP protocol, or crate API claims were made. Research verification skipped.

**Logic**: CORRECT  
**Tests**: GOOD — non-vacuous, catch drift and engine-level regressions  
**Regression risk**: LOW — purely additive test file, no production code changed  
**Research verification**: SKIPPED (no external claims)

PR #6145, opened 2026-04-24
fix(tree-sitter-perl-c): harden autoquoted scanner handling around comments and POD

EffortlessSteven:
### Motivation

- The vendored C scanner had fragile lookahead logic for deciding when a bareword becomes `TOKEN_FAT_COMMA_AUTOQUOTED` or `TOKEN_BRACE_AUTOQUOTED`, and it mishandled cases where line comments or POD-like blocks intervened between the identifier and the quote character.

### Description

- Tightened the post-identifier lookahead in `crates/tree-sitter-perl-c/c-src/scanner.c` to use an explicit loop that robustly skips whitespace and line comments and preserves a two-character lookahead state for `=>` detection.
- Added explicit handling for POD-like directives that begin at column 0 so POD blocks are treated as trivia for the lookahead (but `=cut` without a preceding POD start is not treated as trivia), and the scanner will skip until a matching `=cut` before deciding autoquoting.
- Kept the brace-autoquote behavior intact while making the fat-comma (`=>`) decision safer across intervening trivia by preserving the two-character lookahead when needed.
- Added focused BDD-style regressions in `crates/tree-sitter-perl-c/tests/bdd_workflows.rs` including helpers to collect `autoquoted_bareword` captures and node kinds, plus tests covering comment/POD interactions for both fat-comma and brace autoquoting scenarios.

### Testing

- Ran `cargo test -p tree-sitter-perl-c` and all unit and BDD tests in the crate completed successfully (`test` profile finished with all tests passing).
- Ran formatting via `cargo fmt --all` as part of verification.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb278ac7308333a82402adc2592546)

### DIFF
--- a/crates/tree-sitter-perl-c/c-src/scanner.c
+++ b/crates/tree-sitter-perl-c/c-src/scanner.c
@@ -917,19 +917,85 @@ bool tree_sitter_perl_external_scanner_scan(void *payload, TSLexer *lexer,
 
     // NOTE - TS is annoying about skipping chars after you've hit done
     // mark_end, so we have to do the regular advance so our token actually shows up
-    while (is_tsp_whitespace(c) || c == '#') {
+    bool has_two_char_lookahead = false;
+    while (1) {
       while (is_tsp_whitespace(c)) ADVANCE_C;
       // now we need to skip comments - we get in a funny way if we have a quotelike
       // operator followed by a comment as the quote char
       if (c == '#') {
         ADVANCE_C;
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6146 implements same feature. Your work was validated; any novel tests/edge-cases in your diff can be extracted as follow-up PRs if worth.

PR #6146, opened 2026-04-24
fix(tree-sitter-perl-c): harden autoquoted scanner handling around comments and POD (#6532)

EffortlessSteven:
### Motivation

- The vendored C scanner could mis-handle gaps between an identifier and the operator that decides autoquoted tokens, producing incorrect tokenization around comments and POD-like blocks.
- This change tightens the autoquote lookahead so `TOKEN_FAT_COMMA_AUTOQUOTED` and `TOKEN_BRACE_AUTOQUOTED` are resolved correctly in those edge cases.

### Description

- Added `skip_line_comment` to consume `#`-style comments during the autoquoted lookahead, ensuring comment gaps do not confuse the decision logic in `scanner.c`.
- Implemented `skip_pod_like_block` which treats `=<alpha>` at column 0 as POD-like content to be skipped up to a terminating `=cut`, and preserved non-POD `=` for fat-comma matching.
- Adjusted the autoquote lookahead loop to use the new helpers and to distinguish consumed-but-non-POD `=` from real POD blocks so that `=>` detection remains correct.
- Added a narrow guard so the POD tokenization path does not preempt fat-comma autoquote handling, and added focused regression tests in `crates/tree-sitter-perl-c/tests/scanner_autoquote_regressions.rs` for comment and POD gaps.

### Testing

- Ran `cargo test -p tree-sitter-perl-c --test scanner_autoquote_regressions` and the three new regression tests passed.
- Ran the full crate test suite with `cargo test -p tree-sitter-perl-c` which passed, and `cargo check --all-targets -p tree-sitter-perl-c` which also passed.
- Ran formatting via `cargo xtask fmt` as part of verification and completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb278ac7308333a82402adc2592546)


EffortlessSteven:
### Review: harden autoquoted scanner handling around comments and POD

**Substantive C-level fix in tree-sitter-perl-c's scanner.c**. The autoquoted-bareword scanner previously didn't skip POD blocks or line comments while looking ahead -- so a fat-arrow context like:

```perl
%hash = (
    # this is a comment
    key => 'value',
    =head1
    docs
    =cut
    other_key => 'value',
);
```

...would parse incorrectly because the lookahead tripped over the `#` or `=head1` lines.

**The fix adds two helpers**:

1. **`skip_line_comment`**: standard `#` -> end-of-line skipper, threaded through the autoquoted lookahead.

2. **`skip_pod_like_block`**: detects `=<alpha>` at column 0 (POD start), skips until a line beginning with `=cut` followed by whitespace/newline. Returns `true` if it skipped a POD block. The `column == 0` check is correct -- POD blocks must start at column 0 per perlpod.

**Correctness checks on `skip_pod_like_block`**:

- The `iswalpha(*c)` check after the leading `=` correctly distinguishes POD (`=head1`, `=item`) from non-POD `=` (`==`, `=~`). If the char after `=` isn't alpha, set `consumed_non_pod_equals = true` so the caller can recover -- *but* the first `=` was already consumed via `lexer->advance`. Worth confirming the caller's recovery handles this single-char-consumed state correctly. If the caller assumes nothing was consumed, this is a subtle bug.

- The `=cut` matching uses a hand-rolled char loop (`marker_i` walks through `"cut"`). After matching all three chars, it requires the next char to be one of `\n`, `\r`, ` `, `\t`. Standard POD lexing -- `=cutxyz` is *not* a valid `=cut` marker. Correct.

- The "no terminating `=cut`" case "treats this as consuming to EOF" -- per the doc-comment. This matches how Perl's own lexer treats unterminated POD (consume to EOF, no error). Behavioral parity with perl's tokenizer.

1. The hand-rolled `cut` matcher works but is brittle. If someone later adds `=end` or `=cuts` as another terminator, the pattern needs duplicating. Worth a small `match_marker(lexer, c, "cut")` helper that returns the char after the marker. Reducing surface area for future bugs.

2. The PR diff doesn't show the *callers* of `skip_line_comment` / `skip_pod_like_block` -- worth eyeballing those in the full diff to confirm they thread the `*c` lookahead variable correctly. If a caller forgets to refresh `*c` after calling these helpers, the autoquoted lookahead state is corrupted.

3. Tree-sitter scanners are notoriously hard to test directly. If this PR adds corpus-level tests (Perl source files exercising fat-arrow contexts with embedded comments and POD), great. If not, that's a follow-up.

**Verdict**: real correctness fix at the C scanner level, lexically aligned with Perl's own POD/comment rules. Pre-existing master CI failures are unrelated; #6495 fixes both halves of the master block.

EffortlessSteven:

PR #6166, opened 2026-04-24
feat(dap-status): promote deferred attach/variables/evaluate metrics into the DAP scorecard (#6524)

EffortlessSteven:
### Motivation

- The DAP status page only reported launch success and left attach/session correctness, deep truncation/pagination, and a memory baseline deferred; those gaps hide next-failure modes behind "Deferred Metrics".
- Promote those deferred assertions into an automated, marker-friendly scorecard so CI and the status page surface real measured/skip outcomes rather than TODO text.

### Description

- Extend the `dap_scorecard_harness` test (`crates/perl-dap/tests/dap_scorecard_harness.rs`) to measure: launch success/latency, TCP loopback attach success rate, real-session variables pane correctness, evaluate correctness in a real session, deep truncation/pagination verification (paged variables), and a best-effort memory footprint proxy; the harness emits a machine-readable JSON receipt (`target/dap_scorecard_receipt.json`) and marker-friendly markdown blocks.
- Add defensive behaviour in the harness to `SKIP` gracefully when `perl` is not available and to `SKIP` deep-pagination checks when no qualifying indexed variable is exposed by the session.
- Teach `xtask` status generator (`xtask/src/tasks/update_status/dap.rs`) to read the receipt and render both `DAP_LAUNCH_SCORECARD` and a new `DAP_SESSION_SCORECARD` marker block (or sensible SKIP messages when no receipt is present), and update the coordinator (`xtask/src/tasks/update_status/mod.rs`) to pass the project root through to the generator.
- Update the published status document (`docs/project/status/dap.md`) to include the new session correctness block and refreshed test counts produced by `xtask`.

### Testing

- Ran `cargo test -p perl-dap --test dap_scorecard_harness -- --nocapture` and the scorecard harness passed locally (produced receipt and printed marker blocks).
- Ran `cargo xtask update-status --only dap --write` and `cargo xtask update-status --only dap` to regenerate and verify `docs/project/status/dap.md`, and both succeeded (file updated and subsequent check reported up-to-date).
- Ran `cargo check --workspace --all-targets` and `cargo fmt --all`, both completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb3980f0448333a7cc04be82e50249)


EffortlessSteven:
**Standards review: PASS**

Tight scope (DAP scorecard harness + xtask coordinator only). No banned patterns (unwrap/expect/panic/todo/dbg). Formatting clean. Tests pass locally. Receipt generation properly handles missing perl and graceful skip paths. Root path correctly threaded through xtask coordinator. Ready for deep review.

EffortlessSteven:
## Deep review — APPROVED with fixes

### What was verified

- **Full protocol flow in `probe_attach`**: `TcpListener` → fake `stopped` event via `frame()` → adapter `initialize` → `attach` → `wait_for_event(stopped)` → `disconnect` → `wait_for_event(terminated)`. Traced through `handle_attach` → `TcpAttachSession::connect` → reader thread → `DapEvent::Stopped` forwarding → `handle_disconnect` → `clear_active_session_state` → `TcpAttachSession::disconnect` → `shutdown(Both)` → server read-loop gets `Ok(0)`. Ordering and socket lifecycle are correct.
- **Receipt path consistency**: harness writes to `<workspace_root>/target/dap_scorecard_receipt.json` (two `..` levels from `CARGO_MANIFEST_DIR`); xtask reads `root.join("target/dap_scorecard_receipt.json")`. Paths agree.
- **xtask deserialization tolerances**: `RateMetric` omits `details: Vec<FixtureResult>` — serde ignores unknown fields without `deny_unknown_fields`. Correct.
- **Percentile math**: nearest-rank implementation is correct including n=1 edge.
- **floor vs ceil parity**: `status_for_rate` (display, floor division) and the test assertion (ceil threshold) agree for all realistic pass/total combinations with 80% threshold.
- **Session metrics assertions**: non-vacuous — `variables.status == "PASS"` depends on the adapter returning named variables; `evaluate` checks the result contains "42" against a live Perl session.
- **Security**: no user input flows into any shell command or file path. TCP server bound to `127.0.0.1:0` only. Temp file content is a hardcoded literal. CLEAN.

### Bugs fixed (3 commits pushed to this branch)

**1. `MemoryMetric` duplicate struct removed**
`MemoryMetric` was byte-for-byte identical to `BinaryMetric`. The harness serialized `memory` as `MemoryMetric`; xtask deserialized it as `BinaryMetric`. JSON roundtripped fine by accident (same field names), but any future field addition to one struct would silently break deserialization. Consolidated to `BinaryMetric` everywhere in the harness.

**2. Attach threshold math fixed**
With 3 attach attempts and 80% threshold, `div_ceil(3 × 80 / 100) = div_ceil(2.4) = 3`, which silently enforces **100%** — one failure causes the test to fail, but the status page displays "≥ 80%". The mismatch is a correctness lie in the scorecard semantics. Fixed by increasing from 3 to 5 probes: `div_ceil(5 × 80 / 100) = 4`, so exactly 1 failure is tolerated, matching the stated 80% intent.

**3. Singular/plural grammar in variables detail string**
"globals scope returned 1 named variables" → "1 named variable" / "N named variables".

### Remaining observation (pre-existing, not a defect in this PR)

The `deep_pagination` check will permanently SKIP on the current adapter because `@big` (500 elements) is not exposed with `indexedVariables >= 200` in the globals scope — the adapter appears to surface only scalar `our` variables. The harness SKIP path is correct and the assertion allows PASS or SKIP. This is a pre-existing adapter gap worth a follow-up issue.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-6166",
  "current_version": {
    "sha": "0a3dd550aa3ceffa081528961d6d4a4acf3b355a",
    "updated_at": "2026-04-25T05:33:38Z"
  },
  "label_bindings": [
    {
      "label": "deep-reviewed",
      "bound_at_version": "0a3dd550aa3ceffa081528961d6d4a4acf3b355a",
      "bound_at_timestamp": "2026-04-25T05:34:20Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Deep Review Wrapup

**Summary:** reviewer-deep completed correctness review of PR #6166 (dap-status scorecard promotion). 3 bugs found and fixed forward; `deep-reviewed` label applied.

### Ending State

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/6166
- Changes made: 3 fix commits pushed to branch `codex/promote-deferred-metrics-to-dap-scorecard-mye88x`
  - `5bdbbe4b1` — Remove `MemoryMetric` duplicate struct; increase attach probes from 3→5 to fix 80% threshold math
  - `0a3dd550a` — Correct singular/plural in variables metric message ("1 named variables" → "1 named variable")
- Files touched: `crates/perl-dap/tests/dap_scorecard_harness.rs`
- Status: **approved** — `deep-reviewed` label set, label receipt written at SHA `0a3dd550a`
- Next step: route to green-ci for CI freshness check

### Retrospective

**Hardest part:** The threshold math bug (`div_ceil(3 × 80 / 100) = 3 = 100%`) required careful arithmetic to confirm — the claimed "80%" and the enforced "100%" look identical when all tests pass, making it invisible without deliberately checking the integer ceiling math for small sample sizes. The lesson: any `div_ceil` assertion where n < 5 with an 80% threshold is effectively a 100% assertion.

**Surprising finding:** `MemoryMetric` and `BinaryMetric` in the harness were byte-for-byte identical structs with different names, but the xtask deserializer used only `BinaryMetric` for the `memory` field. Serde silently bridged the structural mismatch. This would have been a silent deserialization failure the moment either struct diverged. The identical-struct pattern is a maintenance trap that the compiler can't catch.

**What would help next time:** For scorecard harnesses, explicitly document the minimum sample count required for each threshold percentage (e.g., "80% threshold needs n≥5 for `div_ceil` to be meaningful"). This would make the threshold mismatch class of bug visible at code-review time without arithmetic.

- **Logical next:** green-ci on PR #6166 to verify CI is green on current HEAD `0a3dd550a`
- **Related:** the `deep_pagination` SKIP is permanent until the DAP adapter exposes `@big` (500-element array) with `indexedVariables >= 200` in the globals scope — only scalar `our` variables are currently surfaced. Worth a follow-up scout issue against `crates/perl-dap/src/debug_adapter/`.
- **Gotcha:** `docs/project/status/dap.md` shows "3/3 (100%)" for attach in the checked-in version — this will become stale (should be 5/5) once the harness is re-run post-merge. This is expected: the doc regenerates from `target/dap_scorecard_receipt.json` which is gitignored.
- **Trap:** `target/dap_scorecard_receipt.json` is gitignored — CI never regenerates it. The dap.md values are always from the last local developer run, not from CI. This is the design intent (best-effort scorecard), but reviewers should not treat dap.md metric values as CI-verified.
- Confidence for next: **high**

EffortlessSteven:
## Diff Audit: CLEAN

Cumulative diff is coherent, spec-aligned, and ready for merge.

**Scope**: Tight to declared changes per issue references.
**Cleanliness**: No artifacts, debug code, or commented blocks.
**Test coverage**: All PRs add focused regression or edge-case tests.
**Commit coherence**: Each PR follows fix → tests → optional mechanical gates pattern.

### Verdict: CLEAN ✓

---

EffortlessSteven:
### Review: promote deferred attach/variables/evaluate metrics into the DAP scorecard

**Substantive scorecard expansion** -- broadens `dap_scorecard_harness` from "launch success rate only" to a comprehensive 5-metric scorecard (launch, attach, variables, evaluate, deep_pagination) and threads a structured JSON receipt (`target/dap_scorecard_receipt.json`) through to `docs/project/status/dap.md` rendering.

**Architecture is right**:

```
dap_scorecard_harness  (test)
    -> probes 5 metrics, writes ScorecardReceipt JSON to target/
    -> asserts threshold satisfaction
        |
        v
xtask update-status (build/CI step)
    -> reads ScorecardReceipt JSON
    -> renders DAP_LAUNCH_SCORECARD and DAP_SESSION_SCORECARD tables in dap.md
    -> falls back to "receipt missing" SKIP rows if JSON absent
```

This is the standard "test produces evidence; doc generator consumes evidence" decoupling that survives partial-runs (CI without `cargo test`, or local without xtask). The fallback messages even tell future contributors which command to run -- nice ergonomics.

**Concrete improvements I noticed**:

1. **`response_success` helper** factors out a 12-line response-decoder that was previously duplicated for `initialize`, `launch`, `attach`, `disconnect`. The new helper:

   ```rust
   fn response_success(response: DapMessage, command: &str) -> Result<Option<Value>, String> {
       match response {
           DapMessage::Response { success, command: actual, body, message, .. } => {
               if actual != command { return Err(format!("expected `{command}` response, got `{actual}`")); }
               if !success { return Err(format!("command `{command}` failed: {}", message.unwrap_or_else(|| "<no message>".into()))); }
               Ok(body)
           }
           _ => Err(format!("expected response message for `{command}`")),
       }
   }
   ```

   Replaces ~60 lines of `match` boilerplate. Returns the body so callers that need to inspect it can; otherwise discardable.

2. **Sample-size correction in attach probe**:

   ```rust
   // Run 5 attach probes so the 80 % threshold (div_ceil(5 * 80 / 100) = 4) is
   // meaningful: exactly 1 failure is tolerated.  With n < 5 the ceil rounding
   // makes the effective threshold 100 %, defeating the intent.
   ```

   This comment is exactly the kind of thing future maintainers will thank you for. Without it, a "let me cut to 3 attach probes to save CI time" PR would silently change the gate from "tolerate 1/5 failures" to "tolerate 0/3 failures." Worth highlighting in the review.

3. **Threshold semantics**:

   ```rust
   let launch_threshold =
       (receipt.launch.total * usize::from(receipt.launch.threshold_pct)).div_ceil(100);
   ```

   `div_ceil(100)` is the right rounding for "≥80%": `(5 * 80).div_ceil(100) = 4`, `(4 * 80).div_ceil(100) = 4` (i.e. 100% required for 4 fixtures), `(10 * 80).div_ceil(100) = 8`. Conservative and predictable.

**Concerns / issues**:

1. **Receipt path is computed via `CARGO_MANIFEST_DIR/../../target/`** -- this hardcodes a workspace-relative assumption. If `perl-dap` is ever cloned/symlinked into a different workspace structure (or built standalone), the receipt lands in the wrong place. Cleaner would be `cargo metadata --format-version 1 | jq -r .target_directory` or threading `CARGO_TARGET_DIR` env var through. Not blocking but worth a follow-up.

2. **Receipt is in `target/`** -- which means `cargo clean` deletes it. If `xtask update-status` runs *after* a clean (common on CI), the doc generator sees no receipt and emits the "receipt missing" SKIP rows. The resulting `dap.md` then diff's against master, and `--check` mode complains about drift. Either:
   - Make the doc generator's "receipt missing" path produce a stable string equivalent to the last-good receipt (impossible without persistence), OR
   - Move the receipt to `.ci/dap_scorecard_receipt.json` (gitignored but durable across clean), OR  
   - Have CI explicitly run the test before `xtask update-status --check`.

3. **`probe_attach` server thread doesn't synchronize with the test's connect attempt.** It calls `listener.accept()` on a thread, so a slow test runner could connect *before* `accept()` returns. Probably fine because `accept` is fast on Linux/macOS, but a flake-prone pattern on heavily-loaded CI.

4. **`probe_session_metrics()` gets one shot at each metric.** If the harness's underlying real-Perl session flakes (a real risk on CI), the receipt records `FAIL` and the assertion trips, even though the *real* feature works in 99% of runs. Worth either:
   - Retry-with-backoff inside `probe_session_metrics` (3 attempts, fail only if all 3 fail).
   - Tagging the test as `#[ignore]` for `pr-fast` and only running in `ci-gate` / `ci-full` where flake retries are budgeted.

5. **The `print_marker_friendly_summary` and `launch_table_from_receipt` markdown-emitting code is duplicated** -- once in the test (for stderr output) and once in `dap.rs` (for the docs). Drift between the two is a real risk. Consider extracting the table-rendering logic into a shared module under `crates/perl-dap/src/scorecard.rs` (or a `perl-dap-scorecard` sub-crate) that both the test and the doc generator can import. Future refactor.

**Verdict**: real expansion of scorecard coverage from "launch only" to "launch + attach + variables + evaluate + deep_pagination," well-architected via the test-emits-JSON / xtask-consumes-JSON pattern, with thoughtful threshold semantics and the kind of inline-comment self-documentation that prevents future regressions. Some flake-resilience and target-path concerns worth following up but not blocking.

EffortlessSteven:
**Diff audit: CLEAN**

✓ Scope: DAP scorecard feature (promote deferred metrics into automated assertions)
✓ Files: 4 files (dap test, docs/status, xtask update)
✓ Intentional eprintln! calls for scorecard output (not debug artifacts)
✓ Issue #6524 aligns with implementation

**Note:** Stale-base detected (53 commits behind master). Recommend `gh pr update-branch` before merge.

EffortlessSteven:

EffortlessSteven:
**refactor-planner**: refactor recommended.

PR is 1008 lines (659 add / 161 del). The bulk is unavoidable test scaffolding (TCP listener, DAP session metrics, JSON receipt round-trip), but there are three concrete consolidation opportunities in `xtask/src/tasks/update_status/dap.rs` that reduce noise without changing behavior.

### Findings

**1. Duplicated percent calc in `format_rate` and `status_for_rate`** (`xtask/src/tasks/update_status/dap.rs:90-107`)

Both functions independently compute `(passed * 100) / total` with the same `total == 0` guard. Same pattern repeats inline in `print_marker_friendly_summary` (`crates/perl-dap/tests/dap_scorecard_harness.rs:412-423` and `:438-451` for launch and attach).

Add a single `RateMetric::pct(&self) -> Option<usize>` (returns `None` when `total == 0`) and derive `format_rate`, `status_for_rate`, and the harness summary lines from it.

**2. Triplicated SKIP fallback table strings** (`xtask/src/tasks/update_status/dap.rs:120-128` and `:160-168`)

Both `launch_table_from_receipt` and `session_table_from_receipt` carry hardcoded multi-line markdown table strings for the receipt-missing case. They are nearly identical in shape and any future schema change requires touching both.

Refactor to one `render_metric_table(headers, rows)` helper that takes a list of `(metric_name, value, target, status)` tuples; both callers build their row list (using "SKIP" + receipt-missing message when receipt is absent) and call the helper. Eliminates the dual hardcoded table blocks.

**3. SKIP-receipt construction repeated three times** (`crates/perl-dap/tests/dap_scorecard_harness.rs:476-503`, plus the manual SKIP fields in the session-metric error path at `:573-579`)

The pattern `BinaryMetric { status: "SKIP", detail: "perl not on PATH".to_string() }` appears 4x with the same string, and the empty `RateMetric` skeleton appears 2x. Add:
- `BinaryMetric::skipped(reason: &str) -> Self`
- `RateMetric::skipped(threshold_pct: u8) -> Self`

This makes the perl-unavailable branch a one-liner per field.

### Suggested plan for green-refactor

1. In `crates/perl-dap/tests/dap_scorecard_harness.rs`, add `BinaryMetric::skipped`, `RateMetric::skipped`, and `RateMetric::pct(&self) -> Option<usize>` constructors/methods.
2. Replace the 4x `BinaryMetric { status: "SKIP", ... }` literals and 2x empty `RateMetric { passed: 0, total: 0, ... }` literals in the perl-unavailable arm.
3. In `xtask/src/tasks/update_status/dap.rs`, add a parallel `RateMetric::pct` (deriving from `passed`, `total`); collapse `format_rate` + `status_for_rate` + their inline equivalents to use it.
4. Extract `render_metric_table(rows: &[(name, value, target, status)]) -> String` and rebuild both `launch_table_from_receipt` and `session_table_from_receipt` from row vectors.

### Files affected
- `crates/perl-dap/tests/dap_scorecard_harness.rs`
- `xtask/src/tasks/update_status/dap.rs`

### Risk
Low — receipt schema and table contents stay identical. The existing `test_generate_dap_status_roundtrip` test pins the marker block format, so any drift is caught.

### Estimated savings
~40-60 lines net deletion across both files; future schema changes touch one location instead of three.

PR #6171, opened 2026-04-24
perf(perl-dap): expand benchmarks to live session launch/attach/stack/variables/evaluate paths

EffortlessSteven:
### Motivation

- The existing `dap_benchmarks` only exercised phase‑1 config/platform work and missed hot paths exercised during live DAP sessions (launch/attach/stepping/variables/evaluate). 
- Adding live‑session benchmarks provides stable, diffable signals for runtime performance regressions across the adapter's core workflows.

### Description

- Added Phase 3 live‑session benchmark groups (`dap_live_launch`, `dap_live_attach`, `dap_live_session`) and stable benchmark names such as `launch_cold`, `launch_warm`, `attach_loopback`, `set_breakpoints_100`, `step_continue_p95`, `stack_trace_live`, `variables_root`, `variables_child_page`, `evaluate_safe_blocked`, and `evaluate_live_simple` in `crates/perl-dap/benches/dap_benchmarks.rs`.
- Implemented helper functions to create temporary Perl scripts and drive adapter flows (`write_script`, `initialize_adapter`, `launch_adapter`, `disconnect_adapter`, and small response parsing helpers) so the benchmarks exercise real adapter code paths without modifying production logic.
- Kept all existing configuration/platform/session benchmarks and updated the bench registration to include the new `live_session_benches` group in `criterion_main!` so the new groups run alongside the old ones.
- Documented how to run the expanded suite and listed the stable live‑session benchmark names in `crates/perl-dap/README.md` under a new `## Benchmarks` section.

### Testing

- Ran `cargo check --all-targets -p perl-dap` which completed successfully with warnings unrelated to the changes.
- Built the bench executable with `cargo bench -p perl-dap --bench dap_benchmarks --no-run` which produced the bench binary successfully.
- Executed the benchmarks (short run) with `cargo bench -p perl-dap --bench dap_benchmarks -- --measurement-time 1 --sample-size 10` and observed the new live groups execute (`dap_live_launch`, `dap_live_attach`, `dap_live_session`) without errors.
- Ran formatting via `cargo xtask fmt` as part of verification which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb39873f008333875f4b7139515097)


EffortlessSteven:
### Review: extend DAP benchmarks to live session paths

**Solid expansion** of the DAP criterion suite from "Phase 1 config/platform only" to comprehensive live-session coverage with stable, diff-friendly benchmark function names. The 10 new benchmark functions cover the full DAP request lifecycle:

| Benchmark | Path measured |
|---|---|
| `launch_cold` | tempdir + write_script + initialize + launch + disconnect (full cold path) |
| `launch_warm` | reused script, just initialize + launch + disconnect |
| `attach_loopback` | TCP listener spawn + initialize + attach + disconnect |
| `set_breakpoints_100` | launch + 100-breakpoint setBreakpoints request |
| `step_continue_p95` | 40 alternating continue/next requests |
| `stack_trace_live` | launch + stackTrace |
| `variables_root` | launch + scopes + variables(root, count=100) |
| `variables_child_page` | launch + scopes + variables(root) + variables(child, start=20, count=50) |
| `evaluate_safe_blocked` | launch + evaluate("system('echo blocked')") (safe-mode rejection path) |
| `evaluate_live_simple` | launch + evaluate("$x") (success path) |

**What's done well**:

1. **Stable benchmark names for diffability.** The README documents:
   ```
   Live-session benchmark function names:
   - launch_cold
   - launch_warm
   ...
   ```
   This is exactly right -- criterion's HTML output and `cargo bench` baseline comparisons key on benchmark names, so any rename invalidates baselines. Pinning the names in README acts as a contract.

2. **`perl_available()` early-skip pattern**:
   ```rust
   if !perl_available() {
       group.bench_function("set_breakpoints_100", |b| b.iter(|| black_box(())));
       group.bench_function("step_continue_p95", |b| b.iter(|| black_box(())));
       // ... emits empty benchmarks for all 7 names
       group.finish();
       return;
   }
   ```
   Rather than skipping the bench entirely (which would make criterion delete the baseline), the bench emits a no-op that preserves the name. Subsequent `cargo bench` diffs against baselines built on Perl-equipped CI will show "much slower" / "failed to record" rather than "missing" -- which is the right signal for "this CI box doesn't have perl, baseline is stale here."

3. **Deterministic Perl env injection in `launch_adapter`**:
   ```rust
   "env": {
       "PERL_PERTURB_KEYS": "0",
       "PERL_HASH_SEED": "0",
       "LC_ALL": "C",
       "TZ": "UTC"
   }
   ```
   Pinning hash randomization, locale, and timezone removes 3-4 known sources of variance in benchmark numbers. Worth highlighting because most "let me bench Perl from Rust" attempts forget at least one of these.

4. **`first_scope_reference` / `first_child_reference` helpers extract real reference IDs from the response** rather than hardcoding magic numbers, with `unwrap_or` fallbacks for when the response shape is unexpected. The fallback values (11, 1102) match the DAP defaults.

**Concerns**:

1. **`launch_cold` measures multiple things at once.** Each iteration creates a tempdir, writes a 120-line script, spawns a new adapter, initializes, launches, disconnects. The "cold" claim is real but the *delta* between cold and warm only isolates the script-write and adapter-spawn cost -- it doesn't isolate process-start cost (which is what most DAP "cold launch" perf work cares about). Consider:
   - `launch_cold_no_script` that pre-writes the script outside the iter
   - `launch_cold_clean_workspace` that writes to a clean tempdir
   This would let regressions in process-start overhead vs. fs overhead vs. adapter-init be diagnosed separately.

2. **Tempdir lifetime in `launch_warm` and `benchmark_live_session_paths`**:
   ```rust
   let warm_dir = tempdir();
   if let Ok(dir) = warm_dir {
       let warm_script = dir.path().join("warm_launch_bench.pl");
       // ... bench iterates here, dir is in scope
   }
   ```
   Good -- the `dir` (TempDir guard) stays alive for the bench's lifetime. The pattern with `let _fixture_guard = fixture_dir;` later does the same. Fine.

3. **`step_continue_p95`** doesn't actually measure p95 (criterion does that automatically across iterations). The name is slightly misleading -- it suggests the bench *itself* picks the p95 of 40 step requests. Consider renaming to `step_continue_40x` or `step_continue_burst` to avoid confusion with criterion's own statistical output.

4. **`variables_child_page`'s `unwrap_or(1102)` fallback for `child_ref`** -- if the response shape changes (e.g. `variablesReference` becomes `varRef`), the bench silently falls back to a fabricated reference ID, and the call goes through some "child not found" path that's faster than the real one. The bench would record a *faster* number, masking a regression. Worth either:
   - Asserting the fallback isn't triggered in a setup phase (`b.iter_with_setup` could verify), OR
   - Failing loudly: `panic!("benchmark setup: failed to find a child variablesReference")`.

5. **`benchmark_live_attach_loopback` server thread** sleeps 15ms then drops the stream. If the criterion runner is faster than that, the attach completes against a half-alive server. If slower, it completes against a dead server. Either way, the bench captures the "attach + immediate-disconnect" path, not a sustained-attach path. Consider keeping the server alive for the duration of `b.iter_with_setup` and only joining the thread in teardown.

6. **README documentation drift risk.** The bench-name list in README must match the literal `bench_function` calls. A rename in the .rs file without a README update silently lies to users. Worth either:
   - A `#[test]` in the same crate that reads the README and asserts each documented name appears as a `bench_function` call in the .rs file.
   - Or, a generated docs section: `cargo bench -p perl-dap --bench dap_benchmarks -- --list` produces the canonical list; have a CI step check README is in sync.

**Verdict**: real, useful expansion of bench coverage from setup-only to full live-session paths, with thoughtful skip-without-perl handling and deterministic env pinning. The minor concerns (cold/warm decomposition, p95 naming, child_ref fallback masking regressions) are refinements not blockers.

EffortlessSteven:

EffortlessSteven:

EffortlessSteven:

PR #6182, opened 2026-04-24
test(perl-dap): harden attach and bridge lifecycle reliability (#6521)

EffortlessSteven:
### Motivation

- Improve reliability and observability of TCP attach lifecycles by promoting a deferred/connect-time assumption into a hardened behavior and by keeping bridge-mode behavior explicit and testable.

### Description

- Ensure `TcpAttachSession::disconnect()` always clears the connected flag so reader-backed sessions cannot remain incorrectly marked connected after shutdown (moved connected clear out of conditional).
- Make `TcpAttachSession::start_reader()` clone the `TcpStream` for the reader thread instead of taking ownership, preserving the write/disconnect handle so disconnect + reconnect is possible.
- Add E2E attach tests in `crates/perl-dap/tests/dap_attach_e2e.rs` covering `stopOnEntry` sequencing and an actionable attach-timeout failure message assertion.
- Add TCP session tests in `crates/perl-dap/tests/tcp_attach_tests.rs` for unreachable-endpoint timeouts, stopped+terminated event emission, and disconnect-after-reader allowing reconnect.
- Add bridge-mode guard tests in `crates/perl-dap/tests/bridge_integration_tests.rs` ensuring `proxy_messages()` fails when not spawned and `shutdown()` is idempotent/no-op without a child.

### Testing

- Ran `cargo test -p perl-dap --test dap_attach_e2e` and all tests passed.
- Ran `cargo test -p perl-dap --test tcp_attach_tests` and all tests passed.
- Ran `cargo test -p perl-dap --test bridge_integration_tests` and all tests passed.
- Ran `cargo xtask fmt` to format changes (completed successfully).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb3992f3188333947ec51a9b1a582a)


EffortlessSteven:
### Review: harden attach and bridge lifecycle reliability

**Two real correctness fixes** to `crates/perl-dap/src/tcp_attach.rs`, plus comprehensive lifecycle tests covering both fixes and adjacent behavior.

**Production-code fixes**:

1. **`disconnect()` always resets the connected flag**, even when no stream was ever established:

   ```rust
   pub fn disconnect(&mut self) -> Result<()> {
       if let Some(stream) = self.stream.take() {
           stream.shutdown(std::net::Shutdown::Both)?;
           tracing::info!("Disconnected from Perl debugger");
       }
       *self.connected.lock().unwrap_or_else(|e| e.into_inner()) = false;  // moved out of `if let`
       Ok(())
   }
   ```

   Previously the `connected = false` write was *inside* the `if let Some(stream)` block, meaning a `disconnect()` call after `start_reader()` (which moved the stream out -- see fix #2) would not reset the flag. The flag would stay `true` forever, breaking idempotent disconnect semantics and confusing higher-level code that polls `is_connected()` to decide reconnection eligibility.

   Subtle but real -- the test `test_tcp_attach_disconnect_after_reader_allows_reconnect` confirms the fix end-to-end: connect, start_reader, disconnect, reconnect, disconnect, all `is_connected()` checks land correctly.

2. **`start_reader()` clones the stream instead of moving it out**:

   ```rust
   let stream = self
       .stream
       .as_ref()
       .context("No stream available")?
       .try_clone()
       .context("Failed to clone TCP stream for reader thread")?;
   ```

   Previously `self.stream.take()` would consume the stream, leaving `self.stream = None`. Then `disconnect()` couldn't `shutdown()` the stream (it was on the reader thread), and the reader thread had to terminate the connection itself -- creating a hairy ownership story.

   With `try_clone`, both the reader thread and the parent session hold a handle to the same underlying socket. `disconnect()` can call `shutdown(Shutdown::Both)` on the parent's clone, which causes the reader's clone to receive `Ok(0)` (EOF) and clean up. Standard pattern for Rust TCP threading.

   `try_clone()` returning `Err` is gracefully handled with `.context(...)?` -- it surfaces as a typed error rather than a panic.

**Test coverage is exhaustive**:

`tests/tcp_attach_tests.rs`:
- `test_tcp_attach_connect_timeout_for_unreachable_endpoint`: routes to a TEST-NET-1 reserved IP (`203.0.113.1`) with a 50ms timeout. Verifies the connect-timeout error path AND that `is_connected()` returns false after a failed connect.
- `test_tcp_attach_reader_emits_stopped_and_terminated_events`: end-to-end test with a fake DAP server that emits framed `stopped` then `terminated` events. Asserts the reader thread correctly demultiplexes them onto the event channel.
- `test_tcp_attach_disconnect_after_reader_allows_reconnect`: the *direct test* for fix #1 -- proves that `disconnect()` correctly resets the connected flag even after `start_reader()` has been called.

`tests/dap_attach_e2e.rs`:
- `dap_attach_e2e_tcp_loopback_stop_on_entry_and_server_stopped`: `stopOnEntry: true` should produce two `stopped` events -- one with `reason="entry"` (synthetic) and one with `reason="pause"` (from the server). Pins the synthesis-and-passthrough behavior.
- `dap_attach_e2e_tcp_attach_timeout_returns_actionable_message`: deliberately drops the listener to force a connection refusal, then asserts the error message contains *all* the actionable diagnostic fields:
  ```
  message.contains("Cannot attach to Perl debugger at 127.0.0.1")
  message.contains("(250ms timeout)")
  message.contains("RemotePort=127.0.0.1")
  ```
  This is the right shape -- the user-facing error must give them enough to fix it (host, port, timeout, the env-var they need to set).

`tests/bridge_integration_tests.rs`:
- `test_bridge_proxy_requires_spawn`: `proxy_messages` without `spawn()` returns a typed `"Child process not spawned"` error.
- `test_bridge_shutdown_without_spawn_is_noop`: `shutdown()` is idempotent on an unspawned bridge -- can be called twice without error. Standard cleanup-path resilience.

1. **Race in `test_tcp_attach_disconnect_after_reader_allows_reconnect`**: server2's `(_socket, _) = listener2.accept(); sleep(200ms);` runs in a thread that's spawned but not synchronized with the connect attempt. If the test runner is slow enough that the connect happens *before* `accept()` arrives, the connect would block. Probably fine in practice because the `accept` is fast, but worth a `Barrier` or pre-listening synchronization for full determinism.

2. **`std::net::Shutdown::Both` is correct for graceful shutdown**, but on some Linux kernels `shutdown` on a half-closed socket can produce `ENOTCONN` and surface as `Err`. The current `?` propagates that. Worth either swallowing `ENOTCONN` (since "shutdown failed because already shut down" is fine) or downgrading to a `tracing::debug!`. Minor.

3. **The fake TCP debugger server pattern repeats** across multiple tests (bind listener, accept in thread, write framed messages, sleep, drop). Consider extracting to a `tests/common/fake_dap_server.rs` helper that exposes a `FakeDapServer::new()` returning `(server_handle, port)` to reduce 200+ lines of similar boilerplate. Future addition.

**Verdict**: real correctness fixes for two distinct lifecycle bugs, well-tested at multiple levels (unit tcp_attach -> e2e attach with fake server -> bridge integration). Pre-existing master CI failures are unrelated; #6495 fixes both halves of the master block.

EffortlessSteven:

EffortlessSteven:

EffortlessSteven:

PR #6184, opened 2026-04-24
perf(perl-dap): make framed debugger-output capture marker-aware and cheaper (#0000)

EffortlessSteven:
### Motivation
- Framed debugger-output capture was cloning and re-normalizing the entire recent-output buffer on every poll, causing unnecessary CPU and lock contention during frequent framed queries.
- The change aims to make capture incremental and cache-aware while keeping the existing begin/end marker protocol and correctness under cancellation, partial markers, timeouts, and interleaved output.

### Description
- Replace `VecDeque<String>` recent-output storage with `RecentOutputLine { id, raw, normalized }` and add a monotonic `recent_output_line_id` to stamp lines for incremental scans.
- Writer (output reader) now appends pre-normalized, ID-stamped `RecentOutputLine` entries to the ring buffer as lines arrive, avoiding repeated normalization during capture. (`process.rs`)
- Implement `snapshot_recent_output_entries_since` and tail-tracking in `capture_framed_debugger_output` so the capture loop only scans newly appended lines rather than rescanning the whole buffer. (`mod.rs`)
- Add exact marker-boundary matching via `line_contains_marker` to avoid false matches for similarly prefixed markers (e.g. `_20` vs `_200`). (`mod.rs`)
- Provide a test-only `append_recent_output_line` helper and update existing parsing tests to use it rather than mutating the old raw ring buffer. (`parsing.rs`)
- Add focused tests for framed capture: cancellation handling, large-history performance fast-path, and exact marker-boundary behavior. (`parsing.rs`)

### Testing
- Ran formatting: `cargo xtask fmt` — success.
- Static check: `cargo check --all-targets -p perl-dap` — success.
- Unit tests (focused): `cargo test -p perl-dap capture_framed_debugger_output` and `cargo test -p perl-dap recent_output` — both passed (new/updated tests OK).
- Full crate test: `cargo test -p perl-dap` — most unit tests passed, but `dap_e2e_workflow_tests::test_e2e_multi_breakpoint_sequence` failed with an existing assertion mismatch (expected breakpoint line 5, got 4), which appears unrelated to the framed-capture microchanges and mirrors an intermittent/e2e expectation difference observed when running the full suite.

Problem: Framed debugger-output capture repeatedly cloned and normalized the full recent-output ring on every poll, causing unnecessary overhead and marker scan work.
Fix: Store recent output as cached normalized entries with monotonic line ids, then capture framed output by incrementally scanning only newly appended lines with exact marker-boundary matching.
Verification: `cargo test -p perl-dap capture_framed_debugger_output`, `cargo test -p perl-dap recent_output`, and `cargo check --all-targets -p perl-dap` pass; full `cargo test -p perl-dap` shows one failing e2e test noted above.

------

### DIFF
--- a/crates/perl-dap/src/debug_adapter/mod.rs
+++ b/crates/perl-dap/src/debug_adapter/mod.rs
@@ -391,6 +391,13 @@ struct DataBreakpointRecord {
     condition: Option<String>,
 }
 
+#[derive(Debug, Clone)]
+struct RecentOutputLine {
+    id: u64,
+    raw: String,
+    normalized: String,
+}
+
 /// Check if the match is an escape sequence (preceded by backslash)
 fn is_escape_sequence(s: &str, match_start: usize) -> bool {
     if match_start == 0 {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6185 implements same feature. Your work was validated; any novel tests/edge-cases in your diff can be extracted as follow-up PRs if worth.

PR #6185, opened 2026-04-24
perf(perl-dap): make framed debugger-output capture marker-aware and cheaper (#6520)

EffortlessSteven:
### Motivation

- Framed debugger-output capture was cloning and re-normalizing the entire recent-output ring on every poll, causing unnecessary CPU and allocations during live debugger queries.

### Description

- Replace `recent_output: VecDeque<String>` with a marker-aware `RecentOutputBuffer` storing `RecentOutputLine { id, raw, normalized }` to cache normalization and provide monotonic line ids.
- Change the debugger output reader to append lines via `append_recent_output_line_locked` so normalization happens once at ingestion.
- Reimplement `capture_framed_debugger_output` to scan incrementally from a `next_scan_id` tail, collecting framed lines without full-buffer rescans, and add `line_contains_full_marker` to avoid marker-prefix collisions.
- Add focused tests covering partial marker arrival, cancellation, timeout, and a microbenchmark `test_framed_capture_marker_scan_microbenchmark` demonstrating the incremental scan is faster than a full rescan.

### Testing

- Ran formatting with `cargo fmt` which succeeded.
- Ran `cargo check --all-targets -p perl-dap` which succeeded.
- Ran `cargo test -p perl-dap` where the test suite mostly passed but one existing end-to-end test failed: `test_e2e_multi_breakpoint_sequence` (asserts a frame line mismatch: expected line 5 but got 4). 
- Ran the focused microbenchmark `cargo test -p perl-dap debug_adapter::parsing::tests::test_framed_capture_marker_scan_microbenchmark` which passed.

------


EffortlessSteven:
### Review: marker-aware framed debugger output capture

**Substantive perf + correctness improvement** to `capture_framed_debugger_output`. Three distinct wins:

1. **O(n) -> O(new lines) per poll**. The previous design called `snapshot_recent_output_lines()` (full clone) and `normalize_debugger_output_line()` (full pass) on every poll iteration. With `RECENT_OUTPUT_MAX_LINES=100` (or whatever the bound is) and a poll interval of `DEBUGGER_FRAME_POLL_MS=10`, that's ~100 string clones + 100 normalizations every 10ms even when nothing new arrived.

   The new design tracks `next_scan_id` and only walks lines with `id >= next_scan_id`, reusing the pre-normalized form stored at insertion time:

   ```rust
   for line in output.lines.iter().filter(|line| line.id >= next_scan_id) {
       if !saw_begin_marker {
           if Self::line_contains_full_marker(&line.normalized, begin_marker) { ... }
       } else if Self::line_contains_full_marker(&line.normalized, end_marker) { ... }
   }
   if let Some(last) = output.lines.back() {
       next_scan_id = last.id.saturating_add(1);
   }
   ```

   This is the right shape: each line is examined at most once per capture call, regardless of how many polls the capture spans. The `next_line_id` is monotonic so `id >= next_scan_id` is a stable filter even after the deque rotates (capacity-bounded ring eviction preserves IDs on remaining entries).

2. **Pre-normalized line storage**:

   ```rust
   struct RecentOutputLine {
       id: u64,
       raw: String,
       normalized: String,
   }
   ```

   `normalize_debugger_output_line` runs once at insertion (in `append_recent_output_line_locked`) instead of per-poll. For a 100-line buffer polled at 100Hz, that's a ~10,000x reduction in normalization work. Worth the ~2x memory cost.

3. **`line_contains_full_marker` is a correctness fix, not just perf**:

   ```rust
   fn line_contains_full_marker(line: &str, marker: &str) -> bool {
       line.match_indices(marker).any(|(idx, _)| {
           let before = line[..idx].chars().next_back();
           let after = line[idx + marker.len()..].chars().next();
           let before_ok = before.is_none_or(|ch| !matches!(ch, 'A'..='Z' | 'a'..='z' | '0'..='9' | '_'));
           let after_ok = after.is_none_or(|ch| !matches!(ch, 'A'..='Z' | 'a'..='z' | '0'..='9' | '_'));
           before_ok && after_ok
       })
   }
   ```

   Previous `line.contains("DAP_BEGIN_100")` would falsely match `DAP_BEGIN_1000`, `DAP_BEGIN_100_extra`, or `prefix_DAP_BEGIN_100`. The word-boundary check (before-after must not be `[A-Za-z0-9_]`) prevents that. Subtle but real -- if a debugger session emits "evaluation: DAP_BEGIN_100..." as content rather than as a frame marker, the previous implementation would mis-attribute it.

**Test coverage is excellent**:

- `test_capture_framed_debugger_output_handles_partial_marker_arrival`: spawns a producer thread that writes the begin/end markers *after* the capture starts polling. Verifies the new design's incremental-scan correctly handles markers arriving over time, not just markers already in the buffer at capture-start.
- `test_capture_framed_debugger_output_respects_cancellation`: sets `cancel_requested` flag before calling capture, verifies it returns `None` and resets the flag. Note: the assertion `!adapter.cancel_requested.load(Ordering::Acquire)` after capture confirms the flag is *consumed* (reset to false), not just observed -- so a subsequent capture call doesn't inherit the stale cancellation. Correct.
- `test_capture_framed_debugger_output_timeout_without_end_marker`: writes only the begin marker, asserts capture times out with `None` and *honors the minimum* `DEBUGGER_QUERY_WAIT_MS` budget (i.e. doesn't return faster than the minimum poll interval, even if the timeout argument is 1ms).
- `test_framed_capture_marker_scan_microbenchmark`: literal benchmark embedded in tests asserting `incremental_elapsed < full_scan_elapsed`. Brittle (microbenchmark in unit tests will be flaky in noisy CI) but pins the perf claim.

**Suggestion (non-blocking)**: the microbenchmark test (`test_framed_capture_marker_scan_microbenchmark`) is the kind of thing that flakes on CI. Two ways to harden:

1. **Move it to `criterion` benches** in `crates/perl-dap/benches/` instead of unit tests. That's the natural home for "this is faster than that" claims.
2. If keeping it as a unit test for visibility, add a tolerance: `assert!(incremental_elapsed * 2 < full_scan_elapsed, ...)` -- the current assertion `incremental < full` would fail on a single-core CI runner with extremely tight timings. A 2x margin captures the intent (incremental is "much faster") without flaking on noise.

**Suggestion (non-blocking)**: the `RecentOutputLine.id: u64` is monotonically incremented via `saturating_add(1)`. After ~18 quintillion lines, `saturating_add` will pin `id` to `u64::MAX` and subsequent IDs will collide. Practically impossible (would require >2 billion lines per second for 6000 years), but a `// `id` is monotonic and saturates at u64::MAX; a long-running session emitting >18Q lines will see all subsequent lines share id=u64::MAX, causing the framed capture to mis-frame.` comment would document the corner case for future maintainers.

**Verdict**: real perf win (likely measurable in p99 latency for evaluate/variables in long-running DAP sessions), real correctness fix (word-boundary marker matching), well-tested. Pre-existing master CI failures are unrelated; #6495 fixes both halves of the master block.

EffortlessSteven:

EffortlessSteven:

PR #6186, opened 2026-04-24
perf(perl-dap): make framed debugger-output capture marker-aware and cheaper (#0000)

EffortlessSteven:
### Motivation

- Framed debugger-output capture rescanned and re-normalized the entire recent-output buffer on every poll, which is expensive for frequent stack/variables/evaluate queries.
- The goal was to make framed capture marker-aware and incremental while preserving the existing begin/end marker protocol and correctness under cancellation, partial arrivals, timeouts, and interleaved output.

### Description

- Introduced `DebuggerOutputLine { line_id, raw, normalized }` and replaced `recent_output: VecDeque<String>` with `recent_output: VecDeque<DebuggerOutputLine>` plus a monotonic `recent_output_line_id` counter to enable cheap incremental scans.
- Added `append_recent_output_line_with_cache` to cache normalization at ingestion time and keep the bounded ring semantics, and wired the live output reader to call this helper.
- Replaced full-buffer rescans in `capture_framed_debugger_output` with a marker-aware incremental scanner that advances from the last seen `line_id`, accumulates framed lines, and returns a completed frame when the end marker arrives.
- Updated unit tests to use the cached ingestion API and added focused tests: marker isolation, partial/interleaved arrival handling, and a microbenchmark-style comparison asserting the new approach returns identical lines and is faster than the legacy full-rescan logic.

### Testing

- Ran `cargo check --all-targets -p perl-dap`, which completed successfully.
- Ran `cargo test -p perl-dap` and observed that the vast majority of unit tests passed and all newly added framed-capture tests passed, but one existing end-to-end test failed: `test_e2e_multi_breakpoint_sequence` (assertion: expected first breakpoint at line 5 but stack frame reported 4), causing the overall test run to fail.
- Ran focused test runs and microbenchmark: `cargo test -p perl-dap test_capture_framed_debugger_output_microbenchmark_marker_aware_faster -- --exact` passed and the marker-aware framing produced identical lines and a measurable timing improvement in the microbenchmark.

------

### DIFF
--- a/crates/perl-dap/src/debug_adapter/mod.rs
+++ b/crates/perl-dap/src/debug_adapter/mod.rs
@@ -112,6 +112,13 @@ const DEBUG_SESSION_TERMINATE_WAIT_MS: u64 = 250;
 const DEBUGGER_QUERY_WAIT_MS: u64 = 75;
 const DEBUGGER_FRAME_POLL_MS: u64 = 10;
 
+#[derive(Clone, Debug)]
+struct DebuggerOutputLine {
+    line_id: u64,
+    raw: String,
+    normalized: String,
+}
+
 fn context_re() -> Option<&'static Regex> {
     CONTEXT_RE
         .get_or_init(|| {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6185 implements same feature. Your work was validated; any novel tests/edge-cases in your diff can be extracted as follow-up PRs if worth.

PR #6189, opened 2026-04-24
refactor(perl-dap): add framed fast path and smarter reuse for stackTrace (#6519)

EffortlessSteven:
### Motivation
- `stackTrace` was reparsing recent debugger output on every request which caused avoidable CPU and parsing churn when the execution state was unchanged.
- Framed `T` output is a more reliable and immediate snapshot and should be the canonical fast path when available.

### Description
- Add a monotonic `stack_trace_epoch` on `DebugAdapter` and per-session `stack_trace_cache_epoch` so parsed frames can be cached and validated against an epoch for freshness.
- Short-circuit `handle_stack_trace` to return cached, already-filtered frames when the session epoch matches, otherwise prefer framed `T` output as the canonical parse path and only fall back to recent-output parsing if framed parsing yields nothing.
- Invalidate and clear cached frames on execution lifecycle transitions (continue/next/stepIn/stepOut/goto), on launch/attach/clear state, and when the output reader observes a stopped/resume transition so cache freshness tracks real state changes.
- Preserve existing user-visibility filtering and attach/placeholder fallback behavior while reducing repeated parse work.

### Testing
- Ran `cargo test -p perl-dap --test stack_trace_provider_tests` and all tests passed.
- Ran `cargo test -p perl-dap --test stack_malformed_debugger_output_tests` and all tests passed.
- Ran `cargo check --all-targets -p perl-dap` and the check completed successfully, and `cargo fmt --all` was executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb399cdfd48333b30da6fcfb53befe)


EffortlessSteven:
### Review: framed fast path and smarter reuse for stackTrace

Real perf optimization. Two coordinated changes:

1. **Stack-trace cache with epoch invalidation**. Each resume operation (`Continue`, `Next`, `StepIn`, `StepOut`, `Goto`) now does:

```rust
session.stack_frames.clear();
session.stack_trace_cache_epoch = None;
session.variables.clear();
self.invalidate_stack_trace_epoch();
```

And `stackTrace` request handlers check `session.stack_trace_cache_epoch == Some(current_epoch)` -- if so, return the cached frames without re-walking. This is the right shape: cache valid until the program runs again, invalidated on every resume operation.

2. **`paginate_stack_frames` + `filter_user_visible_frames`** -- composable helpers presumably used to serve paginated `stackTrace` requests from the cache without re-querying the debugger.

**Correctness checks**:

- The `clear() + None` pattern at every resume site is **repeated** at five places. This is exactly the kind of duplication that drifts -- a future contributor adding a sixth resume kind (e.g. `stepBack` for time-travel debuggers) will probably miss one of the three lines. Strong recommendation: extract a `session.invalidate_for_resume()` helper:

```rust
impl Session {
    fn invalidate_for_resume(&mut self) {
        self.stack_frames.clear();
        self.stack_trace_cache_epoch = None;
        self.variables.clear();
    }
}
```

Then each resume site becomes a single line. Also lets you assert the invariant in tests if needed.

- The `invalidate_stack_trace_epoch()` self-call vs `session.stack_trace_cache_epoch = None` looks redundant. Worth confirming whether the adapter-level epoch and session-level epoch serve different purposes -- if the adapter epoch is the source of truth and session caches just compare against it, then setting `session.stack_trace_cache_epoch = None` is dead. If they're independent, both need to be touched. Either way, a one-line comment explaining the relationship would prevent future code-archaeology confusion.

**Verdict**: real perf win, correct invalidation pattern. The duplicated invalidation block is the only nit -- factoring it into a method preserves the perf gain while removing the drift surface. Pre-existing master CI failures are unrelated; #6495 fixes the clippy_full half.

EffortlessSteven:

EffortlessSteven:

EffortlessSteven:

EffortlessSteven:
Closing — original branch unrebaseable. Patch does not apply cleanly. Tracking issue: https://github.com/EffortlessMetrics/perl-lsp/issues/7511

PR #6193, opened 2026-04-24
perf(perl-dap): add real caching for variables roots, children, and paging (#0000)

EffortlessSteven:
### Motivation

- Variable expansion previously cached only root requests without robust paging which caused repeated root/child requests to re-parse debugger output and wasted latency.
- Need stable `variablesReference` behavior while making repeated variable requests (root and child) cheaper and more accurate.

### Description

- Replace the single `variables` map with explicit per-session `root_variables` and `child_variables` caches stored on `DebugSession` and wired through `DebugAdapter`.
- `handle_variables` now distinguishes scope-root refs vs child refs, serves cached child slices locally using `start/count` slicing, and parses+caches full root sets once up to a bounded cap (`MAX_VARIABLE_CACHE_ENTRIES`).
- Parsing logic now renders children at cache scale and stores child vectors so subsequent paged requests are served from `child_variables` without reparsing.
- Cache invalidation is applied broadly on execution state changes (`continue`, `next`, `stepIn`, `stepOut`, `goto`, `pause`, and attach transitions) to avoid stale expansions, and completions were updated to read names from both root and child caches.
- Added test assertions: non-overlapping pagination windows for large arrays and repeated-root/child latency checks in the performance test to validate cache-fast follow-ups.

### Testing

- Ran `cargo test -p perl-dap --test variables_deep_truncation` and all tests passed.
- Ran `cargo test -p perl-dap --test variables_comprehensive` and all tests passed.
- Ran `cargo test -p perl-dap --test variables_variable_inspection` and all tests passed.
- Ran `cargo check --all-targets -p perl-dap` and the check completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb3a1e25d08333afa3a9f8d2ec281c)

### DIFF
--- a/crates/perl-dap/src/debug_adapter/evaluation.rs
+++ b/crates/perl-dap/src/debug_adapter/evaluation.rs
@@ -408,7 +408,11 @@ impl DebugAdapter {
                 let session_guard = lock_or_recover(&self.session, "debug_adapter.session");
                 if let Some(ref session) = *session_guard {
                     let mut seen = std::collections::HashSet::new();
-                    for vars in session.variables.values() {
+                    for vars in session
+                        .root_variables
+                        .values()
+                        .chain(session.child_variables.values())
+                    {
                         for var in vars {
                             if (stem.is_empty() || var.name.starts_with(stem))
                                 && seen.insert(var.name.clone())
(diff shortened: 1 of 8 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6192 implements same feature. Your work was validated; any novel tests/edge-cases in your diff can be extracted as follow-up PRs if worth.

PR #6197, opened 2026-04-24
fix(perl-dap): harden live-session evaluate correctness and trim fast-path overhead (#0000)

EffortlessSteven:
### Motivation
- Live evaluate requests could misclassify read-only comparisons as assignments and could return ambiguous timeouts by falling back to stale output instead of honoring framed debugger responses.
- Repeated construction of the microcrate `SafeEvaluator` and duplicated parsing work created avoidable overhead in the hot path.
- The admission-control validator must remain strict for side-effecting ops while not blocking common read-only comparison forms.

### Description
- Reuse a single `SafeEvaluator` instance via `OnceLock` to avoid repeated allocations (`shared_safe_evaluator`).
- Prefer deterministic framed-output parsing for live evaluate requests and avoid falling back to general recent output when framed output exists, and add `evaluate_error_from_lines` to surface explicit runtime/debugger error lines instead of returning a timeout.
- Tighten assignment operator detection by introducing a compiled regex that prioritizes multi-char assignment ops and treats `=` in comparison/match contexts as non-assignment via `is_non_assignment_equals`, and add `x=` to the canonical assignment list.
- Update tests to reflect corrected behavior: treat equality/comparison operators as safe in `allowSideEffects: false` and add unit tests for the new error-detection helper.

### Testing
- Ran `cargo test -p perl-dap --test dap_evaluate_comprehensive_tests` and all tests passed.
- Ran `cargo test -p perl-dap --test eval_timeout_and_exception_tests` and all tests passed.
- Ran `cargo test -p perl-dap --test eval_safe_evaluator` and all tests passed.
- Ran `cargo check --all-targets -p perl-dap` and it completed successfully, and `cargo xtask fmt` was executed for formatting; `just pr-fast` was not available in the environment.

------

### DIFF
--- a/crates/perl-dap/src/debug_adapter/evaluation.rs
+++ b/crates/perl-dap/src/debug_adapter/evaluation.rs
@@ -2,6 +2,32 @@
 
 use super::*;
 
+fn shared_safe_evaluator() -> &'static SafeEvaluator {
+    static SAFE_EVALUATOR: OnceLock<SafeEvaluator> = OnceLock::new();
+    SAFE_EVALUATOR.get_or_init(SafeEvaluator::new)
+}
+
+fn evaluate_error_from_lines(lines: &[String]) -> Option<String> {
+    lines.iter().rev().find_map(|line| {
+        let text = DebugAdapter::normalize_debugger_output_line(line);
+        let trimmed = text.trim();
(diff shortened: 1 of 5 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6196 implements same feature. Your work was validated; any novel tests/edge-cases in your diff can be extracted as follow-up PRs if worth.

PR #6199, opened 2026-04-24
fix(perl-dap): harden live-session evaluate correctness and trim fast-path overhead (#0000)

EffortlessSteven:
### Motivation

- Improve real debugger `evaluate` behavior so failures returned by the debugger are classified and surfaced deterministically instead of being reported as timeouts.
- Keep safe-eval admission as a policy check while avoiding repeated work and false positives for read-only comparison operators.
- Reduce unnecessary regex work on common safe-expression fast-paths (e.g. avoid running expensive patterns unless marker substrings are present).

### Description

- Centralized admission checks into `validate_evaluate_expression` and reused a static `SafeEvaluator` instance (`std::sync::LazyLock<SafeEvaluator>`) to avoid repeated microcrate instantiation and duplicate validation logic.
- Prefer parsing per-request framed debugger output (`capture_framed_debugger_output`) deterministically and only fall back to recent output when framing is unavailable, and added `classify_evaluate_error_from_lines` to shape meaningful error/exception messages from debugger output before returning a generic timeout.
- Tightened `validate_safe_expression` fast-paths by short-circuiting expensive regex checks for `&{` and `<*` with cheap `contains` checks before calling the compiled regexes.
- Reworked assignment detection: added a tokenizing `ASSIGNMENT_OPS_RE` to match operator runs and updated the validator to classify only true assignment/compound-assignment tokens (blocking mutations) while allowing read-only comparisons like `==`, `!=`, `<=`, `>=`, and `<=>`.
- Adjusted tests to reflect the new validator behavior (comparisons now pass safe-mode checks) and fixed `setExpression` error message handling.

### Testing

- Ran `cargo test -p perl-dap --test dap_evaluate_comprehensive_tests` and it passed (37 tests, 0 failed).
- Ran `cargo test -p perl-dap --test eval_timeout_and_exception_tests` and it passed (53 tests, 0 failed).
- Ran `cargo test -p perl-dap --test eval_safe_evaluator` and it passed (59 tests, 0 failed).

------

### DIFF
--- a/crates/perl-dap/src/debug_adapter/evaluation.rs
+++ b/crates/perl-dap/src/debug_adapter/evaluation.rs
@@ -3,6 +3,42 @@
 use super::*;
 
 impl DebugAdapter {
+    fn validate_evaluate_expression(expression: &str) -> Option<String> {
+        if let Some(error) = validate_safe_expression(expression) {
+            return Some(error);
+        }
+
+        // Re-run through the shared SafeEvaluator microcrate to keep evaluation
+        // policy aligned with crate-level DAP security logic.
+        static SAFE_EVALUATOR: std::sync::LazyLock<SafeEvaluator> =
+            std::sync::LazyLock::new(SafeEvaluator::new);
+        SAFE_EVALUATOR.validate(expression).err().map(|error| error.to_string())
+    }
+
+    fn classify_evaluate_error_from_lines(lines: &[String]) -> Option<String> {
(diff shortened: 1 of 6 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6196 implements same feature. Your work was validated; any novel tests/edge-cases in your diff can be extracted as follow-up PRs if worth.

PR #6208, opened 2026-04-24
perf(perl-dap): remove measured transport/output overhead in hot DAP paths (#0000)

EffortlessSteven:
### Motivation
- Reduce allocation and flush/lock churn on hot DAP transport paths where benchmarks showed framing/write overhead and per-event flush cost.
- Preserve exact DAP Content-Length framing semantics while lowering per-message overhead.

### Description
- Added helpers `DebugAdapter::write_framed_payload` and `DebugAdapter::write_framed_message` to stream `Content-Length` headers and payload bytes directly to the writer without building an intermediate framed buffer. 
- Changed event thread to batch-send queued events after acquiring the `event_writer` lock (draining `rx.try_recv()`), emitting a single `flush()` per batch to reduce lock/flush churn. 
- Switched response path to use `write_framed_payload` so responses stream header+payload without extra allocation. 
- Added a transport-sensitive Criterion benchmark `benchmark_dap_transport_output` that compares `frame_alloc_and_write` vs `stream_header_and_payload_write`, and removed the now-unused `frame` import from the module.

### Testing
- Ran formatting with `cargo xtask fmt` successfully. 
- Ran transport benchmark with `cargo bench -p perl-dap --bench dap_benchmarks -- dap_transport --measurement-time 3` and observed `stream_header_and_payload_write` ≈ `53.8 ns` vs `frame_alloc_and_write` ≈ `55.7 ns` in the measured run. 
- Ran protocol tests with `cargo test -p perl-dap --test dap_protocol_message_tests` and all tests passed. 
- Ran golden transcript tests with `cargo test -p perl-dap --test dap_golden_transcript_tests` and all tests passed. 
- Ran `cargo check --all-targets -p perl-dap` which completed successfully.

------

### DIFF
--- a/crates/perl-dap/benches/dap_benchmarks.rs
+++ b/crates/perl-dap/benches/dap_benchmarks.rs
@@ -37,9 +37,11 @@ use perl_dap::debug_adapter::DebugAdapter;
 use perl_dap::platform::{
     format_command_args, normalize_path, resolve_perl_path, setup_environment,
 };
+use perl_lsp_rs_core::transport::framing::frame;
 use serde_json::json;
 use std::collections::HashMap;
 use std::hint::black_box;
+use std::io::Write;
 use std::path::PathBuf;
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6211 includes dap_benchmarks.rs for perf validation. Your transport work informed the keeper.

PR #6210, opened 2026-04-24
perf(perl-dap): remove measured transport/output overhead in hot DAP paths (#0000)

EffortlessSteven:
### Motivation

- Benchmarks indicated measurable overhead in DAP transport hot paths from per-event framing allocations, per-message flush churn, and lock contention during bursty event emission.
- Reduce observed CPU/time spent in framing and writer contention while preserving exact DAP `Content-Length` wire semantics.

### Description

- Switched the shared transport writer to a `BufWriter` wrapped in `Arc<Mutex<...>>` to reduce small write syscall overhead and enable batched flushes.
- Replaced allocating `frame(...)` usage with a new `write_framed_payload` helper that writes the `Content-Length` header directly to the writer and then the payload to avoid intermediate allocation.
- Batched pending event messages in the event-handler thread by pulling the first `recv()` and then draining additional messages with `try_recv()` so bursts acquire the lock and flush once per batch instead of once per event.
- Updated imports in `debug_adapter::mod` to add `BufWriter` and removed the now-unused `frame` import, and added a transport-focused Criterion benchmark group (`transport_output`) to `benches/dap_benchmarks.rs` to measure framing and flush strategies.

### Testing

- Ran formatting with `cargo xtask fmt` and applied changes.
- Measured with `cargo bench -p perl-dap --bench dap_benchmarks -- transport_output --measurement-time 3` and observed the direct header write and batched flush variants outperform the legacy allocating and per-message-flush variants (direct header write ~49 ns vs legacy ~80 ns; batched flush ~518 ns vs per-message ~592 ns in this run).
- Verified protocol tests with `cargo test -p perl-dap --test dap_protocol_message_tests` which passed (41 tests passed).
- Verified golden transcript tests with `cargo test -p perl-dap --test dap_golden_transcript_tests` which passed (5 tests passed).
- Performed `cargo check --all-targets -p perl-dap`, which completed successfully (non-fatal test-support warnings unchanged).

------


EffortlessSteven:
Closing as REDUNDANT — #6211 includes dap_benchmarks.rs for perf validation. Your transport work informed the keeper.

PR #6211, opened 2026-04-24
perf(perl-dap): remove measured transport/output overhead in hot DAP paths (#6506)

EffortlessSteven:
### Motivation
- Reduce measured transport and output overhead in hot DAP paths caused by per-message framing allocations and frequent write/flush cycles under bursty event emission.
- Preserve exact DAP Content-Length framing and delivery semantics while lowering lock contention on the shared writer.

### Description
- Batch event channel messages in the event-thread by draining up to `EVENT_WRITE_BATCH_MAX` via `try_recv` before taking the shared writer lock, reducing lock/flush churn.
- Add `write_framed_payload` to write `Content-Length` header and payload directly to the writer without allocating an intermediate framed buffer for each message.
- Emit batched payloads with a single lock acquisition and a single `flush()` per batch, preserving message order and framing semantics.
- Remove now-unused `frame` import from `mod.rs` after switching to direct framed writes.

### Testing
- Ran `cargo test -p perl-dap --test dap_protocol_message_tests` and all tests passed (41 passed). 
- Ran `cargo test -p perl-dap --test dap_golden_transcript_tests` and all tests passed (5 passed).
- Ran `cargo check --all-targets -p perl-dap` and the check completed successfully.
- Ran the `dap_dispatch` benchmark via `cargo bench -p perl-dap --bench dap_benchmarks -- dap_dispatch --measurement-time 1`; the bench completed and reported `dap_threads_request` ≈ 169–173 ns and `dap_stacktrace_request` ≈ 1.25–1.28 µs in this environment.

------

### DIFF
--- a/crates/perl-dap/src/debug_adapter/mod.rs
+++ b/crates/perl-dap/src/debug_adapter/mod.rs
@@ -40,7 +40,7 @@ use crate::tcp_attach::{DapEvent, TcpAttachConfig, TcpAttachSession};
 use crate::types::{Source, StackFrame, Variable};
 use crate::variables::{PerlVariableRenderer, RenderedVariable, VariableParser, VariableRenderer};
 use perl_lexer::DAP_COMPLETION_KEYWORDS;
-use perl_lsp_rs_core::transport::framing::{ContentLengthFramer, frame};
+use perl_lsp_rs_core::transport::framing::ContentLengthFramer;
 use perl_module::path::module_path_to_name;
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
--- a/crates/perl-dap/src/debug_adapter/transport.rs
+++ b/crates/perl-dap/src/debug_adapter/transport.rs
@@ -1,6 +1,16 @@
 //! Transport layer: run (stdin/stdout), run_socket, run_with_io.
 
 use super::*;
+use std::sync::mpsc::TryRecvError;
+
+const EVENT_WRITE_BATCH_MAX: usize = 64;
+
+fn write_framed_payload<W: Write>(writer: &mut W, payload: &[u8]) -> io::Result<()> {
+    writer.write_all(b"Content-Length: ")?;
+    writer.write_all(payload.len().to_string().as_bytes())?;
+    writer.write_all(b"\r\n\r\n")?;
+    writer.write_all(payload)
+}
 
 impl DebugAdapter {
     /// Run the debug adapter server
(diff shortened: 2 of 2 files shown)


EffortlessSteven:

EffortlessSteven:
standards review: clean — DAP transport perf optimization. Replaces single-message recv→serialize→frame→write→flush per event with a batched recv-loop: receives one message blocking, then drains up to EVENT_WRITE_BATCH_MAX=64 additional messages via try_recv before serializing all and writing them under a single writer-lock acquisition + single flush. Disconnect handling is correct (loop terminates after writing whatever batch was assembled). New write_framed_payload helper avoids the intermediate Vec<u8> allocation that frame() imposed. Same wire format. The same helper is reused for the response path. Banned-pattern scan clean.

EffortlessSteven:
## Maintainer-PR Review: QUALITY GAP

The batching and framing optimization is sound architecture, but the measurement claim lacks evidence. The PR body states 'Ran the `dap_dispatch` benchmark...reported...in this environment' but doesn't include before/after numbers, baseline context, or a regression test to prevent framing-order bugs from being silent. The benchmark output section is incomplete.

**Required before sign-off:**
1. Add benchmark baseline (what was the latency before batching?)
2. Add a regression test verifying message ordering is preserved under bursty emission
3. Include the actual measured improvement in the PR body or as a CI artifact

This is a hot path change (event transport); we need proof the trade-off (complexity) delivers.

EffortlessSteven:
**Deep review: logic CORRECT, deferring to maintainer's needs-builder-fix.**

Verified the batching implementation in `transport.rs`:

1. **Message ordering**: Preserved. The MPSC channel is FIFO; batch collection preserves insertion order; `payloads` is built in batch order; writes iterate `payloads` in order. No reordering.

2. **Disconnected handling**: Correct. The `disconnected` flag is checked after writing the final batch, ensuring the batch is flushed before the loop exits.

3. **write_failed semantics**: If a write fails mid-batch, remaining payloads in that batch are silently dropped. This matches the old behavior (broken writer = session over), though it is now more total — a broken pipe drops the rest of the batch rather than continuing to try. Acceptable for DAP.

4. **let-chains syntax**: `if !write_failed && let Err(e) = writer.flush()` is valid — MSRV is 1.92, let-chains were stabilized in 1.88.

5. **Lock sharing**: Both `event_writer` and `shared_writer` are `Arc::clone`s of the same `Mutex<W>`, so event and response writes serialize correctly.

**Decision**: The logic is correct. Not applying `deep-reviewed` — respecting the active `needs-builder-fix` from the maintainer asking for benchmark evidence and an ordering regression test. Those artifacts are warranted for a hot-path change.

PR #6213, opened 2026-04-24
fix(perl-dap): harden bridge adapter lifecycle and shutdown behavior (#0000)

EffortlessSteven:
### Motivation
- Bridge proxying could hang or behave nondeterministically when one side closed (open stdin reads could block the client->server copier) and shutdown escalation was imprecise.  

### Description
- Reworked `proxy_messages` to run the two copy loops as `tokio::spawn` tasks and coordinate completion with `tokio::select!`, aborting the client-input task deterministically when server output completes.  
- Improved `shutdown` to attempt an explicit child stdin shutdown, prefer `SIGTERM` on Unix, and escalate to `start_kill` with a bounded force-wait timeout via new `PLS_FORCE_KILL_WAIT_MS`.  
- Tightened `wait_for_child_exit` polling to use the remaining timeout window for sleep intervals to avoid oversleeping.  
- Added `test_respawn_replaces_previous_child_cleanly` unit test and a lightweight `bridge_lifecycle` benchmark in `dap_benchmarks` to track startup/shutdown overhead over time.  

### Testing
- Ran `cargo test -p perl-dap --test bridge_adapter_unit_tests`, which passed.  
- Ran `cargo test -p perl-dap --test bridge_integration_tests`, which passed.  
- Ran `cargo check --all-targets -p perl-dap`, which passed.  
- Ran `cargo xtask fmt` successfully; `just pr-fast` was not available in this environment so it was not executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb3a60fee083339a15250dd0294d86)

### DIFF
--- a/crates/perl-dap/benches/dap_benchmarks.rs
+++ b/crates/perl-dap/benches/dap_benchmarks.rs
@@ -32,6 +32,7 @@
 //! ```
 
 use criterion::{Criterion, criterion_group, criterion_main};
+use perl_dap::BridgeAdapter;
 use perl_dap::configuration::LaunchConfiguration;
 use perl_dap::debug_adapter::DebugAdapter;
 use perl_dap::platform::{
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6215 is the more complete implementation with unit tests + benchmark. Your bridge_adapter work was validated and incorporated.

PR #6217, opened 2026-04-24
fix(diagnostics): linearize pod export fallback scan (#6503)

EffortlessSteven:
### Motivation

- The PL304 POD-coverage fallback parser in `collect_exported_names_from_source` recomputed a matching line's start offset by rescanning prior lines for every `@EXPORT` match, creating an O(n^2) workload and enabling CPU exhaustion on crafted files.

### Description

- Replace the repeated rescan with a single-pass running offset (`line_start`) while iterating `source.lines()` so the fallback scan is linear-time and preserves emitted spans; change applied in `crates/perl-lsp-rs-core/src/providers/diagnostics/lints/pod_coverage.rs`.

### Testing

- Ran `cargo test -p perl-lsp-rs-core --test pod_coverage_tests` and all tests passed (13 passed, 0 failed).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb2dd94a0c833399475bfc61cf9c4f)


EffortlessSteven:
### Review: linearize pod export fallback scan

**Correctness**: Verified. The replacement keeps the exact same approximation as the prior code (`line.len() + 1` per line, irrespective of CRLF vs LF), so spans and emitted name positions are bit-for-bit identical for any input the original handled. The transformation is a clean O(n^2) -> O(n) reduction:

- Old: for each `@EXPORT` match, recomputes `line_start = source.lines().take(i).map(|l| l.len() + 1).sum()` (full prefix scan per match).
- New: a single running `line_start` advanced by `line.len() + 1` on every iteration, with an early `continue` branch that still bumps the offset before skipping.

Both `continue` and the post-match path advance `line_start`, so the offset stays correct across mixed-content files. No behavioral change visible to callers; the existing `pod_coverage_tests` suite (13 tests) is the regression net.

**Caveat for reviewers**: the `line.len() + 1` approximation is preserved, not fixed. If a future PR wants to be exact about `\r\n` and trailing-no-newline, a separate fix should switch to `source.split_inclusive('\n')` or `memchr`-driven offsets. Out of scope here -- this PR explicitly preserves behavior.

**CI status**: The two failing gates (`clippy_full`, `lsp_tier_b`) are pre-existing master failures, not caused by this PR:

- `clippy_full` is failing on unused imports / dead code in `crates/perl-lsp-rs/tests/multi_root_workspace_tests.rs` (lines 22, 24, 27, 30, 36, 42, 58, 68) -- this file is untouched here.
- `lsp_tier_b` failed on `test_valid_file_execution` in `execute_command_security_tests`, which has no overlap with the diagnostics provider being modified.

Recent master CI runs (commits `88eb1791e`, `abf9f07045`) show the same failure pattern, confirming this is an environmental/master issue and not a regression introduced by this PR.

**Verdict**: Logic is correct and minimal. Once master CI is unblocked, this should sail through.

EffortlessSteven:
## Deep review: linearize POD export fallback scan

**Attack vector**: Algorithmic-complexity DoS in `collect_exported_names_from_source` — the original computed `source.lines().take(i).map(|l| l.len() + 1).sum()` per matching line, giving O(n^2) on files with many `@EXPORT` matches. A crafted file could pin CPU on every diagnostics pass.

**Fix verdict**: SOLID. Replaces the per-match rescan with a single running `line_start` updated by `line.len() + 1` on every iteration, including the early-continue path (verified: continue branch at line 159 increments before `continue`, post-match path increments at end of body). Now O(n).

**Behavior preservation**: The `+1` accounting matches the original (still wrong for `\r\n` files in the same way it was wrong before — not a regression). Offsets emitted to diagnostics are byte-identical to the prior implementation for `\n`-terminated source.

**Sibling vector check**: Grepped `pod_coverage.rs` for similar `source.lines().take(i)` patterns — none. The other line-walking functions (`collect_documented_names`) iterate without offset reconstruction.

EffortlessSteven:
standards review: clean — fixes O(n²) line-start computation in collect_exported_names_from_source. Previously each iteration called `source.lines().take(i).map(|l| l.len() + 1).sum()` to compute line_start, recomputing all prior line lengths from scratch. Replaced with a running line_start accumulator that adds line.len()+1 per iteration (correctly handled in both the early-continue and main branches). Pure performance fix — same byte offsets returned. Banned-pattern scan clean.

EffortlessSteven:

The O(n^2) → O(n) fix to pod_coverage's fallback scan is a clean correctness and performance fix. The diff correctly replaces the repeated rescan with a single-pass running offset (`line_start`). Tests pass. This is the kind of focused perf fix that fits perl-lsp: targeted, measurable improvement to an existing diagnostics path.

PR #6218, opened 2026-04-24
fix(semantic): bound inherited MRO traversal depth (#6502)

EffortlessSteven:
### Motivation
- Prevent unbounded recursive MRO traversal that can be triggered via SUPER:: resolution and cause stack overflow / availability DoS when processing maliciously deep inheritance chains.

### Description
- Add a shared traversal cap `MAX_MRO_TRAVERSAL_DEPTH` to limit ancestor traversal depth.
- Enforce the depth cap in `dfs_ancestor_order` by threading a `depth` parameter through the recursive `walk` helper.
- Enforce the depth cap in `c3_ancestor_order` by threading a `depth` parameter through the recursive `linearize` helper.
- Add two regression tests that construct inheritance chains deeper than the cap and assert resolution returns safely for both DFS and C3 MRO modes.

### Testing
- Ran `cargo test -p perl-semantic-analyzer` and all tests passed (`123` + many integration/unit tests reported as OK). 
- Ran `cargo check --all-targets -p perl-semantic-analyzer` and it completed successfully.
- Ran `cargo clippy -p perl-semantic-analyzer -- -D warnings` and it completed successfully.
- Ran `cargo xtask fmt`; formatting ran in this environment, and `just pr-fast` was not available here (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb2dd92bc083339cde6d9ef0bf759a)


EffortlessSteven:
### Review: bound inherited MRO traversal depth

**Correctness**: Both traversals are guarded consistently and the saturating behavior is sane:

- `dfs_ancestor_order::walk`: depth check returns early before recursing, so parents past the limit simply don't appear in the ancestor list. Existing `seen.insert(...)` cycle guard is unchanged.
- `c3_ancestor_order::linearize`: at the depth limit, returns `vec![package.to_string()]`, which is the same shape as the "package with no parents" branch a few lines below. The C3 merge step then just sees a leaf, so the result remains a valid linearization (truncated at the bound), and the "skip(1)" tail in the caller still drops the originating package consistently.

The cycle guard (`!visited.insert`) and the depth guard are independent, which is the right layering -- depth bounds the *path length*, cycle bounds the *visited set*. A pathological deep-but-acyclic chain (this PR's threat model) is now bounded; cycles were already bounded.

**Test coverage**: Two tests, one for each traversal flavor (`P0 -> P1 -> ... -> P(MAX+10)`), assert the inherited method becomes unresolvable past the bound. That's the right shape -- it pins the *boundary* behavior rather than asserting any specific intermediate result. Note that constructing 1034 packages per test will add a small but non-zero cost to the suite; if that becomes a flaky-time issue we can lower the constant later, but 1024 is a fine default.

1. A short doc comment on `MAX_MRO_TRAVERSAL_DEPTH` would help the next reader -- something like "Cap MRO DFS / C3 recursion to prevent stack exhaustion or pathological linear inheritance chains causing CPU exhaustion." Not required.
2. If we want to be extra defensive, we could also surface a `tracing::debug!` when the bound trips, so a real-world hit shows up in logs instead of silently truncating. Optional; the test already pins the behavior.

**CI status**: As with the sibling pod-coverage PR (#6217), the failing `clippy_full` and `lsp_tier_b` gates here are pre-existing master failures (clippy errors in `multi_root_workspace_tests.rs`, and `test_valid_file_execution` in `execute_command_security_tests`). Neither file is touched by this PR. Recent master-branch CI runs (`88eb1791e`, `abf9f07045`) show the same failures, confirming this is environmental and not a regression from this change.

**Verdict**: Approach and tests are correct and minimal. Once the master CI is unblocked, this is merge-ready as-is.

EffortlessSteven:
## Deep review: bound MRO traversal depth

**Attack vector**: Stack-overflow / availability DoS via maliciously deep `@ISA` / `use parent` inheritance chains processed by `SemanticAnalyzer::resolve_inherited_method_location`. Both `dfs_ancestor_order` and `c3_ancestor_order` use unbounded recursion that could exhaust the OS thread stack on attacker-controlled source.

**Fix verdict**: SOLID. Adds `MAX_MRO_TRAVERSAL_DEPTH = 1024` and threads a `depth` counter through both recursive helpers (`walk` for DFS at line 495, `linearize` for C3 at line 524). Both bail out early when the cap is reached. Tests construct chains of `MAX_MRO_TRAVERSAL_DEPTH + 10` packages and assert resolution returns safely for both DFS and C3 modes.

**Sibling vector check**: Grepped `crates/perl-semantic-analyzer/src` for other recursive parent walkers — `class_model.rs` uses iterative collection only, and the only recursive ancestor walks are the two patched here. No siblings.

EffortlessSteven:
standards review: clean — bounds the recursive walk() and linearize() helpers in resolve_inherited_method_location with MAX_MRO_TRAVERSAL_DEPTH=1024 to prevent unbounded DFS on adversarial @ISA chains. Adds two regression tests (test_resolve_inherited_method_location_limits_dfs_depth and test_resolve_inherited_method_location_limits_c3_depth) that build a 1034-package chain and assert the lookup terminates returning None. The depth threadthrough on the inner walk() and linearize() is mechanically correct — depth+1 on every recursive call, base case checked first. Banned-pattern scan clean.

EffortlessSteven:
## Maintainer-PR Review

**ALIGNED**

Adds MAX_MRO_TRAVERSAL_DEPTH guard to prevent stack overflow on maliciously deep inheritance chains. The constant (1024) is reasonable and applies symmetrically to both DFS and C3 MRO algorithms. Regression tests construct inheritance chains deeper than the cap and assert safe returns. Code follows crate conventions. Scope matches issue spec exactly — a security hardening fix, not feature creep. Quality bar met.

EffortlessSteven:
## deep-review verdict: CORRECT

**#6218 fix(semantic): bound MRO traversal depth at 1024**

Logic verified:

- Both DFS (`walk`) and C3 (`linearize`) paths are depth-bounded with the same constant `MAX_MRO_TRAVERSAL_DEPTH = 1024`.
- DFS: `depth >= 1024` causes early return — correct guard before the recursive `walk` call. The `seen` HashSet already prevents cycles, but cycles with depth > 1024 nodes could stack-overflow before the cycle is detected — the depth cap provides the DoS guardrail independently.
- C3: `depth >= 1024` returns `vec![package.to_string()]` treating the depth-limited node as a leaf with no parents — graceful truncation, no panic.
- Test correctness: chain of 1034 (= 1024 + 10) packages, target at depth 1034. DFS walk stops at depth 1024, never reaches P1034 → `None` ✓. C3 stops similarly → `None` ✓.
- C3 test correctly sets `use mro 'c3'` only on P0 — `receiver_model.mro` dispatch selects `c3_ancestor_order` for P0 queries.
- Fallback is `Ok(None)` not panic — correct for DoS hardening.

Approved.

PR #6223, opened 2026-04-24
fix(refactoring): scope inline-all edits to definition tree (#0000)

EffortlessSteven:
### Motivation
- Prevent workspace-wide inline (`all_occurrences=true`) from applying edits to indexed files outside the defining file's directory tree, which allowed out-of-scope file writes.

### Description
- Filter `refactor_result.file_edits` returned by `inline_variable_all` to only apply edits whose canonicalized path is inside the definition file's canonical directory using a new `path_within_directory` helper.
- Emit and merge a warning when out-of-scope edits are skipped so callers are informed of skipped targets.
- Add a regression test `test_inline_all_occurrences_skips_out_of_scope_file_edits` to verify out-of-scope indexed files are not modified.
- All changes are localized to `crates/perl-refactoring/src/refactor/refactoring.rs`.

### Testing
- Ran `cargo test -p perl-refactoring inline_all_occurrences` which executed the inline-related tests and they passed.
- Ran unit tests that include `test_inline_all_occurrences_routes_to_workspace_lookup` and `test_inline_all_occurrences_skips_out_of_scope_file_edits`, and both tests passed.
- Ran `cargo check --all-targets -p perl-refactoring` and `cargo xtask fmt`, and both succeeded.
- Note: an initial `cargo test` invocation with multiple test-name arguments failed due to incorrect CLI usage and not due to code failure.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb43c28c5883339ae76d0b7ec63218)


EffortlessSteven:
## Deep Review — PR #6223

**Verdict: approved with fix pushed**

### What I verified

**`canonicalize()` correctness (Windows + POSIX)**
`Path::starts_with` does component-wise comparison, not string-prefix matching.  On Windows `canonicalize()` returns `\?\C:\path` style UNC paths; component-wise `starts_with` handles this correctly.  No issue.

**Fails-closed on canonicalize error**
The `_ => false` arm in `path_within_directory` is intentional and safe: if a path in the workspace index cannot be resolved (deleted file, dangling symlink), treating it as out-of-scope is correct.  Silently including an unresolvable path would be worse.  One mild concern: a truly-symlinked file whose target resolves correctly will also resolve correctly, so valid symlinked workspaces are unaffected.  Only dangling symlinks are blocked, which is the right behaviour.

**Warning count accuracy across multiple refactor invocations**
`skipped_out_of_scope` is a local `usize` initialized to `0` on each call, so counts are per-invocation only.  Accurate.

**Target dir is a symlink**
`def_file.parent()` returns the parent of the *symlink path*, then `canonicalize` resolves it.  `file_edit.file_path` is also canonicalized.  Both land on the same canonical basis.  No issue.

**Single-file inline (`all_occurrences: false`)**
Single-file path is entirely separate (`else` branch of `if all_occurrences`).  The scope filter is not applied.  No regression.

### Bug found and fixed

**Missing branch in empty-scoped-edits early return.**

When the workspace index finds edits but *all* of them are filtered as out-of-scope, the original code returned `success: false` with "Symbol not found across workspace" and discarded both `skipped_out_of_scope` (the skip count) and `refactor_result.warnings` (from `inline_variable_all`).  Callers received a misleading "not found" message with no hint that files were filtered.

Fix pushed in commit `9c78d6fc3`:
- Emit "Skipped N out-of-scope file(s) during inline all_occurrences; no in-scope edits remain" instead of "not found" when `skipped_out_of_scope > 0`.
- Merge `refactor_result.warnings` in this early-return path (same as the success path).
- Add `test_inline_all_occurrences_counts_only_in_scope_files_modified`: verifies out-of-scope files are untouched and the scope-skip warning is present when in-scope edits succeed.

All 4 `inline_all_occurrences` tests pass; clippy clean.

PR #6018, opened 2026-04-24
fix(incremental): harden batch edit normalization and UTF-8 boundary fallback (#6018)

EffortlessSteven:
### Motivation
- Batch incremental edits previously assumed ranges were valid and non-overlapping which allowed mid-UTF-8-codepoint and out-of-bounds edits to slip through and produced stale/inconsistent reuse results.
- Make batch edit application safe and deterministic at byte, UTF-8, and range boundaries while preserving current reverse-order application semantics and zero-width insertions.

### Description
- Added `normalize_for_source` to `IncrementalEditSet` to validate/marshal edits for deterministic reverse-order application; it rejects backwards ranges, out-of-bounds ranges, overlapping non-empty ranges, and edits that are not on UTF-8 char boundaries, while preserving zero-width insertions (`crates/perl-parser/src/incremental/incremental_edit.rs`).
- Hardened `IncrementalDocument::apply_edits` to use normalized edits, return an early conservative fallback when normalization or in-place mapping fails, and preserve reverse-order application; added `fallback_parse_batch` to perform the safe full-parse fallback (`crates/perl-parser/src/incremental/incremental_document.rs`).
- Centralized per-edit validation via `map_edit_range`, made `apply_edit_in_place` return a boolean mapped success indicator, and kept safe UTF-8/byte bound checks for both in-place and string-based edit application (`crates/perl-parser/src/incremental/incremental_document.rs`).
- When reuse is attempted for a batch, validate the reused parse against a fresh parser result and conservatively use the fresh parse if the trees diverge (`crates/perl-parser/src/incremental/incremental_document.rs`).
- Added focused regression tests covering overlapping edits, backwards ranges, mid-codepoint edits, one-bad-edit batch fallback, and parity between supported incremental batches and a fresh parse (`crates/perl-parser/src/incremental/incremental_boundary_regressions.rs`) and registered the module for test builds (`crates/perl-parser/src/incremental/mod.rs`).

### Testing
- Ran code formatting: `cargo xtask fmt` (succeeded).
- Type/check: `cargo check --all-targets -p perl-parser` (succeeded).
- Targeted incremental tests: `cargo test -p perl-parser --features incremental --lib incremental::incremental_boundary_regressions::` (all new regression tests passed).
[...]

PR #5969, opened 2026-04-24
refactor(incremental): make perl-parser the single source of truth for incremental parsing (#0000)

EffortlessSteven:
### Motivation

- The incremental parsing implementation was duplicated in `perl-parser` and `perl-incremental-parsing`, creating split-brain risk for correctness and performance fixes. 
- Make ownership explicit and reduce future drift by selecting one canonical home for incremental logic. 
- Provide a low-churn, backwards-compatible migration path so existing imports continue to work.

### Description

- Chose `perl-parser` as the owner of incremental parsing and removed the duplicated implementation from `crates/perl-incremental-parsing/src/incremental/*`. 
- Replaced `crates/perl-incremental-parsing/src/lib.rs` with a thin compatibility shim that re-exports `perl_parser` incremental APIs and common top-level symbols (e.g. `incremental`, `edit`, `Node`, `Parser`). 
- Updated `crates/perl-incremental-parsing/Cargo.toml` to depend on `perl-parser` with the `incremental` feature and updated `README.md` to document the ownership and migration guidance. 
- Updated `Cargo.lock` to reflect the new dependency arrangement and removed the now-deleted implementation files.

### Testing

- Ran `cargo check --workspace --all-targets` which completed successfully. 
- Ran `cargo test -p perl-parser --features incremental incremental_checkpoint` and the targeted incremental checkpoint tests passed. 
- Ran `cargo test -p perl-incremental-parsing --lib` which completed successfully (crate is now a re-export shim and has no unit tests exercising removed code). 
- Ran `cargo fmt --all -- --check` which failed due to pre-existing formatting drift in unrelated files; this change did not introduce the formatting differences.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffbd546083338ba45678dc551b3e)

PR #6224, opened 2026-04-24
fix(incremental): bound subtree cache cloning to avoid DoS (#0000)

EffortlessSteven:
### Motivation
- Incremental subtree caching previously cloned every AST node into cache maps, which deep-copied whole subtrees and could produce O(n^2) memory/time growth on large or adversarial documents when `IncrementalDocument::new` or cache rebuilds ran.
- The intent is to preserve useful subtree reuse for incremental parsing while preventing uncontrolled deep cloning that can crash the LSP process.

### Description
- Restrict cache population in `cache_node` so entries are created only for top-level nodes (`depth <= 1`), leaf nodes, or nodes classified as `SymbolPriority::Critical` instead of caching every node.
- Add a `depth` parameter to `cache_node` and update recursive calls to track depth, and add an `is_leaf_node` helper to detect leaves without cloning.
- Ensure a single `Arc<Node>` is created per cached entry and reused when inserting into both `by_range` and `by_content` to avoid duplicate deep clones.
- Changes are localized to `crates/perl-incremental-parsing/src/incremental/incremental_document.rs` and preserve the existing incremental traversal/recursion logic.

### Testing
- Ran `cargo test -p perl-incremental-parsing` and all unit tests passed (`47 passed; 0 failed`).
- Ran `cargo check --all-targets -p perl-incremental-parsing` and it completed successfully.
- Ran `cargo clippy -p perl-incremental-parsing` (without `-D warnings`) and it completed; the stricter `-D warnings` invocation fails on pre-existing bench lints unrelated to this change.
- Ran formatting via `cargo xtask fmt` which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb26781e288333b37b1f2db8d5e146)

### DIFF
--- a/crates/perl-incremental-parsing/src/incremental/incremental_document.rs
+++ b/crates/perl-incremental-parsing/src/incremental/incremental_document.rs
@@ -1394,6 +1394,70 @@ mod tests {
     }
 
     #[test]
+    fn test_depth_bounded_cache_does_not_grow_quadratically() -> ParseResult<()> {
+        // Verify that caching a deeply-nested program does not cache every
+        // intermediate node. With unbounded caching a k-deep tree produces O(k)
+        // cache entries per subtree root, leading to O(k^2) total entries.
+        // With depth <= 1 bounding the intermediate nodes are skipped, so the
+        // entry count grows roughly linearly with the number of top-level statements.
+        let deeply_nested = r#"
+            package Outer;
+
+            sub level_one {
+                if (1) {
+                    while (1) {
+                        for my $i (1..10) {
+                            foreach my $j (@items) {
+                                my $deep = $i + $j;
+                            }
+                        }
+                    }
+                }
+            }
+
+            sub level_two {
+                if (1) {
+                    while (1) {
+                        for my $x (1..5) {
+                            my $also_deep = $x * 2;
+                        }
+                    }
+                }
+            }
+        "#;
+
+        let doc = IncrementalDocument::new(deeply_nested.to_string())?;
+
+        // Without depth bounding, every node in the two deeply-nested subs
+        // would be cloned into the cache.  With depth <= 1 bounding, only the
+        // top-level package + two subs (depth 0/1) and leaves/criticals survive.
+        // The exact count is implementation-dependent, but it must be well below
+        // the O(total-nodes^2) worst case.  A generous upper bound of 50 entries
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review complete — approved with invariant test added.**

Reviewed `cache_node` depth-bounding fix in `perl-incremental-parsing`.

**Logic verified:**
- `depth <= 1` catches root (Program, depth=0) and top-level statements (depth=1, where `sub`/`package`/`use` live). Intermediate nodes at depth >= 2 are only cached if they are `Critical` priority or leaf nodes. This directly prevents the O(k) clones per subtree that caused DoS.
- `is_leaf_node` and the `match &node.kind` recursion arms are internally consistent: every arm listed as `false` in `is_leaf_node` also has a recursion arm; types handled by `_ => {}` in recursion are correctly returned `true` by `is_leaf_node`.
- The shared `Arc::new(node.clone())` used for both `by_range` and `by_content` is a strict improvement — the old code created two separate deep clones per cached node.
- `hash_node` collision issue (all `Subroutine` nodes hash identically) is pre-existing and not introduced by this PR; `by_range` lookup is unaffected since it keys on `(start, end)`.
- No interaction with PR #6018 (`perl-parser` crate); these are separate crates with separate `IncrementalDocument` implementations.

**Fixed forward:** Added `test_depth_bounded_cache_does_not_grow_quadratically` — verifies that a 5-level-deep program yields ≤ 50 `by_range` entries (ruling out unbounded growth) while preserving ≥ 2 critical symbol entries (the two `sub` definitions). 48/48 tests pass.

EffortlessSteven:
**Closing: superseded by refactor #5969.**

The file this PR patches (`crates/perl-incremental-parsing/src/incremental/incremental_document.rs`) was deleted when #5969 moved incremental parsing logic into `crates/perl-parser/`. The depth-bounded cache fix can't be applied to the original location.

**Follow-up recommendation:** If the subtree-cache DoS is still a concern in the new `perl-parser` incremental module, file a fresh issue pointing to the relevant file there. The fix pattern from this PR (depth-bounded cache with single Arc clone) remains sound; it just needs re-targeting to the new location.

The work + tests were sound — the code moved underneath us. Thank you.

EffortlessSteven:

PR #6226, opened 2026-04-24
fix(perl-parser): avoid quadratic heredoc location scans (#0000)

EffortlessSteven:
### Motivation

- The heredoc anti-pattern detector computed line/column positions by rescanning the source from the start for every regex match, producing O(n^2) behavior on files with many matches and enabling local CPU exhaustion.
- The change replaces repeated prefix scans with a single precomputed line-starts index so location mapping is efficient and deterministic for untrusted workspace files.

### Description

- Added `build_line_starts(code: &str) -> Vec<usize>` and a new `location_from_start(line_starts: &[usize], offset: usize, start: usize) -> Location` helper for constant-time line/column mapping.
- Updated the `PatternDetector::detect` signature to accept `line_starts: &[usize]` and adjusted all detectors to use the precomputed index for location computation.
- Precompute `line_starts` once in `AntiPatternDetector::detect_all` and pass it to each detector so per-match work no longer rescans the source file.
- The only file changed is `crates/perl-parser/src/heredoc_anti_patterns.rs`, keeping the change focused to the parser anti-pattern detector.

### Testing

- Ran `cargo test -p perl-parser heredoc_anti_patterns::tests::` and the heredoc anti-pattern unit tests passed.
- Ran `cargo check --all-targets -p perl-parser` and it completed successfully.
- Ran `cargo clippy -p perl-parser -- -D warnings` and no warnings were reported.
- Ran `cargo test -p perl-parser` which executed the crate test suite but reported one unrelated existing test failure (`refactor::refactoring::tests::validation_tests::test_cleanup_respects_retention_count`), not caused by these changes.
- Ran `cargo xtask fmt`; formatting completed, and `just pr-fast` was not available in the environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb267835b48333a31ccfa351389fe0)


EffortlessSteven:
**Deep review complete — approved with edge-case tests added.**

Reviewed `build_line_starts` + `location_from_start` replacement for the old O(n^2) `code[..start].lines().count()` scan.

**Logic verified:**
- `partition_point(|&s| s <= start)` correctly finds the zero-indexed line for all positions including byte 0, positions exactly on `\n`, and positions on line 2+.
- The new implementation is strictly more correct than the old one: `code[..n].lines().count()` over-counts by 1 when `code[..n]` ends with a non-trailing newline (e.g., a match position of 3 in `"foo\nbar"` returned line=1 on the old path, line=0 on the new path — new is correct).
- `build_line_starts` is never empty (always pushes 0 first), so `get(line).unwrap_or(0)` is a no-op fallback that can never fire.
- `\r\n` line endings: both old and new treat `\r` as column width, not a line break, which is consistent.
- The `offset` parameter in `detect_all` hardcodes 0 for all callers; this pre-existing design limitation is not made worse by this PR.

**Fixed forward:** Added 3 missing edge-case tests — first-byte match (line=0, column=0), third-line match (line=2, column=0, offset=22), and mid-line non-zero column (column=4). All 16 tests pass.

PR #6227, opened 2026-04-24
fix(code-actions): avoid quadratic line offset scans (#0000)

EffortlessSteven:
### Motivation

- The modernization code actions computed byte offsets by calling a function that scanned all lines from the start on every invocation, producing O(n^2) CPU cost for files with many short lines and enabling a local/IPC-triggerable CPU exhaustion (DoS).

### Description

- Precompute line start offsets once per source with a new `line_start_offsets` helper and reuse the table instead of rescanning on each lookup.
- Updated `find_two_arg_open`, `find_legacy_require_version`, and `find_die_in_module` to index into the precomputed `line_offsets` when constructing `TextEdit` ranges.
- Replaced the old per-call `line_start_offset` scanner with the linear `line_start_offsets` implementation in `crates/perl-lsp-rs-core/src/providers/code_actions/modernize.rs`, keeping behavior unchanged.

### Testing

- Ran `cargo test -p perl-lsp-rs-core modernize`, which passed (all modernize tests OK). 
- Ran `cargo check --all-targets -p perl-lsp-rs-core`, which completed successfully. 
- Ran `cargo fmt --all -- --check`, which passed. 
- Ran `cargo clippy -p perl-lsp-rs-core --all-targets -- -D warnings`, which failed due to a pre-existing unrelated `expect()` use in `crates/perl-lsp-rs-core/tests/g3_feature_flag_routing.rs` and is not caused by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb43c9b2288333b1637c7532031fec)


EffortlessSteven:
**Deep review: APPROVED** — `deep-reviewed` label set, 9 edge case tests pushed (commit `ff88ff2d0`).

**What I verified:**

1. **CRLF correctness (subtle improvement beyond the stated goal):** The new `line_start_offsets` scans raw bytes and counts only `\n`, so offset for line N+1 = (byte index of `\n`) + 1. For CRLF input (`"foo\r\nbar"`), this yields [0, 5, ...] where 5 is the 'b'. The old O(n²) `line_start_offset` used `line.len() + 1` where `line` was already stripped of `\r\n` by `str::lines()`, yielding [0, 4, ...] — off by 1 byte per prior CRLF line. This PR silently fixes that corruption in addition to the stated DoS fix.

2. **Off-by-one: none.** `offsets[line_idx]` is the byte position of the first character of that line. `line_end = line_start + line.len()` (where `line` from `str::lines()` excludes the terminator) gives the byte position of the terminator — the correct half-open range for a TextEdit that replaces line content without touching the newline.

3. **LSP positions: correct.** `SourceLocation { start, end }` are byte offsets. `line_start_offsets` produces raw byte offsets. `line.len()` in Rust is byte length, not char length. The arithmetic is byte-level throughout, correct for the upstream LSP→UTF-16 conversion path.

4. **Index bounds: provably safe.** `len(offsets) = 1 + count(\n)` and `len(lines()) ≤ count(\n) + 1`. So `offsets.len() ≥ lines().len()` always. No index-out-of-bounds panic is possible, including for empty file and only-newlines inputs.

5. **Security:** No shell operations, no path operations. DoS vector (O(n²) via LSP IPC) removed.

**What I pushed (commit `ff88ff2d0`):** 9 new tests covering LF, CRLF, no-trailing-newline, empty, only-newlines, and two end-to-end CRLF byte-offset regression tests for `find_two_arg_open` and `find_legacy_require_version` that would have caught the old CRLF bug.

**Out-of-scope follow-up:** `find_pragma_insert_pos` uses the same old `line.len() + 1` pattern (pre-existing on master, not introduced here), so pragma insertion is still CRLF-incorrect. Worth a targeted follow-up fix.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-6227",
  "current_version": {
    "sha": "ff88ff2d0c17afeef3a3bd3fe8946492e085fe00",
    "updated_at": "2026-04-24T12:41:01Z"
  },
  "label_bindings": [
    {
      "label": "deep-reviewed",
      "bound_at_version": "ff88ff2d0c17afeef3a3bd3fe8946492e085fe00",
      "bound_at_timestamp": "2026-04-24T12:41:34Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Review Wrapup

**Summary:** reviewer-deep completed deep correctness review of PR #6227. Outcome: 9 edge case tests added, `deep-reviewed` label set, label receipt written.

**PR improved:** https://github.com/EffortlessMetrics/perl-lsp/pull/6227
**Changes made:** 1 commit pushed (`ff88ff2d0`) — 9 new tests in `crates/perl-lsp-rs-core/src/providers/code_actions/modernize.rs`
**Test artifacts:** `modernize.rs` (same file as the fix — tests live inline in the module)
**Status:** approved
**Next step:** ready for green-ci and diff-auditor passes, then ops merge

---

**Ending state:**

All 38 modernize tests pass (29 pre-existing + 9 new). Clippy clean. No banned patterns introduced.

Tests added:
- `test_line_start_offsets_lf` — exact byte positions verified for LF input
- `test_line_start_offsets_crlf` — byte positions for CRLF; catches the pre-existing off-by-1 the new code silently fixed
- `test_line_start_offsets_no_trailing_newline` — index count matches line count, all in-bounds
- `test_line_start_offsets_empty_file` — no panic, one sentinel offset
- `test_line_start_offsets_only_newlines` — all `line_idx` values within `offsets` bounds
- `test_two_arg_open_crlf_byte_offsets` — end-to-end: CRLF file, edit range starts at byte 13 (not 12)
- `test_require_version_crlf_byte_offsets` — same regression check for `find_legacy_require_version`
- `test_get_modernize_actions_empty_source` — no panic on empty input
- `test_get_modernize_actions_only_newlines` — no panic on newlines-only input

---

**Retrospective:**

The PR's stated fix (O(n²) → O(n)) is correct and complete. The more interesting discovery was that the new byte-scanning approach also silently fixes a pre-existing CRLF off-by-1 bug in the old `line_start_offset` implementation (it used `line.len() + 1` where `line` was already stripped of `\r\n`). The new code is strictly more correct. The existing tests were all title-based assertions — they would not have caught a byte-offset regression. Adding tests that assert on `loc.start` values directly is the right pattern for TextEdit providers; title tests are necessary but not sufficient.

What would have helped: the PR description didn't mention CRLF at all. Noting "the new byte scanner also fixes CRLF" would have made the scope clearer for reviewers.

---

- Logical next: green-ci verify on HEAD `ff88ff2d0`, then diff-auditor
- Out-of-scope follow-up: `find_pragma_insert_pos` in the same file still uses `line.len() + 1` on `str::lines()` output — same CRLF off-by-1 bug for pragma insertion position. Pre-existing, not introduced by this PR. Worth a targeted follow-up scout/fix.
- Gotcha: `SourceLocation { start, end }` in this crate are raw byte offsets, not char or UTF-16 offsets. The LSP→UTF-16 conversion happens upstream. Tests that assert on `.start` and `.end` must use byte arithmetic.
- Confidence for next: high

PR #6236, opened 2026-04-24
fix(ast): enforce traversal conformance for child-bearing parser nodes (#6494)

EffortlessSteven:
### Motivation

- Downstream consumers used hand-written per-`NodeKind` child traversal which allowed child-bearing variants (notably `StatementModifier`) to be skipped by diagnostics, semantic tokens, highlights and navigation providers.

### Description

- Replace ad-hoc `match NodeKind { ... }` child traversal in the diagnostics walker with the canonical `Node::for_each_child` contract, so `walk_node` visits every child-bearing child uniformly.
- Add table-driven traversal conformance tests in `diagnostics::walker` that assert `walk_node` visitation equals `Node::for_each_child` for representative snippets, including `print "ok" if $x = 5;`, nested modifiers, and all six modifiers (`if`, `unless`, `while`, `until`, `for`, `foreach`).
- Switch several consumers to rely on the same traversal contract: replace manual child extraction in `semantic_tokens::walk_ast_full`, `document_highlight::get_children` + call sites, and navigation `visit_children` with `for_each_child`-driven traversal to remove silent skip paths.
- Small refactor to unify recursion semantics: semantic token walker now propagates early returns correctly while using `for_each_child`.

### Testing

- Ran formatting: `cargo fmt` (succeeded).
- Ran focused tests for the new diagnostics walker: `cargo test -p perl-lsp-rs-core statement_modifier_visits_statement_and_condition && cargo test -p perl-lsp-rs-core nested_statement_modifiers_traverse_correctly && cargo test -p perl-lsp-rs-core all_statement_modifier_variants_traverse_correctly && cargo test -p perl-lsp-rs-core traversal_conformance_for_child_bearing_node_kinds` (all passed).
- Ran full crate test: `cargo test -p perl-lsp-rs-core` (most tests passed, but the workspace contains a pre-existing unrelated failing test `test_perl_lsp_cargo_toml_removed_g1b_imports` that expects a missing file; this is not caused by this change).
- Ran parser crate tests: `cargo test -p perl-parser-core` (this exposed an existing unrelated parser test failure `test_deref_hash_subscript_regex_key` in the workspace; not introduced by this patch).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53d0f078833391b11ffd2900db22)

PR #6237, opened 2026-04-24
fix(ast): enforce traversal conformance for child-bearing parser nodes (#6493)

EffortlessSteven:
### Motivation

- Downstream providers used hand-maintained `match NodeKind` child dispatch which can silently skip child-bearing variants (notably `StatementModifier`) causing diagnostics, semantic tokens, symbols, and highlights to miss parts of the AST.

### Description

- Replace the diagnostics walker with a simple pre-order traversal delegating to the canonical `Node::children()` contract so all child-bearing `NodeKind`s are visited consistently.
- Replace the ad-hoc child dispatch in the semantic-token walker (`walk_ast_full`) with iteration over `node.children()` to keep traversal logic unified and future-proof against `NodeKind` growth.
- Make `DocumentHighlightProvider::get_children` delegate to `Node::children()` (returning `None` for empty lists) so provider-local child discovery no longer drifts from the AST contract.
- Add unit tests in the diagnostics walker that assert the walker order matches a DFS over `Node::children()` for representative child-bearing nodes and add a parser-backed regression test that verifies `StatementModifier` visits both the statement and the condition for all six modifiers and nested modifiers.

### Testing

- Ran formatting: `cargo xtask fmt` / `cargo fmt` succeeded.
- Ran focused diagnostics walker unit tests: `cargo test -p perl-lsp-rs-core walker::tests::` succeeded (both new tests pass).
- Ran broader suites: `cargo test -p perl-lsp-rs-core` completed but one existing unrelated integration test failed (`green_tdd_wave_g1b_regression_hardening::test_perl_lsp_cargo_toml_removed_g1b_imports`), not introduced by this change.
- Ran parser tests: `cargo test -p perl-parser-core` completed but one existing unrelated parser regression test failed (`test_deref_hash_subscript_regex_key`), not caused by these traversal changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53d0f078833391b11ffd2900db22)


EffortlessSteven:
### Note: superseded by #6239

Same title and intent as #6236, #6237 (this PR), #6239, and #6313 -- four codex iterations of the same closeout. **#6239 is the strongest variant**:

- Uses `node.for_each_child` (allocation-free) instead of `node.children()`.
- Centralizes traversal as `walk_preorder` in `perl-parser-core/src/engine/ast.rs`.
- Conformance test lives in parser-core, next to the contract it pins.

Recommending close in favor of #6239. See full review there.

EffortlessSteven:

### Scope discipline: CLEAN
All three files changed (walker.rs, semantic_tokens.rs, document_highlight/mod.rs) are LSP providers that were explicitly mentioned in the issue motivation. The diff is surgical: removes duplicate child-traversal logic and replaces it with delegation to the canonical `Node::children()` interface. No unrelated refactoring, no feature creep.

This PR eliminates a problematic pattern (ad-hoc hand-maintained match statements for child dispatch) rather than introducing one. It enforces a *single* canonical source of truth for AST traversal via the existing `Node::children()` method in perl-ast. This is correct architectural practice.

### Quality bar: MET
- **Tests**: Two distinct tests with clear purposes. Unit test (`walk_node_matches_canonical_children_traversal_for_child_bearing_nodes`) verifies conformance against 11 representative node kinds; regression test (`statement_modifier_walks_statement_and_condition_for_all_modifiers`) directly validates the bug fix (StatementModifier children were being silently skipped in the walker).
- **Consistency**: Follows project idioms—delegates to canonical interfaces rather than reimplementing traversal logic.
- **Documentation**: No public API changes; internal refactoring. No doc debt introduced.
- **Backwards compatibility**: `Node::children()` is the existing public interface; providers now use it correctly.

### Concrete observation
The old `walker.rs` had a silent bug: `NodeKind::StatementModifier` fell into the `_ => {}` catch-all, causing diagnostics to skip the statement and condition nodes under postfix modifiers (if/unless/while/until/for/foreach). The new code fixes this by delegating to `Node::children()`, which correctly yields both children for every StatementModifier variant. The regression test validates this fix end-to-end with real parsed Perl code.

### Verdict: ALIGNED
Implementation fits perl-lsp's direction and quality bar. The conformance consolidation is justified by eliminating maintenance burden across three critical LSP providers and fixing a real correctness bug in diagnostics traversal.

---

EffortlessSteven:
Superseded by #6239 — same Codex task (task_e_69eb53d0f078833391b11ffd2900db22), same root cause fix, but #6239 is architecturally superior: it extracts the canonical `walk_preorder` into `perl-parser-core` (the right crate) and delegates everything to `Node::for_each_child`, whereas this PR keeps a hand-expanded walker in `perl-lsp-rs-core` which is the exact drift pattern we're trying to eliminate. #6239 has already received deep-reviewed. Issue #6493 is covered by the fix in #6239 (via issue #6492 / same spec).

PR #6238, opened 2026-04-24

EffortlessSteven:
### Motivation

- Downstream consumers had drifted from the AST child-contract causing some child-bearing variants (notably `StatementModifier`) to be skipped by diagnostics, semantic tokens, highlights and reference code.  
- Consistent traversal is required so diagnostics and other providers never silently omit subtree visits (e.g. postfix conditionals).

### Description

- Replaced the diagnostics walker implementation with a single canonical traversal that calls `node.for_each_child(...)` so every child-bearing `NodeKind` is visited consistently.  
- Added unit tests in the diagnostics walker to assert statement-modifier behavior (visits both statement and condition), nested modifier traversal, coverage of all six modifiers (`if`, `unless`, `while`, `until`, `for`, `foreach`), and a table-driven conformance test that compares walker visits to `count_nodes()` across many child-bearing node kinds.  
- Rewrote/updated other providers to rely on the shared child-contract: `walk_ast_full` in semantic tokens now uses `for_each_child` (and preserves early-return semantics), document highlights now recurses via the shared children iteration (while keeping special flattening for signatures/subroutines), and the semantic analyzer’s parent-map helpers use `node.children()` directly.  
- Removed several bespoke, hand-maintained match-based child enumerations so traversal behaviour is driven from the AST impl (`for_each_child` / `children()`), eliminating silent skip cases.

### Testing

- Ran `cargo test -p perl-lsp-rs-core diagnostics::walker::tests` and the new diagnostics walker tests passed (4 tests, all ok).  
- Ran `cargo test -p perl-lsp-rs-core --test diag_snap --test assignment_in_condition_tests` and those targeted provider tests passed.  
- Ran a broader `cargo test -p perl-lsp-rs-core` which exercised many unit tests; the workspace had one existing unrelated failure (`green_tdd_wave_g1b_regression_hardening::test_perl_lsp_cargo_toml_removed_g1b_imports`) due to a missing `crates/perl-lsp/Cargo.toml` in this environment and not caused by these changes.  
- Ran `cargo test -p perl-parser-core` which surfaced an unrelated parser test failure in this environment (`test_deref_hash_subscript_regex_key`) that is not caused by the traversal changes; the focused traversal tests and provider smoke tests above passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53d0f078833391b11ffd2900db22)


EffortlessSteven:
Redundant with #6237 (per parser-cluster triage 2026-04-25). Same Codex task scope. #6237 has minimal scope (3 LSP provider files only). #6238's declaration.rs extension can be cross-pollinated into #6239 (the architectural co-keeper) if novel. Keeping #6237 + #6239.

PR #6250, opened 2026-04-24
fix(parser): close quote-like regex substitution and heredoc accuracy gaps (#0000)

EffortlessSteven:
### Motivation

- Perl's quote-like operators (regex, substitution, transliteration, qw, heredoc) are a high-risk parser seam and the corpus still included several small buckets tied to delimiter and heredoc edge cases.
- A focused regression corpus is needed to lock in correct handling of unusual delimiters, paired/unpaired forms, whitespace-sensitive `qw` forms, slash-vs-division ambiguity, heredoc adjacency, and to ensure malformed-but-recoverable inputs preserve diagnostics.

### Description

- Added a new focused regression test file at `crates/perl-parser-core/tests/fix_quote_like_worklist_regressions.rs` exercising `s///` with unusual delimiters, `qr<>`/`qr{}`, `tr`/`y` paired and unpaired forms, `qw()` comment/space-sensitive forms, slash ambiguity contexts, heredocs adjacent to blocks/phase keywords/quote-like syntax, and malformed-but-recoverable quote-like inputs.
- The tests assert clean parsing for valid Perl idioms and assert specific diagnostics for invalid cases such as invalid substitution modifiers and missing replacement sections.
- The test expectations were aligned with the parser diagnostics text so failures surface as concrete assertions rather than brittle substring matches.

### Testing

- Ran `cargo test -p perl-parser-core --test fix_quote_like_worklist_regressions`, which passed.
- Ran `cargo test -p perl-parser-core quote`, which passed.
- Ran `cargo test -p perl-parser-core regex`, which passed in the full-suite invocation (note: an earlier isolated run surfaced an unrelated existing test which was investigated during the session but did not indicate regression from this change).
- Ran `cargo test -p perl-parser-core heredoc` and `cargo test -p perl-parser-core slash_ambiguity`, both of which passed.
- `just parser-audit`, `just corpus-sweep-check`, and `just cpan-corpus-check` were not executed here because `just` is not available in the execution environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53ddfec883339190e20f2c4da713)

### DIFF
--- a/crates/perl-parser-core/tests/fix_quote_like_worklist_regressions.rs
+++ b/crates/perl-parser-core/tests/fix_quote_like_worklist_regressions.rs
@@ -1,0 +1,55 @@
+//! Focused corpus-style regressions for quote-like, regex, and heredoc seams.
+//! These cases mirror worklist categories that historically appeared in sweep buckets.
+
+mod cpan_test_helpers;
+use cpan_test_helpers::{assert_clean_parse, assert_has_error};
+
+#[test]
+fn subst_unusual_delimiter_pairs_clean_parse() {
+    assert_clean_parse(r#"$x =~ s<foo><bar>g;"#);
+    assert_clean_parse(r#"$x =~ s{foo}[bar]r;"#);
+    assert_clean_parse(r#"$x =~ s!foo!bar!;"#);
+}
+
+#[test]
+fn qr_paired_delimiters_clean_parse() {
+    assert_clean_parse(r#"my $re1 = qr<foo\s+bar>i;"#);
+    assert_clean_parse(r#"my $re2 = qr{(?:a|b){2}}x;"#);
+}
+
+#[test]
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6251 implements the same quote-like parser fixes with integrated test coverage.

Both address the same issue (unusual delimiters, mixed paired/non-paired forms in s///, tr, qr, qw), but #6251 includes the actual parser changes to quote.rs whereas this PR is test-only. The implementation in #6251 is the authoritative fix; its tests are sufficient.

**Extracted and noted for #6251 consideration**: Your regression test names and structure are clear and well-organized (fix_quote_like_worklist_regressions.rs with focused assertion patterns). Recommend merging test names if #6251's tests are less comprehensive.

Thank you — the test-first approach helps validate the fix approach.

PR #6252, opened 2026-04-24
fix(parser): handle typed catch with block syntax (#6494)

EffortlessSteven:
### Motivation
- Prevent valid Perl `catch Class::Name with { ... }` forms from producing `expected_left_brace`/block-shape ambiguities by recognizing Error.pm-style typed catch filters.

### Description
- Extend `parse_try` in `crates/perl-parser-core/src/engine/parser/control_flow.rs` to consume a class-like filter sequence (identifier / `::`) after `catch` and require the `with` keyword before the block, emitting a diagnostic if `with` is missing.
- Add a diagnostic-aware test helper and two focused regressions in `crates/perl-parser-core/tests/fix_expected_left_brace.rs` that assert the valid `catch ... with { }` form parses cleanly and the same form without `with` produces the new diagnostic.
- Implementation is conservative: it only accepts class-like filters in the `catch` position and surfaces an explicit syntax error instead of silently accepting invalid variants.

### Testing
- Ran `cargo test -p perl-parser-core fix_expected_left_brace` and the suite passed.
- Ran `cargo test -p perl-parser-core declaration`, `cargo test -p perl-parser-core signature`, `cargo test -p perl-parser-core native_class`, and `cargo test -p perl-parser-core variable_attributes` and all targeted tests passed.
- Ran `cargo run -p xtask --no-default-features -- corpus-audit --fresh --corpus-path .` and `cargo run -p xtask -- parser-corpus-sweep --output /tmp/parser-sweep-after.json`; the sweep completed and the `expected_left_brace` first-error count decreased (5 -> 4) in the system corpus.
- Notes: `just parser-audit` was not available in this environment and enforcing the committed baseline (`parser-corpus-sweep --baseline ... --enforce`) fails here due to the environment baseline mismatch, not due to the change itself.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53e1a7488333870bd40667cfce8f)


EffortlessSteven:
### Review: handle typed catch with block syntax

Targets `Error.pm`-style typed catch:

```perl
catch Some::Error with { ... }
catch Other::Error::Subclass with { ... }
```

Without the `with` keyword expected, the parser sees `Some::Error` after `catch`, then a `{`, and tries to read it as a block label or fails. This PR teaches the catch parser to absorb a chain of `Identifier (`::` Identifier)*` between `catch` and the block, *but only when the chain is followed by a literal `with` token* -- which is the right precision filter, because plain `catch ($e) { ... }` and `catch { ... }` (no filter) must continue to parse without `with`.

**Correctness checks**:

- The existing `var.is_none()` guard means the typed-catch path only fires when there's no `($e)` capture variable -- correct, because `catch Some::Error ($e) with { ... }` doesn't seem to be valid Error.pm syntax. Worth a one-line doc-comment confirming this. If it *is* valid in some dialects, the gate may need to widen.
- The `peek_kind() != LeftBrace` guard means plain `catch { ... }` (no filter, no `with`) still parses normally. Good.
- The `consumed_filter` flag ensures the synthetic "Expected 'with' before catch block" diagnostic only fires when there *was* something class-like consumed -- otherwise an empty `catch ` followed by garbage would mis-attribute the error to a missing `with`. Correct.

**Style nit (non-blocking)**: the inline `is_component = kind == TokenKind::Identifier && self.tokens.peek()?.text.as_ref() != "with"` reads better hoisted into a small `is_class_path_component` helper, especially since the same `text.as_ref() != "with"` check is repeated implicitly when terminating the loop. Optional.

**Verdict**: targeted, correct, well-bounded. The strict `with` requirement is the right call -- it preserves error quality when someone writes `catch SomeClass { ... }` accidentally. Pre-existing master CI failures are unrelated; #6495 fixes the clippy_full half.

EffortlessSteven:
## Deep Correctness Review — APPROVED

**Verified:** `parse_try` typed-catch extension correctly handles Error.pm-style `catch ClassName with { }` forms without regressing the existing paren-variable form `catch ($e) { }`.

### What I checked

**Control flow logic (`control_flow.rs`):**
- The guard `var.is_none() && peek_kind() != LeftBrace` correctly restricts class-filter consumption to the case where no paren variable was already parsed. `catch ($e) ClassName with { }` is not a real Error.pm combination, so skipping filter consumption in that case is correct.
- The `while let Some(kind) = self.peek_kind()` filter loop correctly excludes `with` via `text.as_ref() != "with"` before consuming any Identifier token. The loop breaks at the first non-identifier/non-`::` token, leaving `with` to be consumed by the explicit check after.
- The diagnostic "Expected 'with' before catch block" is pushed via `self.errors.push(...)` which maps directly to `output.diagnostics` in `ParseOutput::with_errors`. The `assert_has_diagnostic` test helper correctly queries that field.
- When `consumed_filter = true` and `with` is missing but the next token IS `{`, the code pushes an error and then `parse_block()` succeeds — producing a recovered AST with a diagnostic. The test `test_try_catch_typed_without_with_still_errors` correctly validates this.

**Edge cases evaluated:**
- `catch { }` (no var, no filter, bare block): guard `peek_kind() != LeftBrace` is false → filter loop skipped → `parse_block()` called directly. Correct.
- `catch SomeClass::Sub::Error with { }` (multi-segment name): loop correctly consumes all `Identifier` and `DoubleColon` tokens before breaking at `with`. Correct.
- EOF mid-filter: `while let Some(kind) = self.peek_kind()` naturally exits on None → `consumed_filter` may be true → error pushed → `parse_block()` propagates Err up. Recovered by `parse_with_recovery`. Acceptable.
- `catch with { }` (class named "with"): the `is_component` check short-circuits on `text == "with"`, so filter loop breaks immediately, `consumed_filter = false`, no action → `parse_block()` called on `with` which fails. This is an obscure edge case (class named "with" is not standard Perl) and is not a regression.

**Tests:**
- `test_try_catch_typed_with_block_is_valid`: uses `assert_no_diagnostics` which checks both `output.diagnostics` and sexp for error markers. Sufficient for the happy path.
- `test_try_catch_typed_without_with_still_errors`: verifies the exact diagnostic message text. Correct.

**Corpus impact:** PR body notes `expected_left_brace` first-error count decreased 5→4 in the system corpus sweep. No regression observed.

No bugs found. The implementation is conservative and correct for its stated scope.

EffortlessSteven:
## Green-CI Verdict: GREEN

All required checks pass on current HEAD (db0a40dd8d97bbce2ed6f7f8944010ef663b7a12):
- **PR Smoke (Fast Feedback)**: SUCCESS
- **CI Gate (Merge-Blocking)**: SUCCESS
- **UX Regression Tests**: SUCCESS
- **Windows Guardrails**: All SUCCESS
- **Draft PR guard**: SUCCESS
- **Compile All Targets**: SUCCESS
- **GitGuardian**: SUCCESS

PR is not draft, mergeable state is UNKNOWN (no conflicts detected).

Label: ci-green set ✓

Ready for ops merge.

EffortlessSteven:

### Scope discipline: CLEAN
PR implements typed catch parsing per issue #6494 — adds logic to consume class-like filters (Foo::Bar::Baz) when they appear between 'catch' and the block, and enforces 'with' keyword. Two test cases verify the happy path (with 'with') and the error case (without 'with'). Matches the spec exactly.

Follows existing parser error reporting (ParseError::syntax with position). No new patterns.

### Quality bar: MET
- Tests verify both valid syntax (Git::Error::Command with {...}) and error detection ('with' requirement).
- Uses existing error assertion helper (assert_has_diagnostic).
- Covers the constraint: class-like filter → 'with' keyword required before block.

### Verdict: ALIGNED
Parser improvement with focused, correct test coverage. Addresses a real Perl::Try::Tiny / Error.pm syntax variant that was previously mis-parsed.

---

EffortlessSteven:

Parser enhancement for Error.pm-style typed catch syntax:
- Adds parser support for 'catch Type::Name with { ... }' pattern
- Tests validate both the valid syntax and the error case (missing 'with')
- Error messages are clear ("Expected 'with' before catch block")
- No unexpected scope creep; focused on control flow parsing
- Follows existing parser patterns and error handling

Ready for merge once CI gates pass.

PR #6254, opened 2026-04-24
fix(parser): close declaration and block-shape corpus ambiguity buckets (#6489)

EffortlessSteven:
### Motivation
- Reduce false-positive parser errors in signature and block-shape related buckets by handling invocant separators and phase-keyword labels where valid Perl permits them.

### Description
- Accept a single invocant `:` separator inside `parse_signature` so method/sub signatures like `($self: $arg)` parse cleanly without triggering `signature_param` errors. (`crates/perl-parser-core/src/engine/parser/variables.rs`)
- Allow phase-keyword tokens (`BEGIN`, `END`, `CHECK`, `INIT`, `UNITCHECK`) as valid loop-control label targets so constructs like `last CHECK;` and `CHECK: while (...) { last CHECK; }` are accepted. (`crates/perl-parser-core/src/engine/parser/statements.rs`)
- Added focused regression tests covering invocant signatures, signature-span coverage for invocant forms, and phase-keyword labels targeted by loop-control. (tests: `fix_signature_param_validation_3359.rs`, `fix_signature_span_4243.rs`, `fix_phase_keyword_as_label_2389.rs`)

### Testing
- Ran `cargo test -p perl-parser-core declaration` and relevant declaration tests passed locally. 
- Ran `cargo test -p perl-parser-core signature` and signature tests (including new invocant tests) passed locally. 
- Ran `cargo test -p perl-parser-core native_class` and `cargo test -p perl-parser-core variable_attributes` and both sets passed. 
- Performed `cargo check --all-targets -p perl-parser-core`, `cargo clippy -p perl-parser-core`, and `cargo xtask fmt` which completed successfully. 
- Ran the repo corpus audit via `cargo run -p xtask -- corpus-audit --fresh --corpus-path .` which produced `corpus_audit_report.json` with the repo corpus parsing cleanly; system/CPAN-wide sweep gates (`parser-corpus-sweep` / `cpan-corpus sweep`) were not completed due to local baseline/corpus availability mismatches and therefore reported environment-related failures (not regressions from this change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53e1a7488333870bd40667cfce8f)

PR #5679, opened 2026-04-23
docs(editors): add Codex CLI LSP bridge setup (#0000)

EffortlessSteven:
### Motivation

- Add documentation so Codex CLI users can integrate `perllsp` via an MCP LSP bridge and use real LSP operations from Codex sessions.

### Description

- Create `docs/EDITORS/CODEX_CLI_SETUP.md` with a sample MCP bridge config, registration example for `perllsp --stdio`, validation steps, and troubleshooting, and add a link/short section to `docs/how-to/EDITOR_SETUP.md` pointing to the new guide.

### Testing

- Ran `cargo check -p perl-lsp-rs-core` and `cargo test -p perl-lsp-rs-core --no-run`, both of which completed successfully, and attempted `cargo xtask fmt` which failed due to pre-existing, unrelated compile errors in `xtask/src/tasks/metrics/lsp_stats.rs`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea3375217c83339aeb5fddce081231)

PR #6255, opened 2026-04-24
fix(parser): close declaration and block-shape corpus ambiguity buckets (#6493)

EffortlessSteven:
### Motivation

- Reduce parser ambiguity where valid Perl was misclassified into `expected_left_brace`, `expected_colon`, `expected_variable`, `expected_identifier`, and `signature_param` buckets by handling `class` as a DSL call and signature parameter attributes correctly.

### Description

- Narrow entry into native `class` declaration parsing so `class` is only treated as a declaration when the following token shape (`Identifier`, `DoubleColon`, or `:`) matches a class header, otherwise it remains a normal expression/call (`crates/perl-parser-core/src/engine/parser/statements.rs`).
- Extend signature parameter parsing to consume trailing parameter attributes/traits and their optional parenthesized argument lists (e.g. `:$attr(...)`) so such attributes do not trigger signature-separator recovery (`crates/perl-parser-core/src/engine/parser/variables.rs`).
- Add focused regressions: a signature-attribute regression in `fix_signature_param_validation_3359.rs` and a guard ensuring bareword `class(...)` calls are not forced into native-class parsing in `native_class_isa_tests.rs`.

### Testing

- Ran targeted test suites and they passed: `cargo test -p perl-parser-core declaration` (passed), `cargo test -p perl-parser-core signature` (passed), `cargo test -p perl-parser-core native_class` (passed), and `cargo test -p perl-parser-core variable_attributes` (passed).
- Ran the focused test files `fix_signature_param_validation_3359` and `native_class_isa_tests` via `cargo test -p perl-parser-core --test fix_signature_param_validation_3359 --test native_class_isa_tests` and both completed successfully.
- Attempted higher-level corpus gates (`just parser-audit`, `just corpus-sweep-check`, `just cpan-corpus-check`) but `just` is not available in this environment so those commands were not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53e1a7488333870bd40667cfce8f)


EffortlessSteven:
### Review: `class` as identifier + signature parameter attributes

**Not a duplicate of #6254 despite the shared title** -- this PR closes two distinct ambiguity buckets:

1. **`class` keyword used as a regular identifier** in legacy code. The fix peeks the *second* token: `parse_class()` only fires if the next token is `Identifier`, `DoubleColon`, or `Colon` (covering `class Foo`, `class Foo::Bar`, and `class Foo : isa(Base)`). For anything else (`class()`, `$class->method`, `my $class = ...`), fall through to the regular expression parser. This is the right disambiguation strategy.

2. **Signature parameter attributes**: `sub f ($x :param, $y :reader(foo)) { ... }` -- Object::Pad-style and frameworks like Future::AsyncAwait emit these in the wild. The new `consume_signature_param_attributes` helper consumes `: ident` (with optional `(...)` argument balanced by depth tracking) and reports unterminated parens cleanly.

**Correctness checks**:

- The `Class` arm's guard:
  ```rust
  if matches!(
      self.tokens.peek_second().map(|t| t.kind),
      Ok(TokenKind::Identifier) | Ok(TokenKind::DoubleColon) | Ok(TokenKind::Colon)
  )
  ```
  is conservative -- if `peek_second` returns an `Err`, the guard fails and we fall through to non-class parsing. Correct: a lexer error shouldn't push us into class parsing.
- Worth a one-line comment confirming this is the intended "fall through on lexer-error" behavior, otherwise the next reader will assume the `Err` arm is dead code.
- The depth-tracking in `consume_signature_param_attributes` correctly handles nested parens like `:reader(foo(bar))`. The unterminated case produces a syntax error rather than silently EOF-ing -- right call.

**Suggestion (non-blocking)**: the parameter attributes are currently *consumed but not preserved*. The doc-comment says "preserve only span/shape for now" -- worth filing a follow-up to actually attach them to the AST node so semantic analysis can see them later (especially for `:reader(name)` which downstream class-modeling code will care about). Optional, separate concern.

**Verdict**: real bug fixes, well-scoped, lands cleanly alongside #6254. Pre-existing master CI failures are unrelated; #6495 fixes the clippy_full half.

EffortlessSteven:
Linkage check: linked issue #6493 is already closed. Verify this PR is still needed; if it duplicates work merged via another PR that closed #6493, close this one as superseded.

EffortlessSteven:
## Deep Correctness Review — REBASE REQUIRED, CONDITIONAL APPROVE

**Status:** PR is CONFLICTING vs master (master is 61 commits ahead; `statements.rs`, `variables.rs`, and `fix_signature_param_validation_3359.rs` all have upstream changes). Rebase is required before this can land. The correctness analysis below is based on the logic as written — it's sound, with one guard concern noted.

### What I verified

**`class` keyword guard (`statements.rs`):**

The narrowed guard correctly prevents `class("Widget::Role")` from dispatching to `parse_class()`. When the guard fails (peek_second is `LeftParen`, `LeftBrace`, EOF, or any other non-guarded token), the `_` arm calls `parse_expression_statement()`, which routes through `expressions/primary.rs` where `TokenKind::Class` is treated as a bareword identifier — correct call-expression parsing.

One concern: the guard includes `Ok(TokenKind::Colon)` for the case where `class :isa(...)` would be written without a class name. But `parse_class()` unconditionally calls `parse_qualified_name(false)` which immediately calls `expect(TokenKind::Identifier)?` — this will fail when the first token after `class` is `:`. The Colon arm in the guard produces a worse result (parse error) than the fallthrough (expression statement). However, `class :isa(Foo) { }` without a name is not valid Perl 5.38+ (`class` requires a name), so this is not a regression against real code. Similarly for `DoubleColon` (no Perl class starts with `::`).

Recommendation: after rebase, consider removing `Ok(TokenKind::DoubleColon) | Ok(TokenKind::Colon)` from the guard unless `parse_class` is updated to handle leading `::` or name-less class syntax. For now this is a latent error path, not a real regression.

**Signature attribute parsing (`variables.rs`):**

`consume_signature_param_attributes` is called after `parse_variable()` and before the default-value check. The ordering is correct for `$x :param = 5`:
1. Parse `$x`
2. Consume `:param`  
3. See `= 5`, parse as default

The depth-tracking loop for parenthesized attribute args (`($x :reader(foo, bar))`) correctly handles nested parens and commas inside the arg list. Returning `Err` when `depth != 0` (unterminated attribute parens) is the right escape hatch.

The `named` parameter detection at the start of `parse_signature_param` checks for a leading `:` before the variable, while `consume_signature_param_attributes` checks for `:` after the variable. These are non-overlapping because the leading `:` is consumed (setting `named = true`) before `parse_variable()` runs, so the trailing `:param` scan never sees the leading colon.

Named parameter default values are silently dropped in the NodeKind selection (`NodeKind::NamedParameter` takes only `variable`). This is a pre-existing limitation, not introduced here.

**Test coverage:**
- `test_signature_param_attributes_parse_without_signature_separator_error`: checks that the exact error string "Expected comma or closing parenthesis in signature" is absent. This is the right regression guard since that message comes from `parse_signature` when it sees `:` where it expects `,` or `)`.
- `bareword_class_call_is_not_forced_into_native_class_declaration`: `class("Widget::Role")` is the key positive regression test for the guard change. Correct.

**Action required:** Rebase onto current master. After rebase, apply `deep-reviewed`. No logic bugs block approval — the implementation is correct for its scope.

EffortlessSteven:
## Maintainer-PR review: ALIGNED

Closes two corpus dirt buckets in one diff with proper guards:

1. **Class-keyword disambiguation**: `class` only triggers native-class parsing when followed by `Identifier | DoubleColon | Colon`. This protects DSL code that exports `class` as a normal function (`class("Widget::Role")`), method calls (`class->new()`), and fat-arrow autoquote (`class => ...`).
2. **Signature parameter attributes**: Adds `consume_signature_param_attributes` to handle `sub f ($x :param, $y :reader(get_x))`, with paren-balancing for nested attribute arguments and proper `end` tracking that respects subsequent default values.

Test coverage is thorough: 8 new edge-case regression tests in `fix_class_disambig_edge_cases.rs` covering bareword call, hash-key autoquote, multi-attr, nested parens in attr args, attr+default coexistence, package-qualified class names, and named-param disambiguation. Plus regression tests in the existing `fix_signature_param_validation_3359.rs` and `native_class_isa_tests.rs`.

This is squarely on the parser-accuracy ratchet path toward the v0.13.0 release — Object::Pad-style code in the wild is a real pattern we need to handle. Architectural fit is correct (changes scoped to `engine/parser/{statements,variables}.rs`, no leakage into other layers).

Quality bar met: real fix with proper guards, comprehensive edge cases, clear comments at fix sites explaining the invariant.

Note: PR is currently `DIRTY` (merge conflict). Operator should rebase before merge.

EffortlessSteven:
Merge conflict resolved via cherry-pick (fresh-root master rebuild scenario per the #5679->#6716 pattern; no merge-base existed). Cherry-picked the 3 unique commits (`cf79349ff` parser fix, `91d782977` red tests, `7cb457ffc` additional tests) onto current master. The redundant `522c47f93` xtask/ci_scope.rs commit was dropped — master already contains `perl-workspace-index` rename.

One semantic conflict required additional fix: this PR's new `consume_signature_param_attributes` helper interacted poorly with master's #6254 invocant-separator support, breaking `test_invocant_only_signature_is_accepted` and `test_double_invocant_separator_is_rejected`. Added a peek_second guard so the helper only fires on true bareword identifiers (not sigil-prefixed variable tokens), leaving bare `:` for parse_signature's invocant branch. New commit `bf3e05476` documents the interaction.

Local verification: `cargo test -p perl-parser-core --test fix_class_disambig_edge_cases --test fix_signature_param_validation_3359 --test native_class_isa_tests` -> 13 + 17 + 7 = all pass. Lib tests: 578/579 pass; the single failure (`syntax::path_security::tests::test_resolve_completion_base_rejects_absolute`) reproduces on plain master (pre-existing bit-rot, not introduced by this PR). `xtask fmt` clean. CI re-running on rebased HEAD.

EffortlessSteven:
## Diff Audit: CLEAN

Reviewed full diff: 5 files, +157/-3.

- `statements.rs`: narrows `TokenKind::Class` arm with `peek_second` guard (Identifier | DoubleColon | Colon)
- `variables.rs`: adds `consume_signature_param_attributes` helper (rejects sigil-prefixed identifiers as attributes, balances nested parens, returns updated `end`); wires it into `parse_signature_param`
- `fix_class_disambig_edge_cases.rs`: 8 new edge-case regressions (method call, fat-arrow autoquote, multiple attrs, nested parens in attr args, attr+default coexistence, package-qualified class name, named-param colon)
- `fix_signature_param_validation_3359.rs`: +1 regression
- `native_class_isa_tests.rs`: +1 regression for bareword `class("Widget::Role");`

All edits are scoped to `perl-parser-core`. The implementation correctly handles span end with attributes-then-default, the `peek_second` guard preserves both DoubleColon (qualified class name) and Identifier paths. Tests directly exercise the changed code paths. No artifacts, no contamination, no scope drift.

Verdict: CLEAN.

EffortlessSteven:
## Refactor analysis

**Verdict: No refactor needed.**

The class-keyword guard in `statements.rs` is a tightly-scoped four-line `match` arm guard (`peek_second()` discriminates Identifier / DoubleColon / Colon), and the `consume_signature_param_attributes` helper in `variables.rs` is already extracted out of `parse_signature_param` with a clear narrow contract (return updated `end` offset; balance nested parens via depth counter; reject sigil-prefixed identifiers as method-invocant disambiguation). The `next_is_attr_ident` predicate inside the helper reads cleanly. The eight-test edge-case coverage in `fix_class_disambig_edge_cases.rs` is appropriately exhaustive for a parser disambiguation fix.

EffortlessSteven:
PR-responder: gemini-code-assist and coderabbitai both rate-limit notices — no actionable suggestions. PR Smoke fmt failure is master bit-rot in `crates/perl-pragma/tests/comprehensive_unit_tests.rs` — PR (perl-parser-core declaration disambiguation) doesn't touch perl-pragma. Recommend cascade-update after master fmt fix lands. Ready for next gate.

EffortlessSteven:
ci-green sweep: PR Smoke (Fast Feedback) failed on Check formatting (perl-pragma/Cargo.toml fmt). Routing to needs-ci-fix.

EffortlessSteven:
Cascade-update post #6849 (master fmt fix) + #6852 (xtask gates demote-to-advisory). CI re-running.

PR #6257, opened 2026-04-24
fix(parser): recover delimiter handoff ownership (#6492)

EffortlessSteven:
### Motivation
- Reduce false-positive unclosed-delimiter diagnostics when an outer closing delimiter appears while an inner closer is missing, by classifying those as ownership handoff recovery cases.

### Description
- Update `expect_closing_delimiter` to treat mismatched closing delimiters as handoff points by consulting a new helper `is_delimiter_handoff_point`, so the parser can insert the expected closer and leave the outer closer for the enclosing frame (`crates/perl-parser-core/src/engine/parser/helpers.rs`).
- Add `is_delimiter_handoff_point(expected: TokenKind)` which recognizes symmetric mismatches between `)`, `]`, and `}` and signals an InsertedCloser recovery without consuming the current token (`crates/perl-parser-core/src/engine/parser/helpers.rs`).
- Add two focused regression tests that exercise bracket↔paren handoff recovery diagnostics to lock the behavior: missing inner `]` with outer `)` and missing inner `)` with outer `]` (`crates/perl-parser-core/tests/recovery_missing_closer.rs`).

### Testing
- Ran `cargo test -p perl-parser-core recovery`, and the targeted recovery tests passed (all recovery-related tests reported OK).
- Ran `cargo test -p perl-parser-core paren` and `cargo test -p perl-parser-core bracket`, both suites passed.
- Ran `cargo test -p perl-parser-core unclosed`, which passed.
- Attempted `just parser-audit`, `just corpus-sweep-check`, and `just cpan-corpus-check` but `just` is not available in this environment; other `cargo test` verification was performed and succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53e6c3288333b7319bdc43d99a28)

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/helpers.rs
+++ b/crates/perl-parser-core/src/engine/parser/helpers.rs
@@ -703,7 +703,7 @@ impl<'a> Parser<'a> {
             self.consume_token()?;
             return Ok(());
         }
-        if self.is_delimiter_recovery_point() {
+        if self.is_delimiter_recovery_point() || self.is_delimiter_handoff_point(kind) {
             let pos = self.current_position();
             let site = Self::recovery_site_for_closer(kind);
             self.errors.push(ParseError::Recovered {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Superseded by #6259 (same Codex burst, task_e_69eb53e6c3...).

**Same scope**: All three PRs in this Codex burst (#6257, #6258, #6259) target the same delimiter-recovery owner-bucket reduction. They are different implementations of the same goal.

**Why #6259 wins**:
- Full sign-off pipeline (`deep-reviewed`, `maintainer-pr-reviewed`, `diff-audited`), CI running, MERGEABLE.
- This PR is currently CONFLICTING with master and has no sign-off labels.
- #6259 has broader test coverage (14 owner-bucket tests vs. 2 handoff tests here) and covers more code paths (postfix.rs + statements.rs vs. helpers.rs only).

**What this PR adds that #6259 doesn't**: `is_delimiter_handoff_point` for symmetric sibling-closer mismatches (`)` while expecting `]`, etc.). This is a complementary refinement to the site-aware approach. File a follow-up issue if cross-bracket handoff scenarios remain in the corpus after #6259 lands.

PR #6258, opened 2026-04-24
fix(parser): classify and close delimiter recovery corpus buckets (#6491)

EffortlessSteven:
### Motivation
- Delimiter-recovery logic treated many delimiter-follow patterns as hard errors, inflating unclosed-delimiter buckets in corpus analysis.
- The goal is to make missing-closer recovery owner-aware so only genuinely malformed input gets hard errors while common valid Perl patterns recover with structured diagnostics.

### Description
- Make `expect_closing_delimiter` owner-aware by accepting the expected `TokenKind` and checking for owner-specific follower tokens before inferring an inserted closer (`InsertedCloser`).
- Harden follower logic so `)`/`]`/`}` recovery considers nearby owners like `->`, outer closers, and `=>` to avoid cascaded unclosed errors.
- Switch several postfix/hash-subscript brace expectations to use `expect_closing_delimiter(...)` so missing `}` in postfix/subscript contexts emits structured recovery rather than a hard parse error.
- Add focused tests in `recovery_missing_closer.rs` covering missing `}` before `;`, missing `}` before `->` (postfix chains), and missing `)` before `=>`, and adjust existing tests to assert `InsertedCloser` recoveries.

### Testing
- Ran `cargo test -p perl-parser-core --test recovery_missing_closer` and the test binary passed (all recovery assertions OK).
- Ran `cargo test -p perl-parser-core recovery`, `cargo test -p perl-parser-core unclosed`, `cargo test -p perl-parser-core paren`, and `cargo test -p perl-parser-core bracket` and all targeted suites passed.
- Ran `cargo check --all-targets -p perl-parser-core`, `cargo clippy -p perl-parser-core -- -D warnings`, and `cargo xtask fmt`, all of which completed successfully.
- `just parser-audit`, `just corpus-sweep-check`, and `just cpan-corpus-check` were not executed because `just` is not available in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53e6c3288333b7319bdc43d99a28)


EffortlessSteven:
### Review: classify and close delimiter recovery corpus buckets (helpers + recovery-point logic)

**Note: PR #6259 has the same title and shares the postfix.rs change.** They diverge after that:

- **This PR (#6258)**: replaces `self.expect(TokenKind::RightBrace)` with `self.expect_closing_delimiter(...)` at four sites in `expressions/postfix.rs`, AND modifies `helpers.rs::is_delimiter_recovery_point` to take an `expected: TokenKind` parameter so the recovery-point classification depends on what closer was wanted. This is the *deeper* change.
- **#6259**: same four `postfix.rs` edits, plus one `statements.rs` edit at the for-loop's closing paren, plus a new test file `fix_delimiter_owner_buckets.rs`.

**Recommendation: land this PR's helper/recovery-point change AND #6259's `statements.rs` site + the new test file**. They're not exclusive but they need to merge in the right order:

1. Land #6258 first (introduces the `expect_closing_delimiter` machinery and the deeper `is_delimiter_recovery_point(expected)` signature).
2. Rebase #6259 onto master, drop the now-redundant postfix.rs edits, keep the `statements.rs` edit and the new test file.

Trying to merge them in the opposite order will produce a conflict in `postfix.rs` (already-applied edits) and #6259 will silently lose the helper improvements.

**Correctness check on the recovery-point change**:

```rust
fn is_delimiter_recovery_point(&mut self, expected: TokenKind) -> bool {
```

Threading the *expected* kind into the recovery decision means we can correctly distinguish "we were expecting `)` but hit `;`" (recoverable, the user forgot the close paren) from "we were expecting `}` but hit `;`" (less recoverable in some contexts). Worth confirming the recovery sites the helper is called from can all genuinely benefit from the kind information -- if any caller passes an arbitrary `TokenKind` that doesn't matter, the parameter is added cost without payoff. From a quick scan, the four `postfix.rs` callsites all pass a closing delimiter (`}` or similar), so the precision is real.

**Verdict**: solid foundational change, should land first in the cluster. Pre-existing master CI failures are unrelated; #6495 fixes the clippy_full half.

EffortlessSteven:
Superseded by #6259 (same Codex burst, task_e_69eb53e6c3...).

**Overlap**: Both PRs make the identical 4-site change in `postfix.rs` (lines 64, 163, 326, 599) replacing `expect(RightBrace)` with `expect_closing_delimiter(RightBrace)` in hash subscript / hash slice paths.

**Why #6259 wins**:
- Full sign-off pipeline (`deep-reviewed`, `maintainer-pr-reviewed`, `diff-audited`), CI running, MERGEABLE.
- Broader test coverage: new `fix_delimiter_owner_buckets.rs` with 14 owner-bucketed tests + 4 `RecoverySite` assertions (HashSubscript, ArgList, etc.) in `recovery_missing_closer.rs`.
- Also extends `statements.rs` `tie()` paren handling.

**What this PR adds that #6259 doesn't**: owner-aware follower logic in `is_delimiter_recovery_point` (recognizes `->`, `=>`, sibling closers as recovery points). This is a complementary refinement that could land as a follow-up issue if #6259's site-aware approach proves insufficient. File a follow-up if delimiter buckets remain noisy after #6259 merges.

PR #6267, opened 2026-04-24

EffortlessSteven:
### Motivation
- While typing, users frequently produce incomplete interpolation indexers inside double-quoted strings (e.g. `$hash{incomplete`, `$array[0`, `$obj->{field`) and the lexer could consume the closing quote which caused the parser to emit top-level ERROR nodes and lose useful AST for symbol extraction.

### Description
- Add string-aware balanced-segment scanning in the lexer so interpolation tail scans stop at the string closing quote instead of consuming the rest of the literal; this preserves the outer string token and avoids ERROR propagation (changes in `crates/perl-lexer/src/lib.rs`).
- Add a parser-side diagnostic pass that inspects interpolated `TokenKind::String` text and records a targeted syntax diagnostic for unclosed `{` or `[` interpolation tails without converting the AST into ERROR nodes (`crates/perl-parser-core/src/engine/parser/expressions/primary.rs`).
- Add focused tests exercising incomplete and complete interpolation scenarios to ensure variables like `$hash`/`$array` remain present in the AST and diagnostics are still produced (`crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs`).

### Testing
- Ran `cargo test -p perl-parser-core --test string_interpolation_incomplete_tests` and the new focused tests passed.
- Ran `cargo test -p perl-parser-core string_interpolation` and the targeted string-interpolation test group passed.
- Ran `cargo test -p perl-parser-core -q` (full crate test); the suite mostly passed but there remains a separate existing failing test `test_deref_hash_subscript_regex_key` which predates these changes and still fails (observed during verification).
- `just parser-audit` and `just cpan-corpus-check` could not be executed in this environment because `just` is not installed.

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -864,7 +864,12 @@ impl<'a> PerlLexer<'a> {
     }
 
     #[inline]
-    fn consume_balanced_segment(&mut self, open: char, close: char) -> Option<usize> {
+    fn consume_balanced_segment(
+        &mut self,
+        open: char,
+        close: char,
+        string_delim: Option<char>,
+    ) -> Option<usize> {
         if self.current_char() != Some(open) {
             return None;
         }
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #6268, opened 2026-04-24

EffortlessSteven:
### Motivation

- Incomplete indexing inside double-quoted interpolation (e.g. `"Key: $hash{incomplete"` or `"Element: $array[0"`) could cause the lexer to consume the closing `"` and produce an unterminated-string/ERROR node that pollutes the top-level AST and breaks symbol extraction.

### Description

- Add `consume_balanced_segment_in_double_quote` in the lexer to parse balanced segments inside interpolations but stop before an unescaped closing `"` so the outer string token remains intact.
- Replace interpolation-tail balanced-segment calls in the double-quote interpolation path to use the new `consume_balanced_segment_in_double_quote` for `[...]`, `{...}`, and `(...)` contexts to avoid swallowing the string terminator.
- Add parser-level regression tests in `crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs` covering incomplete hash and array interpolation forms and a baseline complete case.
- Add semantic-analyzer tests in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs` asserting that base variable references for `$hash` and `@array` remain extractable when their interpolation indexing is incomplete.

### Testing

- Ran `cargo test -p perl-parser-core string_interpolation` and it passed.
- Ran `cargo test -p perl-parser-core double_quote_incomplete` and the new/in-scope tests passed (`4` tests reported as passing).
- Ran `cargo test -p perl-parser-core` which mostly passed but failed one pre-existing unrelated test (`test_deref_hash_subscript_regex_key`) producing an `Error` node; this failure appears unrelated to the interpolation change.
- Ran `cargo test -p perl-semantic-analyzer` filtered to the new tests (via name prefix) and the two new semantic tests passed.
- Could not run `just parser-audit` or `just cpan-corpus-check` in this environment because `just` is not installed.

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -897,6 +897,54 @@ impl<'a> PerlLexer<'a> {
         None
     }
 
+    /// Consume a balanced segment inside a double-quoted string interpolation.
+    ///
+    /// Unlike `consume_balanced_segment`, this variant stops before an unescaped
+    /// closing double-quote so interpolation-local delimiter mistakes (like
+    /// `$hash{incomplete"`) do not consume the string terminator and poison the
+    /// whole token as "unterminated string".
+    #[inline]
+    fn consume_balanced_segment_in_double_quote(
+        &mut self,
+        open: char,
+        close: char,
+    ) -> Option<usize> {
+        if self.current_char() != Some(open) {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #6269, opened 2026-04-24
fix(parser): recover incomplete string interpolation indexing without ERROR nodes (#0000)

EffortlessSteven:
### Motivation

- Double-quoted string interpolations with incomplete hash/array subscripts (e.g. `"Key: $hash{incomplete"`, `"Element: $array[0"`) could cause the lexer to swallow the closing quote and produce an unterminated-string error that poisoned top-level parsing and prevented variable-symbol extraction.
- Improve editor/typing UX by making interpolation-local failures recover without creating top-level `ERROR` nodes and by preserving the base variable parts for symbol extraction.

### Description

- Add a recovery-aware lexer helper `consume_balanced_segment_in_dq_interpolation` in `crates/perl-lexer/src/lib.rs` that stops at the enclosing `"` when a `{...` or `[...]` interpolation tail is unterminated, instead of consuming the string terminator.
- Use the new helper when scanning interpolation tails for method-call tails, array slices and hash-expression tails inside double-quoted string interpolation so incomplete tails don't turn the rest of the input into an unterminated-string `Error` token.
- Add focused regression tests in `crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs` covering incomplete hash-key, array-index, arrow-hash-deref, mixed incomplete interpolation, a complete-interpolation control case, and a lexer-level assertion that the base variable parts (`$hash`, `$array`) are preserved in incomplete cases.

### Testing

- Ran `cargo test -p perl-parser-core --test string_interpolation_incomplete_tests` and it passed (6 tests).
- Ran `cargo test -p perl-parser-core string_interpolation` and it passed for the string-interpolation test group.
- Ran `cargo test -p perl-parser-core`; an unrelated test (`test_deref_hash_subscript_regex_key`) failed on an initial run but the subsequent run completed successfully (full crate tests passed locally after re-run).
- Formatting applied with `cargo fmt`; `just parser-audit` and `just cpan-corpus-check` were not executed in this environment because `just` is not installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb53f391d88333863ff64d42e3d555)

PR #6270, opened 2026-04-24
fix(parser): recover incomplete string interpolation indexing (#0000)

EffortlessSteven:
### Motivation
- Incomplete indexing in double-quoted interpolations (e.g. `"Key: $hash{incomplete"`, `"Element: $array[0"`) currently lets the interpolation consume the closing quote and produce top-level `ERROR` nodes, killing downstream symbol extraction and LSP features while users type.

### Description
- Add a lexer-local recovery helper `consume_interpolation_segment_in_dq` in `crates/perl-lexer/src/lib.rs` that scans `[`/`{` interpolation tails but returns early if an unescaped `"` appears so the outer double-quoted string can terminate normally.
- Use the new helper when parsing interpolation tails for `->[...]`, `->{...}`, direct `[...]`, and direct `{...}` in double-quoted strings so incomplete subscripts are preserved as partial parts rather than forcing parser ERROR nodes.
- Add parser regression tests in `crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs` covering incomplete and complete cases (`$hash{incomplete`, `$array[0`, `$obj->{field`, `$array[$i` and baselines).
- Add semantic analyzer tests in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs` to assert that `$hash` and `$array` remain extractable as symbol references from incomplete interpolations.

### Testing
- Ran `cargo test -p perl-parser-core --test string_interpolation_incomplete_tests` and it passed (5 tests).
- Ran focused semantic analyzer tests `cargo test -p perl-semantic-analyzer incomplete_hash_index_interpolation_still_registers_reference` and `... incomplete_array_index_interpolation_still_registers_reference` and both passed.
- Ran `cargo test -p perl-parser-core` which completed but reported one unrelated failing test (`test_deref_hash_subscript_regex_key`) that preexisted/appeared outside the scoped interpolation change.
- `just parser-audit` and `just cpan-corpus-check` were not executed in this environment because `just` is not available.

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -897,6 +897,87 @@ impl<'a> PerlLexer<'a> {
         None
     }
 
+    /// Consume a balanced segment used by interpolation tails inside a double-quoted
+    /// string. Unlike `consume_balanced_segment`, this local recovery helper stops
+    /// before an unescaped `"` that appears before the closing delimiter so the
+    /// outer string parser can still terminate the string token cleanly.
+    fn consume_interpolation_segment_in_dq(&mut self, open: char, close: char) -> Option<usize> {
+        if self.current_char() != Some(open) {
+            return None;
+        }
+
+        let mut depth = 1usize;
+        self.advance();
+
+        while let Some(ch) = self.current_char() {
+            match ch {
+                '\\' => {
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #6271, opened 2026-04-24

EffortlessSteven:
### Motivation

- Bare unary-style calls (e.g. `is_ready $obj ? 1 : 0`) could greedily consume a following ternary or other low-precedence operator as their argument, producing incorrect ASTs.

### Description

- Treat `TokenKind::Question` as a terminator for bare/indirect call argument collection in `parse_indirect_call` so `?` stops indirect-call arg parsing.
- Treat `TokenKind::Question` as a statement/bare-call boundary in `should_continue_bare_call_after_block` and `is_at_statement_end` so low-precedence word operators and `?` bind outside bare calls.
- Add focused regression tests in `crates/perl-parser-core/tests/fix_bare_call_low_precedence_binding.rs` covering `is_ready $obj ? 1 : 0`, `do_thing @args or die`, `do_thing $x and return`, `transform $x // $fallback`, and a nested bare-call-with-ternary case.

### Testing

- `cargo test -p perl-parser-core ternary` ran and passed.
- `cargo test -p perl-parser-core indirect` ran and passed.
- The new focused test binary `cargo test -p perl-parser-core --test fix_bare_call_low_precedence_binding` ran and passed.
- `cargo test -p perl-parser-core` was run; the change is correct for the introduced regressions but one unrelated existing test (`test_deref_hash_subscript_regex_key`) failed in this environment and should be investigated separately.
- `cargo xtask fmt` completed successfully and `just`-based checks were not runnable in this environment (`just` not installed).

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -377,6 +377,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
+                        | TokenKind::Question
                 )
             )
         {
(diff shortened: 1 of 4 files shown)


EffortlessSteven:

PR #6272, opened 2026-04-24

EffortlessSteven:
### Motivation

- Bare unary-style calls with sigil-leading arguments could greedily absorb ternary and other low-precedence operators, e.g. `my $x = is_ready $obj ? 1 : 0;` should parse as `(is_ready $obj) ? 1 : 0` but could be mis-parsed as `is_ready($obj ? 1 : 0)`.

### Description

- Treat the ternary `?` token as a terminator for indirect/bare-call argument collection by adding `TokenKind::Question` to the terminator/match sets in `calls.rs` and `postfix.rs`.
- Extend statement-end detection to consider `?` as a statement boundary so low-precedence word operators stay outside bare-call parsing.
- Ensure bare-call sigil-peek path only captures the high-precedence argument so ternary and other low-precedence operators bind outside the call (logic already uses a high-precedence parse for the sigil-peek case and is preserved with the new terminator handling).
- Add focused regression tests in `crates/perl-parser-core/tests/fix_bare_call_low_precedence_binding.rs` covering `is_ready $obj ? 1 : 0`, `do_thing @args or die`, `do_thing $x and return`, `transform $x // $fallback`, and a nested bare-call-with-ternary case.

### Testing

- `cargo xtask fmt` was run and completed successfully.
- `cargo test -p perl-parser-core --test fix_bare_call_low_precedence_binding` passed (all tests in the new test file succeeded).
- `cargo test -p perl-parser-core ternary` passed with existing ternary-suite tests succeeding.
- `cargo test -p perl-parser-core indirect` passed with indirect-call tests succeeding.
- `cargo test -p perl-parser-core` was executed and reported one unrelated test failure (`test_deref_hash_subscript_regex_key` in `tests/test_edge_cases_deep.rs`), which appears independent of these changes.

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -374,6 +374,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::WordAnd
                         | TokenKind::WordXor
                         | TokenKind::WordNot
+                        | TokenKind::Question
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #6273, opened 2026-04-24

EffortlessSteven:
### Motivation
- Prevent bare unary-style calls from greedily absorbing low-precedence operators so expressions like `my $x = is_ready $obj ? 1 : 0;` parse as `(is_ready $obj) ? 1 : 0` rather than `is_ready($obj ? 1 : 0)`.

### Description
- Treat `TokenKind::Question` as a disqualifier/terminator for indirect/bare-call argument collection in `crates/perl-parser-core/src/engine/parser/expressions/calls.rs` so `?` no longer begins an indirect-object/argument list.
- Add `TokenKind::Question` to statement/bare-call boundary checks in `crates/perl-parser-core/src/engine/parser/expressions/postfix.rs` and `crates/perl-parser-core/src/engine/parser/helpers.rs` so bare-call flows stop before ternary and similar low-precedence operators.
- Ensure the sigil-peek bare-call path continues to parse only a high-precedence argument (`parse_shift()`), preserving the intended operator binding behavior outside the bare call.
- Add focused regression tests in `crates/perl-parser-core/tests/fix_bare_call_low_precedence_binding.rs` covering `is_ready $obj ? 1 : 0;`, `do_thing @args or die;`, `do_thing $x and return;`, `transform $x // $fallback;`, and a nested bare-call-with-ternary case.

### Testing
- Ran `cargo xtask fmt` successfully.
- Ran `cargo test -p perl-parser-core --test fix_bare_call_low_precedence_binding` and all tests passed.
- Ran `cargo test -p perl-parser-core ternary` and `cargo test -p perl-parser-core indirect` and both suites passed.
- Ran full `cargo test -p perl-parser-core`; the run completed but reported one unrelated pre-existing failing test `test_edge_cases_deep::test_deref_hash_subscript_regex_key` (not caused by these changes).
- Could not run `just parser-audit` or `just corpus-sweep-check` in this environment because `just` is not available.

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -91,6 +91,7 @@ impl<'a> Parser<'a> {
                 | TokenKind::RightBrace
                 | TokenKind::RightParen
                 | TokenKind::Comma
+                | TokenKind::Question
                 | TokenKind::Eof => return false,
                 _ => {}
             }
@@ -374,6 +375,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::WordAnd
                         | TokenKind::WordXor
                         | TokenKind::WordNot
+                        | TokenKind::Question
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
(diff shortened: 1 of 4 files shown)


EffortlessSteven:

PR #6274, opened 2026-04-24

EffortlessSteven:
### Motivation
- Prevent bare unary-style calls from greedily absorbing ternary (`?:`) and other low-precedence operators so expressions like `my $x = is_ready $obj ? 1 : 0;` parse as a bare call followed by a ternary expression.
- Ensure related patterns (`do_thing @args or die`, `do_thing $x and return`, `transform $x // $fallback`) bind the low-precedence operator outside the bare call.

### Description
- Treat `TokenKind::Question` (`?`) as a terminator/disqualifier when collecting indirect-call or bare-call arguments so `?` does not get consumed into the call argument list; updated checks in `calls.rs`, `postfix.rs`, and `helpers.rs`.
- In the sigil-peek bare-call path keep parsing only a high-precedence argument via `parse_shift()` so lower-precedence operators (ternary/word-ops/`//`) remain outside the call (change is enforced by preserving the high-precedence parse for the single sigil argument).
- Add focused regression tests in `crates/perl-parser-core/tests/named_unary_bare_call_binding_tests.rs` that assert AST shapes for the targeted cases rather than only clean-parse, covering `is_ready $obj ? 1 : 0`, `do_thing @args or die`, `do_thing $x and return`, `transform $x // $fallback`, and a nested bare-call-with-ternary inside a larger expression.

### Testing
- Ran `cargo test -p perl-parser-core --test named_unary_bare_call_binding_tests` and the new focused tests passed.
- Ran `cargo test -p perl-parser-core ternary` and `cargo test -p perl-parser-core indirect` and both suites passed.
- Ran full `cargo test -p perl-parser-core`; most tests passed, but an existing unrelated test (`test_edge_cases_deep::test_deref_hash_subscript_regex_key`) failed in this environment (pre-existing failure surfaced during the run).
- Ran `cargo fmt`; `just parser-audit` and `just corpus-sweep-check` were not available in this environment (`just: command not found`).

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -374,6 +374,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::WordAnd
                         | TokenKind::WordXor
                         | TokenKind::WordNot
+                        | TokenKind::Question
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
--- a/crates/perl-parser-core/src/engine/parser/expressions/postfix.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/postfix.rs
@@ -1198,6 +1198,7 @@ impl<'a> Parser<'a> {
                 | Some(TokenKind::WordOr)
                 | Some(TokenKind::WordXor)
                 | Some(TokenKind::WordNot)
+                | Some(TokenKind::Question)
                 | Some(TokenKind::DataMarker)
                 | Some(TokenKind::Eof)
                 | None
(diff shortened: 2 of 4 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6282 implements the same bare-call/ternary binding fix with more comprehensive test coverage (7 tests vs 4 in this variant).

Both fix the same issue in helpers.rs, postfix.rs, and calls.rs with identical core changes. #6282 adds edge cases for:
- Block-list functions (grep) absorbing ternary arguments
- Nested ternary expressions
- Direct block expressions before ternary

**Extracted and noted for #6282 consideration**: Your assertion helpers (first_non_sub_stmt, assert_call_name) are clean and more maintainable than #6282's match-heavy style — recommend adopting those patterns in #6282's final iteration.

Thank you — the comparison revealed better test organization patterns.

PR #6276, opened 2026-04-24
ci(parser): stabilize CPAN corpus ratchet for accuracy closeout (#0000)

EffortlessSteven:
### Motivation
- Batch CPAN installs were timing out at 120s for 50-distribution batches, causing expensive per-distribution retries and noisy nightly ratchet signals that undermine accuracy closeout.

### Description
- Increase the batch cpanm hard timeout from `Duration::from_secs(120)` to `Duration::from_secs(300)` and reduce the install batch size from `50` to `25` in `xtask/src/tasks/cpan_corpus.rs`, leaving ratchet, cache, and known-clean manifest behavior unchanged.

### Testing
- `cargo build -p xtask` succeeded; `cargo xtask cpan-corpus install --reset` (observed run) printed `Batch 1/40: installing 25 distributions...`, confirming the new batch size; `just cpan-corpus-check` was not available in this environment so the equivalent `cargo run -p xtask -- cpan-corpus sweep --enforce` was executed and completed but reported a ratchet regression against the committed baseline due to a partial local corpus rather than this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb55aaa4c083338bf587cc643739bf)

### DIFF
--- a/xtask/src/tasks/cpan_corpus.rs
+++ b/xtask/src/tasks/cpan_corpus.rs
@@ -36,7 +36,7 @@ const CPANM_STANDALONE_URL: &str = "https://cpanmin.us";
 const METACPAN_API: &str = "https://fastapi.metacpan.org/v1/distribution/_search";
 /// Hard timeout for a batch cpanm invocation. Batch installs fall back to
 /// per-distribution retries when one distribution wedges inside configure/build.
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6278 is the keeper (adds CPANM_BATCH_SIZE named constant + regression test guard).

PR #6279, opened 2026-04-24

EffortlessSteven:
### Motivation

- Prevent bare unary-style / bare-call forms from greedily absorbing ternary (`? :`) and other low-precedence follow-up operators so constructs like `my $x = is_ready $obj ? 1 : 0;` parse with the ternary outside the call.

### Description

- Stop collecting indirect/bare-call arguments when a `?` token appears by adding `TokenKind::Question` to the call/indirect-call terminator checks in `crates/perl-parser-core/src/engine/parser/expressions/calls.rs`.
- Extend the same `?`-as-terminator logic and statement-end detection in `crates/perl-parser-core/src/engine/parser/expressions/postfix.rs` so `or`, `and`, `//`, and other low-precedence operators bind outside bare calls and print/send implicit-list forms.
- Add focused regression tests in `crates/perl-parser-core/tests/named_unary_low_precedence_binding_tests.rs` exercising `is_ready $obj ? 1 : 0`, `do_thing @args or die`, `do_thing $x and return`, `transform $x // $fallback`, and a nested-case inside a larger expression.
- Changed files: `calls.rs`, `postfix.rs`, and added the new test file under `tests/`.

### Testing

- Ran targeted tests `cargo test -p perl-parser-core ternary`, `cargo test -p perl-parser-core indirect`, and `cargo test -p perl-parser-core named_unary_low_precedence_binding` and they passed.
- Ran full crate tests with `cargo test -p perl-parser-core` which exposed a pre-existing unrelated failure (`test_deref_hash_subscript_regex_key` in `tests/test_edge_cases_deep.rs`) and is not caused by these changes.
- `just parser-audit` and `just corpus-sweep-check` were not executed in this environment because `just` is not available.

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -374,6 +374,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::WordAnd
                         | TokenKind::WordXor
                         | TokenKind::WordNot
+                        | TokenKind::Question
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #6280, opened 2026-04-24

EffortlessSteven:
### Motivation

- Bare unary-style/bare-call patterns were greedily absorbing low-precedence operators (ternary `? :`, `and`, `or`, `//`, etc.), causing those operators to bind inside the bare call instead of outside it.

### Description

- Treat `TokenKind::Question` as a call/argument terminator in indirect-call argument collection so `?` stops argument parsing for indirect/bare calls (`crates/perl-parser-core/src/engine/parser/expressions/calls.rs`).
- Treat `TokenKind::Question` as a statement-end in postfix parsing so postfix/bare-call detection stops before a ternary continuation (`crates/perl-parser-core/src/engine/parser/expressions/postfix.rs`).
- Add focused regression tests that assert ternary and other low-precedence operators remain outside bare calls, covering `is_ready $obj ? 1 : 0`, `do_thing @args or die`, `do_thing $x and return`, `transform $x // $fallback`, and a nested case (`crates/perl-parser-core/tests/bare_call_binding_regression_tests.rs`).

### Testing

- Ran `cargo test -p perl-parser-core --test bare_call_binding_regression_tests` and the new regression tests passed.
- Ran `cargo test -p perl-parser-core ternary` and `cargo test -p perl-parser-core indirect` and both passed.
- Ran `cargo test -p perl-parser-core` which exercises the whole crate and surfaced an unrelated existing failure in `test_deref_hash_subscript_regex_key`; this failure appears unrelated to the changes in this PR.

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -374,6 +374,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::WordAnd
                         | TokenKind::WordXor
                         | TokenKind::WordNot
+                        | TokenKind::Question
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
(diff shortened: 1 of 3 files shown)


EffortlessSteven:

PR #6281, opened 2026-04-24

EffortlessSteven:
### Motivation

- Bare unary-style/bare-call parsing was greedily absorbing low-precedence follow-up operators (ternary `? :`, `or`, `and`, `//`), causing expressions like `is_ready $obj ? 1 : 0` to be parsed incorrectly.

### Description

- Stop argument-collection for indirect/bare calls at the ternary `?` token by adding `TokenKind::Question` to the terminator checks in `parse_indirect_call` (`crates/perl-parser-core/src/engine/parser/expressions/calls.rs`).
- Treat `?` as a statement/bare-call boundary in the bare-call continuation and builtin/indirect-object paths by updating checks in `postfix.rs` (`crates/perl-parser-core/src/engine/parser/expressions/postfix.rs`) and `helpers.rs` (`crates/perl-parser-core/src/engine/parser/helpers.rs`).
- Add focused regression tests in `crates/perl-parser-core/tests/bare_call_binding_regression_tests.rs` covering: `is_ready $obj ? 1 : 0;`, `do_thing @args or die;`, `do_thing $x and return;`, `transform $x // $fallback;`, and a nested bare-call-with-ternary case.

### Testing

- Ran `cargo test -p perl-parser-core --test bare_call_binding_regression_tests`, which passed.
- Ran `cargo test -p perl-parser-core ternary` and `cargo test -p perl-parser-core indirect`, both passed.
- Ran `cargo test -p perl-parser-core` but observed one unrelated existing failure in `test_edge_cases_deep::test_deref_hash_subscript_regex_key` (parsing `my $x = ${$ref}{m};`) that predates and is outside the scope of this change.
- Ran `cargo xtask fmt` successfully; `just parser-audit` and `just corpus-sweep-check` were not available in this environment (`just: command not found`).

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/calls.rs
@@ -377,6 +377,7 @@ impl<'a> Parser<'a> {
                         | TokenKind::RightBrace
                         | TokenKind::RightParen
                         | TokenKind::RightBracket
+                        | TokenKind::Question
                 )
             )
         {
--- a/crates/perl-parser-core/src/engine/parser/expressions/postfix.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/postfix.rs
@@ -551,6 +551,7 @@ impl<'a> Parser<'a> {
                                                     | TokenKind::WordAnd
                                                     | TokenKind::WordXor
                                                     | TokenKind::WordNot
+                                                    | TokenKind::Question
                                             )
                                         )
                                     {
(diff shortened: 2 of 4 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6282 implements the same scope with superior test design (semantic AST checks vs S-expr string matching) and cleaner implementation (consolidated token checks). Both PRs correctly fix ternary binding after bare calls; #6282's 6 tests provide better maintainability. Thank you for the contribution.

PR #6283, opened 2026-04-24

EffortlessSteven:
### Motivation

- While users type, double-quoted strings frequently contain incomplete interpolation tails like `$hash{incomplete` or `$array[0` that should not produce top-level `ERROR` nodes and should preserve extractable variable references.
- The lexer was consuming balanced segment scans past the closing `"`, which could collapse interpolation-local incompleteness into a catastrophic token/parse error.

### Description

- Add `consume_balanced_segment_with_terminator` to the lexer to allow balanced-segment scanning to stop early when encountering a caller-specified terminator (used to avoid consuming the surrounding `"`).
- Use the new terminator-aware routine in double-quoted interpolation handling for `${...}`, `$var[...]`, `$var{...}`, and `->` tails so incomplete `StringPart` tails like `"{incomplete"`, `"[0"`, `"-> {field"`, and `"[$i"` are preserved as partial `StringPart` entries instead of forcing an `ERROR` token (file: `crates/perl-lexer/src/lib.rs`).
- Add lexer unit-test cases and parser clean-parse regression tests to verify incomplete and complete interpolation behavior (`crates/perl-lexer/tests/comprehensive_unit_tests.rs` and new `crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs`).

### Testing

- Ran `cargo test -p perl-lexer interpolated_string_preserves_complex_tails` and it passed.
- Ran `cargo test -p perl-parser-core --test string_interpolation_incomplete_tests` and it passed.
- Ran `cargo test -p perl-parser-core string_interpolation` and it passed.
- Ran the full `cargo test -p perl-parser-core` in this environment and observed an unrelated existing test failure (`test_deref_hash_subscript_regex_key`) that pre-existed and is outside the scope of this change; the interpolation-specific tests introduced here passed.
- `just parser-audit` and `just cpan-corpus-check` were not executed because `just` is not available in this environment (error: `just: command not found`).

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -864,7 +864,12 @@ impl<'a> PerlLexer<'a> {
     }
 
     #[inline]
-    fn consume_balanced_segment(&mut self, open: char, close: char) -> Option<usize> {
+    fn consume_balanced_segment_with_terminator(
+        &mut self,
+        open: char,
+        close: char,
+        terminator: Option<char>,
+    ) -> Option<usize> {
         if self.current_char() != Some(open) {
             return None;
         }
@@ -872,6 +877,11 @@ impl<'a> PerlLexer<'a> {
         let mut depth = 1usize;
         self.advance();
         while let Some(ch) = self.current_char() {
+            if terminator == Some(ch) {
+                // Recovery: stop before a known outer delimiter so callers can continue.
+                return None;
+            }
+
             match ch {
                 '\\' => {
                     self.advance();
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6287 implements the same scope more completely (includes both lexer tests and parser.rs primary changes that #6283 lacks). Both PRs fix incomplete string interpolation indexing recovery. Extracted from your PR: comprehensive lexer-unit test coverage pattern that complements #6287's test suite. Thank you.

PR #6284, opened 2026-04-24

EffortlessSteven:
### Motivation
- While users are typing, incomplete indexing inside double-quoted interpolations (e.g. `$hash{incomplete`, `$array[0`) caused the lexer to consume through the closing quote and the parser to produce top-level ERROR nodes, degrading IDE features and symbol extraction.

### Description
- Add `consume_balanced_segment_until_terminator` in the lexer to stop balanced scanning for `[...]` / `{...}` when an unescaped string terminator (`"`) is encountered so the surrounding string token is preserved (`crates/perl-lexer/src/lib.rs`).
- Use the lexer recovery when parsing interpolation tails (array/hash/arrow subscripts/method tails) so incomplete subscripts do not swallow the closing quote and remain as string parts (`crates/perl-lexer/src/lib.rs`).
- Add a parser-side helper `record_incomplete_interpolation_indexing` that emits a diagnostic for incomplete interpolation indexing without producing a propagated `ERROR` node and attach it when consuming double-quoted string tokens (`crates/perl-parser-core/src/engine/parser/expressions/primary.rs`).
- Add focused regression tests that assert no top-level `ERROR` nodes, preservation of interpolated string content, and that diagnostics are recorded for incomplete cases while complete interpolations remain clean (`crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs`).

### Testing
- Ran `cargo test -p perl-parser-core --test string_interpolation_incomplete_tests` and all new tests passed (5/5).
- Ran `cargo test -p perl-parser-core string_interpolation` and it completed successfully.
- Ran `cargo test -p perl-parser-core` which exposes an existing unrelated failing test (`test_edge_cases_deep::test_deref_hash_subscript_regex_key`) that predates this change and remains failing; this failure is unrelated to the interpolation recovery.
- Ran `cargo xtask fmt` to format code; `just parser-audit` / `just cpan-corpus-check` were not run because `just` is not available in this environment.

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -897,6 +897,56 @@ impl<'a> PerlLexer<'a> {
         None
     }
 
+    /// Like `consume_balanced_segment`, but stops before an unescaped string terminator.
+    ///
+    /// This is used for interpolation tails inside double-quoted strings (e.g. `$h{key}`).
+    /// If the interpolation subscript/index is incomplete (missing `}` or `]`) and a closing
+    /// `"` is reached, we must *not* consume through that quote, otherwise the entire string
+    /// is misclassified as unterminated.
+    #[inline]
+    fn consume_balanced_segment_until_terminator(
+        &mut self,
+        open: char,
+        close: char,
+        terminator: char,
+    ) -> Option<usize> {
+        if self.current_char() != Some(open) {
+            return None;
+        }
+
+        let mut depth = 1usize;
+        self.advance();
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6287 implements the same scope more completely, including lexer test coverage. Both fix incomplete string interpolation recovery. Your PR's approach of using Result<> returns in tests is excellent Rust practice; recommend this pattern for follow-up hardening PRs. Thank you.

PR #6287, opened 2026-04-24
fix(parser): recover incomplete interpolation indexing in strings (#0000)

EffortlessSteven:
### Motivation

- Improve IDE/editing experience by preventing unterminated interpolation tails inside double-quoted strings from producing top-level ERROR nodes that poison the AST and block symbol extraction for the sigil target.

### Description

- Add a quote-aware lexer helper `consume_balanced_segment_in_string` and use it when scanning interpolation tails inside double-quoted strings so incomplete `{`/`[`/`->` tails stop at the closing quote and preserve the variable part as a token (`crates/perl-lexer/src/lib.rs`).
- Make the lexer emit `InterpolatedString` parts for incomplete tails (e.g. `StringPart::Expression("{incomplete")`, `StringPart::ArraySlice("[0")`) and extend lexer unit tests accordingly (`crates/perl-lexer/tests/comprehensive_unit_tests.rs`).
- Add parser-side, non-fatal diagnostics for unclosed interpolation delimiters detected inside string tokens while continuing to return a normal `String` AST node so the parse does not produce a top-level `ERROR` node (`crates/perl-parser-core/src/engine/parser/expressions/primary.rs`).
- Add focused parser tests that assert incomplete interpolation cases produce a clean AST (no ERROR nodes) and a diagnostic, plus a completeness case to ensure no regressions (`crates/perl-parser-core/tests/string_interpolation_incomplete_tests.rs`).

### Testing

- Ran `cargo test -p perl-parser-core --test string_interpolation_incomplete_tests` and it passed (all new parser tests OK).
- Ran `cargo test -p perl-lexer interpolated_string_preserves_complex_tails` and it passed (lexer tests OK).
- Ran `cargo test -p perl-parser-core string_interpolation` and it passed (filtered suite OK).
- Ran full `cargo test -p perl-parser-core` which revealed an unrelated existing regression in `test_edge_cases_deep::test_deref_hash_subscript_regex_key` (one failing test) that is not introduced by these changes.
- Ran `cargo fmt --all` successfully and left code formatted; `just parser-audit` / `just cpan-corpus-check` were not executed in this environment because `just` is not available.

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -863,6 +863,11 @@ impl<'a> PerlLexer<'a> {
         }
     }
 
+    /// General-purpose balanced-segment consumer (no quote-boundary recovery).
+    ///
+    /// For use inside double-quoted string interpolation where the outer `"` must
+    /// act as a recovery boundary, use [`consume_balanced_segment_in_string`] instead.
+    #[allow(dead_code)]
     #[inline]
     fn consume_balanced_segment(&mut self, open: char, close: char) -> Option<usize> {
         if self.current_char() != Some(open) {
@@ -2905,7 +2910,8 @@ impl<'a> PerlLexer<'a> {
                                             )));
                                         }
                                         Some('(') => {
-                                            let _ = self.consume_balanced_segment('(', ')');
+                                            let _ = self
+                                                .consume_balanced_segment_in_string('(', ')', '"');
                                             parts.push(StringPart::MethodCall(Arc::from(
                                                 &self.input[tail_start..self.position],
                                             )));
@@ -2930,7 +2936,9 @@ impl<'a> PerlLexer<'a> {
                                                 }
                                             }
                                             if self.current_char() == Some('(') {
-                                                let _ = self.consume_balanced_segment('(', ')');
+                                                let _ = self.consume_balanced_segment_in_string(
+                                                    '(', ')', '"',
+                                                );
                                             }
                                             parts.push(StringPart::MethodCall(Arc::from(
                                                 &self.input[tail_start..self.position],
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -897,6 +897,51 @@ impl<'a> PerlLexer<'a> {
         None
     }
 
+    #[inline]
+    fn consume_balanced_segment_in_string(
+        &mut self,
+        open: char,
+        close: char,
+        terminator: char,
+    ) -> Option<usize> {
+        if self.current_char() != Some(open) {
+            return None;
+        }
+
+        let mut depth = 1usize;
+        self.advance();
+        while let Some(ch) = self.current_char() {
+            match ch {
+                '\\' => {
+                    self.advance();
+                    if self.current_char().is_some() {
+                        self.advance();
+                    }
+                }
+                c if c == terminator => {
+                    // Local recovery for interpolation tails in quoted strings:
+                    // stop at the closing quote so the outer string parser can
+                    // still terminate this token cleanly.
+                    return None;
+                }
+                c if c == open => {
+                    depth += 1;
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
## Standards Review (Complete) ✓

### Scope
Tight + focused. Four files (lexer + parser-core). Actual commit c74fdf95e touches only the interpolation recovery as described. Branch contains 24 stale prior commits (merge-base staleness, not scope drift) — these should be rebased out before merge.

### Banned Patterns
- **Lexer helper** (`consume_balanced_segment_in_string`): Safe byte-level scanning, no unwrap/expect/panic/todo/dbg ✓
- **Parser helpers**: `find_unclosed_interpolation_delimiter` + `consume_balanced_in_interpolated_string` use saturating arithmetic and safe pattern matching ✓
- **Test file**: All 5 tests use `Result<(), Box<dyn std::error::Error>>` with `?` operator, no bare assert!/unwrap/expect ✓

### Code Quality
- Lexer recovery is quote-aware (terminator parameter prevents early stop at outer quote) — solves root problem
- Parser diagnostics non-fatal: returns valid `String` node + records diagnostic
- Test coverage: 4 incomplete cases + 1 regression case, clean assertion hygiene
- No code duplication

### Minor Observation
Diagnostic message includes delimiter name ('{' vs '['), but tests only verify presence of Unclosed + interpolated, not specific delimiter. Non-blocking (error recovery works, diagnostic is clear), but a nice-to-have for future would be a test asserting the exact delimiter appears in the message.

**Review-reviewed ✓** | Routing: needs-deep-review

EffortlessSteven:
## Deep Review Complete

**Logic: FIXED (fix-forward commit 85b09b961)**

### Findings fixed directly on branch

**1. Paren-tail recovery gap (missing call-site update)**
Three `consume_balanced_segment` call sites inside the `->` dispatch (two `'('` variants and the `method(` identifier-then-paren form) were not updated to `consume_balanced_segment_in_string`. This meant `"$obj->method(arg"` and `"$obj->(incomplete"` still swallowed the closing quote and produced a poisoned AST. All three are now updated.

**2. Parser-side diagnostic gap for paren tails**
`find_unclosed_interpolation_delimiter` handled `->{ }` and `->[ ]` but not `->method(` or `->(`. Extended to detect unclosed `(` in both `$obj->method(incomplete` and `$obj->(incomplete` patterns. The new test `double_quote_incomplete_arrow_paren_call` exposed this and now passes.

**3. Missing test coverage**
- Added lexer unit tests for `"$obj->method(arg"` (method paren tail) and `"$obj->(incomplete"` (coderef paren tail).
- Added parser test `double_quote_incomplete_block_deref` for `"${incomplete"` (block-dereference form) — this path uses the `Some('{')` branch in the lexer, which was updated in the PR but had no dedicated test.

**4. Dead-code lint**
`consume_balanced_segment` is now unused at all call sites; added `#[allow(dead_code)]` with a doc comment explaining the relationship to `consume_balanced_segment_in_string`.

### Relationship to PR #6269

**Both PRs are open** — the review request incorrectly stated #6269 was merged (`gh pr view 6269` shows `"state":"OPEN","mergedAt":null`). They address the same root problem with different designs:
- #6269: `consume_balanced_segment_in_dq_interpolation` (hardcoded `"` boundary, depth-1-only guard, covers `(` call sites)
- #6287: `consume_balanced_segment_in_string(open, close, terminator)` (parameterized, any-depth terminator check) + parser-side diagnostic layer

These will conflict on merge. The orchestrator should evaluate which to take forward — #6287 is now more complete after this fix and adds the parser-layer diagnostic that #6269 lacks.

### Verdict

Logic: CORRECT after fix | Tests: 7/7 parser + 1540 lexer passing | Regression risk: LOW (pre-existing `test_deref_hash_subscript_regex_key` failure is unrelated, present on master) | Research verification: SKIPPED (no external claims)

`deep-reviewed`

PR #6306, opened 2026-04-24
fix(parser-core): accept mixed quote-like replacement delimiters and sync modifiers (#0000)

EffortlessSteven:
### Motivation

- The strict substitution/transliteration parsing rejected valid Perl forms where a paired-pattern is followed by an unpaired replacement delimiter (e.g. `s{pat}/repl/`).
- Transliteration (`tr`/`y`) and other quote-like paths had the same gap and could misreport or fail on valid idioms.
- The substitution modifier diagnostic text was narrower than the validator’s accepted set and needed to be kept in sync.

### Description

- Relax `extract_substitution_parts_strict` to allow an unpaired replacement delimiter after a paired pattern (accept `s{pat}/repl/`, `s{pat}|repl|`, etc.) while preserving strict error paths for malformed forms.
- Apply the same mixed-delimiter acceptance to `extract_transliteration_parts_strict` so `tr`/`y` with paired search and unpaired replacement delimiters parses correctly.
- Add `is_valid_quote_delimiter` helper used by strict parsing to centralize what constitutes a valid non-alphanumeric, non-whitespace delimiter.
- Update substitution invalid-modifier diagnostic text in the parser host (`expressions/primary.rs`) to include the full accepted modifier set and add focused regression tests that cover mixed-delimiter forms and invalid-modifier guardrails.

### Testing

- Ran `cargo test -p perl-parser-core --test fix_subst_unusual_delimiters_2732`, which passed (new mixed-delimiter tests and guards OK).
- Ran `cargo test -p perl-parser-core quote`, `cargo test -p perl-parser-core slash_ambiguity`, and `cargo test -p perl-parser-core heredoc`, which all passed.
- `cargo test -p perl-parser-core regex` reported a failing test (`test_edge_cases_deep::test_deref_hash_subscript_regex_key`) that appears to be a pre-existing edge-case unrelated to these changes; it did not fail due to the mixed-delimiter work.
- `just parser-audit`, `just corpus-sweep-check`, and `just cpan-corpus-check` were not run in this environment because `just` is not available here (`just: command not found`).

------

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/primary.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/primary.rs
@@ -227,7 +227,7 @@ impl<'a> Parser<'a> {
                             let message = match e {
                                 quote_parser::SubstitutionError::InvalidModifier(c) => {
                                     format!(
-                                        "Invalid substitution modifier '{}'. Valid modifiers are: g, i, m, s, x, o, e, r",
+                                        "Invalid substitution modifier '{}'. Valid modifiers are: g, i, m, s, x, o, e, r, a, d, l, u, n, p, c",
                                         c
                                     )
                                 }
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6304 implements the same fix. Both address the same parser accuracy issue in this wave. Thank you.

PR #6307, opened 2026-04-24
fix(parser): disambiguate quote-like hash key after deref (#0000)

EffortlessSteven:
### Motivation
- Prevent valid Perl `${$ref}{m}` from being tokenized as a malformed quote-like operator (`m};`) which caused spurious parse errors and error-bucket noise.

### Description
- Add a lexer ambiguity guard in `crates/perl-lexer/src/lib.rs` that detects `{m}`-style sequences immediately following `{` and prevents the lexer from treating `m}` as a quote-like delimiter start by checking the previous non-space character (`previous_nonspace_before`).
- Add a token-stream regression test `tokenizes_scalar_deref_then_hash_subscript` in `crates/perl-parser-core/tests/token_stream_tests.rs` asserting the expected token sequence for `${$ref}{m};`.

### Testing
- Ran `cargo test -p perl-parser-core --test token_stream_tests tokenizes_scalar_deref_then_hash_subscript` which passed.
- Ran `cargo test -p perl-parser-core --test test_edge_cases_deep test_deref_hash_subscript_regex_key` which passed.
- Ran the relevant parser suites: `cargo test -p perl-parser-core regex`, `cargo test -p perl-parser-core quote`, `cargo test -p perl-parser-core heredoc`, and `cargo test -p perl-parser-core slash_ambiguity`; these suites passed locally.
- `just parser-audit`, `just corpus-sweep-check`, and `just cpan-corpus-check` were not executed because `just` is not available in this environment.

------

### DIFF
--- a/crates/perl-lexer/src/lib.rs
+++ b/crates/perl-lexer/src/lib.rs
@@ -1834,6 +1834,15 @@ impl<'a> PerlLexer<'a> {
         (None, None)
     }
 
+    /// Return the previous non-whitespace character before `index`.
+    fn previous_nonspace_before(&self, index: usize) -> Option<char> {
+        self.input
+            .get(..index)?
+            .chars()
+            .rev()
+            .find(|c| !c.is_whitespace())
+    }
+
     /// Is `c` a valid quote-like delimiter? (non-alnum, including paired)
     fn is_quote_delim(c: char) -> bool {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6304 implements the same fix. Both address the same parser accuracy issue in this wave. Thank you.

PR #6313, opened 2026-04-24
fix(ast): enforce traversal conformance for child-bearing parser nodes (#6485)

EffortlessSteven:
### Motivation
- Downstream AST consumers used bespoke traversal logic that could silently skip child-bearing `NodeKind`s (notably `StatementModifier`), causing diagnostics, highlights, and semantic tokens to miss subtrees.

### Description
- Replace bespoke per-variant recursion in the diagnostics walker with the canonical `node.children()` traversal so every child-bearing `NodeKind` is visited consistently (updates `crates/perl-lsp-rs-core/src/providers/diagnostics/walker.rs`).
- Align the semantic-token walker to the same canonical traversal by delegating to `node.children()` recursion (updates `crates/perl-lsp-rs-core/src/providers/semantic_tokens/semantic_tokens.rs`).
- Make document-highlights fall back to `node.children()` for unhandled kinds to avoid future silent skips while keeping existing special cases (updates `crates/perl-lsp-rs-core/src/providers/document_highlight/mod.rs`).
- Add traversal-conformance tests: table-driven child-bearing samples and targeted tests asserting `print "ok" if $x = 5` visits both the `print` statement and the assignment condition, plus tests covering all six modifiers and nested modifiers (tests added to `diagnostics::walker` and `semantic_tokens` test modules).

### Testing
- Ran targeted provider tests in `perl-lsp-rs-core` and the new traversal tests passed: `cargo test -p perl-lsp-rs-core walk_ast_full_matches_canonical_ast_children` and `cargo test -p perl-lsp-rs-core statement_modifier_traversal_covers_statement_and_condition` both succeeded.
- Full `cargo test -p perl-lsp-rs-core` completed but encountered an unrelated failing test in the suite (`green_tdd_wave_g1b_regression_hardening::test_perl_lsp_cargo_toml_removed_g1b_imports`) that is not caused by these changes.
- `cargo test -p perl-parser-core` was run and exhibited an unrelated existing failure (`test_edge_cases_deep::test_deref_hash_subscript_regex_key`) which predates and is not introduced by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb562b964083338dcacb0b7065e784)

PR #6320, opened 2026-04-24

EffortlessSteven:
### Motivation
- Downstream consumers were using ad-hoc, hand-maintained child traversals and could silently skip child-bearing `NodeKind`s (notably `StatementModifier`), causing diagnostics, tokens, highlights, and reference features to miss subtrees.
- Use the canonical AST `children()` contract so every consumer walks the same set of children and avoids per-site omissions.

### Description
- Replace the diagnostics walker implementation with a single canonical traversal using `node.children()` in `crates/perl-lsp-rs-core/src/providers/diagnostics/walker.rs` and remove the long hand-maintained `match` over `NodeKind` variants.
- Add unit tests in `diagnostics/walker.rs` that (a) assert all six statement modifiers (`if`, `unless`, `while`, `until`, `for`, `foreach`) visit both statement and condition, (b) verify nested modifiers recurse, and (c) perform a table-driven conformance check comparing `walk_node` output against `Node::for_each_child` for every child-bearing `NodeKind`.
- Standardize consumer patterns that previously duplicated traversal logic by switching the semantic token walker `walk_ast_full` to iterate `node.children()` and by defaulting `document_highlight::get_children` to `node.children()` (retaining special-case signature flattening for `Subroutine`/`Method`).

### Testing
- Ran formatting: `cargo fmt --all` — success. 
- Ran targeted diagnostics walker tests: `cargo test -p perl-lsp-rs-core providers::diagnostics::walker::tests:: -- --nocapture` — all added walker tests passed. 
- Ran focused provider tests: `cargo test -p perl-lsp-rs-core --test assignment_in_condition_tests` and `cargo test -p perl-parser-core --test cpan_statement_modifiers` — both passed. 
- Full crate test runs: `cargo test -p perl-lsp-rs-core` and `cargo test -p perl-parser-core` were executed; the failures observed are due to unrelated pre-existing tests in this workspace snapshot (`test_perl_lsp_cargo_toml_removed_g1b_imports` and `test_deref_hash_subscript_regex_key`) and are not caused by the walker changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb562b964083338dcacb0b7065e784)


EffortlessSteven:
Closing as REDUNDANT — #6313 is the primary PR for this accuracy closeout focus area. Both address the same parser accuracy issue. Thank you.

PR #6321, opened 2026-04-24

EffortlessSteven:
### Motivation
- Downstream consumers (diagnostics, semantic tokens, highlights, symbol extraction) diverged in how they traversed the AST, which allowed child-bearing nodes like `StatementModifier` to be partially skipped and produced incorrect results.
- Enforce a single canonical traversal contract so every child-bearing `NodeKind` is visited consistently across providers.

### Description
- Replace manual `match NodeKind` recursion in the diagnostics walker with the canonical `for_each_child` traversal and add focused unit tests that confirm `StatementModifier` visits both `statement` and `condition` and that nested modifiers for `if/unless/while/until/for/foreach` traverse completely.
- Switch the semantic tokens collector to recurse via the canonical `node.children()` helper so it descends all child-bearing node kinds uniformly instead of a separate manual match.
- Simplify document-highlights child extraction to use `node.children()` by default while preserving the existing special-case flattening for subroutine/method `signature` parameters.
- Change the symbol extractor fallback to descend `node.children()` for unhandled node kinds so new child-bearing node kinds are not silently skipped, and add a table-driven conformance test (`crates/perl-parser-core/tests/ast_children_conformance_tests.rs`) that asserts expected direct-child counts for all child-bearing `NodeKind`s (including all six statement modifiers).

### Testing
- Ran formatting and static checks: `cargo xtask fmt` and `cargo fmt --all` succeeded and `cargo check --all-targets -p perl-lsp-rs-core && cargo check --all-targets -p perl-parser-core` succeeded.
- Unit tests covering the change passed: `cargo test -p perl-lsp-rs-core` targeted tests `walks_statement_modifier_statement_and_condition` and `walks_nested_statement_modifiers_for_all_supported_modifiers` passed, and `cargo test -p perl-parser-core --test ast_children_conformance_tests` passed.
- Exercised provider smoke tests: `cargo test -p perl-lsp-rs-core --test diag_snap` and `cargo test -p perl-lsp-rs-core --test assignment_in_condition_tests` passed.
- Noted unrelated pre-existing failures in full suite runs: a workspace-specific missing-path test (`test_perl_lsp_cargo_toml_removed_g1b_imports`) and a parser edge-case (`test_deref_hash_subscript_regex_key`) remained failing in full `cargo test` runs and are unrelated to the traversal changes above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb562b964083338dcacb0b7065e784)


EffortlessSteven:
Closing as REDUNDANT — #6313 is the primary PR for this accuracy closeout focus area. Both address the same parser accuracy issue. Thank you.

PR #6328, opened 2026-04-24
perf(perl-pragma): add benchmark suite for build/query workloads (#0000)

EffortlessSteven:
### Motivation

- The `perl-pragma` crate had no benchmark surface to measure the cost of building the pragma map and servicing point queries, leaving optimizer work without objective measurements.

### Description

- Add a Criterion benchmark suite at `crates/perl-pragma/benches/pragma_benchmarks.rs` exposing stable benchmark names: `build_small_file`, `build_large_file`, `query_random_offsets`, `query_monotonic_offsets`, `final_state_lookup`, `version_compat_walk_style`, and `scope_analyzer_walk_style`.
- Provide deterministic synthetic AST fixtures and a fallback realistic parsed fixture (via `demo_workspace/main.pl`) to exercise both build and query-heavy workloads and produce repeatable measurements.
- Wire up dev dependencies and a bench target in `crates/perl-pragma/Cargo.toml` (add `criterion` and `perl-parser` under `[dev-dependencies]` and `[[bench]]` entry) and update the lockfile to record the new bench deps.
- Document how to run the suite in `crates/perl-pragma/README.md` and replace usages of the deprecated `criterion::black_box` with `std::hint::black_box` in the bench code.

### Testing

- Executed `cargo bench -p perl-pragma` which completed and produced timing results for each named benchmark (per-iteration and total sample statistics).
- Ran `cargo xtask fmt` to ensure formatting; the task completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb777185b48333bed1dd367ff3c1ad)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1814,7 +1814,9 @@ dependencies = [
 name = "perl-pragma"
 version = "0.12.4"
 dependencies = [
+ "criterion",
  "perl-ast",
+ "perl-parser",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1816,7 +1816,6 @@ version = "0.12.4"
 dependencies = [
  "criterion",
  "perl-ast",
- "perl-parser",
 ]
 
 [[package]]
--- a/crates/perl-pragma/Cargo.toml
+++ b/crates/perl-pragma/Cargo.toml
@@ -18,7 +18,6 @@ perl-ast = { workspace = true }
 
 [dev-dependencies]
 criterion.workspace = true
-perl-parser = { workspace = true }
 
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/crates/perl-pragma/Cargo.toml
+++ b/crates/perl-pragma/Cargo.toml
@@ -16,8 +16,17 @@ categories = ["development-tools", "parser-implementations"]
 [dependencies]
 perl-ast = { workspace = true }
 
+[dev-dependencies]
+criterion.workspace = true
+perl-parser = { workspace = true }
+
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [lints]
 workspace = true
+
+
+[[bench]]
+name = "pragma_benchmarks"
+harness = false
--- a/crates/perl-pragma/README.md
+++ b/crates/perl-pragma/README.md
@@ -17,6 +17,26 @@ in the source.
   `Vec<(Range<usize>, PragmaState)>`, and offers `state_for_offset()` to
   query it.
 
+## Benchmarks
+
+Run the Criterion benchmark suite for build-heavy and query-heavy pragma workloads:
+
+```bash
+cargo bench -p perl-pragma
+```
+
+Benchmarks include stable names that can be diffed over time:
+
+- `build_small_file`
+- `build_large_file`
+- `query_random_offsets`
+- `query_monotonic_offsets`
+- `final_state_lookup`
+- `version_compat_walk_style`
+- `scope_analyzer_walk_style`
+
+Criterion reports per-benchmark timing statistics (including time per iteration and total sample estimates), which can be compared between runs.
+
 ## Workspace Role
 
 Tier 1 leaf crate. Depends only on `perl-ast`. Consumed by
--- a/crates/perl-pragma/benches/pragma_benchmarks.rs
+++ b/crates/perl-pragma/benches/pragma_benchmarks.rs
@@ -1,4 +1,4 @@
-use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
 use perl_ast::SourceLocation;
 use perl_ast::ast::{Node, NodeKind};
 use perl_pragma::PragmaTracker;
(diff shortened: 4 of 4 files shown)


EffortlessSteven:
**Deep review complete — fixes pushed directly.**

Two correctness issues found and fixed in commit 9e9068268.

**Fix 1 — version_compat_walk_style was benchmarking dead code.**
`demo_workspace/main.pl` has only `use strict; use warnings` — no version pragma, no `use feature`. Both `has_feature("say")` and `has_feature("builtin")` always returned false, so the score accumulator never incremented and the feature-lookup branch in `PragmaState::has_feature` was never exercised. Replaced `realistic_parsed_ast()` with a `synthetic_version_ast()` fixture that emits `use v5.36` (populates say, state, unicode_strings, signatures, defer, isa, etc.) and an explicit `use feature 'builtin'` — both has_feature checks now fire on most offsets, and the benchmark measures what it claims to measure.

**Fix 2 — scope_analyzer_walk_style Criterion stable name was wrong.**
`bench_with_input(BenchmarkId::from_parameter("dense_offsets"), ...)` inside a group named `scope_analyzer_walk_style` makes Criterion register the ID as `scope_analyzer_walk_style/dense_offsets`, not the bare name listed in the README. Switched to `bench_function("scope_analyzer_walk_style", ...)` — ID now matches documentation. This matters for `--bench` filter strings and HTML output comparisons.

**Side effects of fix:** `realistic_parsed_ast()` (now unused) removed; `perl-parser` dev-dependency removed (was only needed for that function); `BenchmarkId` import removed (now unused); Cargo.lock updated.

**Verified correct (unchanged):**
- All 7 stable benchmark names now match the README.
- `'signatures'` (single-quoted arg) correctly enables the feature — `pragma_arg_items` strips quotes via `trim_matches('\'')`.
- `map.last().map_or(..., |(range, _)| range.end)` destructures the `Vec<(Range<usize>, PragmaState)>` correctly.
- `final_state_lookup` querying `usize::MAX / 2` is valid and measures the binary-search fast path.
- `cargo check --all-targets -p perl-pragma` and `cargo clippy --all-targets -p perl-pragma` both clean.
- CI `compile_all_targets` (merge-gate, required) covers bench targets via `cargo check --workspace --all-targets`.

Logic: FIXED (version_compat dead-code loop; scope_analyzer ID mismatch)
Regression risk: LOW — bench file only, no production code touched
Research verification: SKIPPED (no external Perl/LSP/crate API claims)

EffortlessSteven:
## Reviewer Wrapup

**Summary:** reviewer-deep completed deep correctness review of PR #6328 (perl-pragma benchmark suite). 2 bugs fixed, 1 commit pushed, deep-reviewed label set.

---

**PR improved:** https://github.com/EffortlessMetrics/perl-lsp/pull/6328
**Changes made:** 2 fixes applied (3 files changed: pragma_benchmarks.rs, Cargo.toml, Cargo.lock)
**Test artifacts:** none (bench-only PR, no test files)
**Status:** approved — deep-reviewed label set
**Next step:** green-ci (verify CI passes on current HEAD 9e9068268)

---

**What was fixed:**

1. `version_compat_walk_style` exercised a loop where `has_feature("say")` and `has_feature("builtin")` always returned false — `demo_workspace/main.pl` has no `use v5.X` pragma so the feature set is always empty. Replaced `realistic_parsed_ast()` with `synthetic_version_ast()` which includes `use v5.36` + explicit `use feature 'builtin'`. Both checks now fire, and the `perl-parser` dev-dependency (only needed by the removed function) was dropped.

2. `scope_analyzer_walk_style` used `bench_with_input(BenchmarkId::from_parameter("dense_offsets"), ...)` inside a benchmark group, making Criterion's stable ID `scope_analyzer_walk_style/dense_offsets` — not the bare name in the README. Switched to `bench_function` so the ID matches.

**What was verified correct and left unchanged:**
- `'signatures'` (single-quoted) correctly enables the feature — `pragma_arg_items` strips quotes via `trim_matches('\'')`.
- `map.last().map_or(..., |(range, _)| range.end)` correctly destructures `Vec<(Range<usize>, PragmaState)>`.
- `final_state_lookup` querying `usize::MAX / 2` is a valid and useful binary-search probe.
- `std::hint::black_box` (not deprecated `criterion::black_box`) — correct.
- CI `compile_all_targets` covers bench targets via `cargo check --workspace --all-targets`.

---

**Retrospective:**

The key trap here was that `version_compat_walk_style` silently benchmarked a loop body where all branches were always false — it compiled, ran, and produced timing numbers, but those numbers measured the overhead of calling `state_for_offset` 512 times and discarding the result, not the cost of feature lookups. This class of bug (benchmark that measures near-nothing due to fixture not matching the feature) is invisible from compilation alone.

The `BenchmarkId` naming mismatch was easier to spot — the group + bench_with_input pattern is a known Criterion footgun where the stable ID grows a `/parameter` suffix that documentation-writers routinely forget.

Next time: for bench-only PRs, the first thing to check is whether each fixture actually exercises the code path the benchmark name claims to measure. Read the fixture data before reading the bench body.

---

- Logical next: green-ci on current HEAD `9e9068268`
- Related: the nightly CI `benchmarks` gate only runs `perl-parser` bench, not `perl-pragma`. A follow-up to wire `perl-pragma` into `just benchmarks` and the nightly gate would complete the "stable benchmark names for regression tracking" goal.
- Gotchas: `PragmaTracker::build` emits an entry for *every* pragma it sees, including `no warnings 'experimental'` — the map for `synthetic_version_ast` will have entries, so `map.last()` returns a non-None value and the `.max(512)` fallback is never exercised. That's fine — the fallback is only for empty-map safety.
- Confidence for next: high — bench compiles clean, two real bugs fixed, no production code touched.

EffortlessSteven:
**PR-responder investigation complete.**

Deep-reviewer already fixed both CI failures:

1. **version_compat_walk_style** — benchmarked dead code path. Fixture  had no version pragma, so  checks always returned false. Replaced with synthetic fixture emitting `use v5.36` and explicit `use feature 'builtin'` (commit 9e9068268).

2. **scope_analyzer_walk_style** — Criterion benchmark ID mismatch. Used `bench_with_input(BenchmarkId::from_parameter(...))` which made stable ID `scope_analyzer_walk_style/dense_offsets` instead of bare `scope_analyzer_walk_style` documented in README. Switched to `bench_function()` to match (same commit).

**Local verification:** `cargo check --workspace --all-targets` passes clean on commit 9e9068268. Re-triggered CI with `gh pr update-branch` — Compile All Targets and PR Smoke should now pass.

PR #6331, opened 2026-04-24
test(perl-pragma): expand regressions for conditional, builtin, feature-bundle, and lexical-scope edges (#0000)

EffortlessSteven:
### Motivation

- The pragma test surface lacked focused regressions for widened forms and obvious asymmetries around `strict` qw-lists, feature bundles, `builtin` imports/removals, conditional `if/unless` forms, `eval` behavior, and lexical restore semantics.

### Description

- Add a dedicated regression bank `crates/perl-pragma/tests/pragma_surface_regressions.rs` with 13 small fixture-style tests covering `use/no strict qw(...)`, `use/no feature ':all'`, `use/no builtin qw(...)`, conditional `no if/unless` targets, version-bundle + explicit feature toggles, nested `eval`/`eval STRING` behavior, and package/phase lexical restore.
- Implement `remove_builtin_imports` and wire it into `no builtin ...` and conditional `no if/unless ... builtin ...` handling so lexical builtin removals are tracked.
- Extend `apply_feature_state` so `use feature ':all'` when enabling populates the latest supported bundle surface (uses v5.40 bundle for `:all`).
- Tighten `strict`/`no strict` argument parsing to handle `qw(...)` and quoted lists via `pragma_arg_items` so selective toggles behave symmetrically with the widened surface.

### Testing

- Ran `cargo test -p perl-pragma` and all tests (existing + new regressions) passed.
- Ran `cargo check --all-targets -p perl-pragma` and it completed successfully.
- Ran formatting via `cargo xtask fmt` as part of verification and it completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb777808508333b9dbe001daa33a54)

### DIFF
--- a/crates/perl-pragma/src/lib.rs
+++ b/crates/perl-pragma/src/lib.rs
@@ -301,6 +301,13 @@ fn apply_feature_state(state: &mut PragmaState, args: &[String], enabled: bool) 
 
     for arg in args {
         for item in feature_items(arg) {
+            if enabled && item == ":all" {
+                for feature in features_enabled_by_version(PerlVersion::new(5, 40)) {
+                    changed |= enable_feature_name(state, feature);
+                }
+                continue;
+            }
+
             if !enabled && item == ":all" {
                 let had_features =
                     !state.features.is_empty() || state.unicode_strings || state.signatures_strict;
@@ -378,6 +385,17 @@ fn add_disabled_warning_category(state: &mut PragmaState, category: &str) {
     }
 
     state.disabled_warning_categories.push(category.to_string());
+}
+
+fn remove_builtin_imports(state: &mut PragmaState, args: &[String]) {
+    if args.is_empty() {
+        state.builtin_imports.clear();
+        return;
+    }
+
+    let names_to_remove: Vec<String> =
+        args.iter().flat_map(|arg| builtin_import_names(arg)).collect();
+    state.builtin_imports.retain(|import| !names_to_remove.iter().any(|name| name == import));
 }
 
 fn pragma_arg_items(arg: &str) -> Vec<String> {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
## Deep Review — PR #6331

**Verdict: Approved with one fix-forward improvement.**

### What I verified

**remove_builtin_imports behavior-preserving?** Yes. The new function correctly handles both the bare-args case (`no builtin;` → clear all via `state.builtin_imports.clear()`) and the selective-args case (retain imports not in the removal set). The `builtin_import_names` helper correctly parses both `qw(...)` forms and individual quoted/unquoted names.

**Feature bundle ":all" symmetry.** The `enabled && item == ":all"` and `!enabled && item == ":all"` branches are deliberately asymmetric — this is correct Perl semantics. `use feature ':all'` enables the v5.40 bundle surface (all current features). `no feature ':all'` clears everything currently in state. This matches Perl's behavior.

**Quote stripping before :all check.** `pragma_arg_items` strips outer quotes before matching, so `"':all'"` → `":all"` → matches the ":all" branch. Correct.

**Conditional-form builtin paths both have `return`.** The `no if/unless ... builtin` handler returns after pushing state. The `use if/unless ... builtin` handler also returns. No fall-through to the direct match arm.

**Tests pass.** 132/132 tests pass on the PR branch (5 suites including the new regression suite of 13 tests).

### Fix applied

**Missing test: bare `no builtin;` (empty args → clear all imports).** The PR tests selective removal and conditional removal, but not the no-arg form that clears everything. Added `no_builtin_bare_clears_all_imports` test — 133/133 pass after the addition. Pushed as commit `912d94dfbe7f4ecaae0c2875b6f8e45743b6b8b1`.

### Not a bug (but noted)

The `state_for_offset` function only uses `range.start` for partition, not `range.end`. This means tests only need `start <= offset` not `start <= offset < end`. The test offsets are all valid given this contract.

PR #6345, opened 2026-04-24
perf(pragma-consumers): migrate strict_warnings, version_compat, and scope_analyzer to PragmaMap (#0000)

EffortlessSteven:
### Motivation

- Replace sentinel-lookup and repeated binary-search patterns when querying pragma state with explicit APIs and a cursor-based fast path to reduce allocation/work in hot consumers while preserving existing semantics.

### Description

- Added `PragmaTracker::state_for_offset_with_cursor`, `PragmaTracker::final_state`, and an internal `normalized_state_for_index` helper to `crates/perl-pragma/src/lib.rs` and consolidated lookup normalization logic there.
- Replaced the `usize::MAX` sentinel lookup in `strict_warnings` with `PragmaTracker::final_state` to obtain the explicit final/top-level pragma state. 
- Migrated `version_compat` to use a caller-owned monotonic `pragma_cursor` with a backward-offset fallback to the binary-search path. 
- Threaded a `pragma_cursor` and a `pragma_last_offset` through `scope_analyzer` recursion and helpers so the analyzer uses the cheaper cursor path where traversal order allows; preserved non-monotonic fallbacks to maintain identical behavior. 
- Fixed a formatting literal in one semantic test message (escaped braces) to address a compile-time format-string issue.

### Testing

- Ran `cargo test -p perl-pragma` and the crate unit tests passed. 
- Ran `cargo test -p perl-lsp-rs-core strict_warnings` (targeted lints test) and it passed. 
- Ran `cargo test -p perl-semantic-analyzer` and the semantic analyzer tests (including `scope_analyzer`) passed after the migration. 
- Ran `cargo check --workspace --all-targets` and the workspace checked successfully.
- Note: a prior full `cargo test -p perl-lsp-rs-core` run showed one failing test (`test_perl_lsp_cargo_toml_removed_g1b_imports`) caused by a missing top-level fixture file unrelated to the pragma changes; targeted tests exercised by this PR are green.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb778322e88333942a0295228d70aa)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/diagnostics/lints/strict_warnings.rs
+++ b/crates/perl-lsp-rs-core/src/providers/diagnostics/lints/strict_warnings.rs
@@ -60,11 +60,10 @@ pub fn check_strict_warnings(node: &Node, diagnostics: &mut Vec<Diagnostic>) {
 
     let pragma_map = PragmaTracker::build(node);
     // Query the top-level pragma state (after all scoped blocks have exited).
-    // Using usize::MAX ensures we get the last entry, which reflects the
-    // restored top-level state after any eval/sub/block scopes have closed.
-    // This avoids the false-negative from .any() which sees eval-interior ranges.
+    // This reflects the restored top-level state after any eval/sub/block
+    // scopes have closed and avoids false-negatives from eval-interior ranges.
     // signatures_strict is included to honour `use feature 'signatures'` (#4038).
-    let top_level_state = PragmaTracker::state_for_offset(&pragma_map, usize::MAX);
+    let top_level_state = PragmaTracker::final_state(&pragma_map);
     let mut has_strict = top_level_state.strict_vars
         || top_level_state.strict_subs
         || top_level_state.strict_refs
@@ -589,7 +588,8 @@ mod tests {
     #[test]
     fn eval_then_top_level_strict_then_eval_no_strict_restores_correctly() {
         // eval { no strict; } after top-level use strict.
-        // usize::MAX must land on the restore entry (strict=true), not the inner no-strict.
+        // Final-state lookup must land on the restore entry (strict=true),
+        // not the inner no-strict.
         let diags =
             strict_warnings_diags("use strict;\nuse warnings;\neval { no strict; };\nmy $x = 1;\n");
         assert!(
--- a/crates/perl-lsp-rs-core/src/providers/diagnostics/lints/version_compat.rs
+++ b/crates/perl-lsp-rs-core/src/providers/diagnostics/lints/version_compat.rs
@@ -143,10 +143,21 @@ pub fn check_version_compat(node: &Node, diagnostics: &mut Vec<Diagnostic>) {
     };
 
     let pragma_map = PragmaTracker::build(node);
+    let mut pragma_cursor = 0usize;
+    let mut pragma_last_offset = None;
 
     // Second pass: walk AST for version-gated constructs.
     walk_node(node, &mut |n| {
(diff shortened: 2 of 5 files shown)


EffortlessSteven:
Closing as REDUNDANT — #6346 implements the same pragma-consumer migration with an additional fix: escaping format-string braces in scope_and_symbol_tests.rs to resolve a pre-existing workspace compile error.

Both PRs migrate strict_warnings, version_compat, and scope_analyzer identically; #6346 is more complete due to the format test fix.

**Extracted and noted for #6346 consideration**: Identical migration patterns; no architectural variants in this PR.

Thank you — the unified approach ensures consistent adoption of the new query APIs.

EffortlessSteven:
## Deep Review — PR #6345

**Verdict: APPROVED** with tests pushed directly to the branch.

### What I verified

**1. Monotonic cursor semantics — documented, contract is correct**

The doc comment on `state_for_offset_with_cursor` states the non-decreasing contract clearly and promises a transparent binary-search fallback for non-monotonic queries. Both call sites (version_compat, scope_analyzer) implement a pre-guard that routes backward offsets to the plain `state_for_offset` path, leaving the cursor unchanged. The internal `with_cursor` fallback then handles any stale-cursor state on the next call.

**2. Fallback path when offset < cursor position — correct**

Traced through the case where the scope_analyzer pre-guard fires (does NOT reset `pragma_cursor`) and the next call arrives at an offset between the backward offset and the old cursor position. The `with_cursor` internal fallback (`offset < pragma_map[*cursor - 1].0.start`) correctly re-runs `partition_point` and resets the cursor. No silent correctness hole here.

**3. `final_state()` vs `state_for_offset(usize::MAX)` — provably equivalent**

`state_for_offset(usize::MAX)` calls `partition_point(|r| r.start <= usize::MAX)` which returns `pragma_map.len()` (all entries qualify). `final_state` passes `pragma_map.len()` directly to `normalized_state_for_index`. Identical path.

**4. Thread safety — not a concern**

The cursor is a plain `usize` allocated on the caller's stack and passed as `&mut usize`. No shared mutable state between threads. The `PragmaTracker` struct itself is stateless (all methods take `pragma_map` as a parameter). No `Arc`/`Mutex` needed and none added.

### What was missing

Zero unit tests for the three new public APIs (`state_for_offset_with_cursor`, `final_state`, `normalized_state_for_index` via the public surface). I added 6 targeted tests to `crates/perl-pragma/tests/comprehensive_unit_tests.rs`:

- `final_state_empty_map_returns_default`
- `final_state_equals_state_for_offset_usize_max` — directly proves the equivalence claim
- `cursor_monotonic_matches_binary_search` — 6 ascending offsets, cursor vs. binary search
- `cursor_non_monotonic_fallback_is_correct` — backward jump then forward recovery
- `cursor_query_at_exact_range_boundary` — edge case: offset == range.start
- `cursor_out_of_bounds_is_clamped_and_correct` — cursor = len+999 clamp guard

All 129 tests pass (`cargo test -p perl-pragma`).

### One minor observation (not a blocker)

The `version_compat` fallback path calls `state_for_offset(...)` without updating `pragma_cursor` after the backward detour. This means the cursor is stale until the next `with_cursor` call's internal clamping fires. The result is still correct (tested), but it costs one extra `partition_point` on the first post-detour monotonic query. A `pragma_cursor = 0; pragma_last_offset = None;` reset inside the fallback arm would recover the cursor sooner — but it's a micro-optimization, not a bug.

EffortlessSteven:

PR #6408, opened 2026-04-24
feat(perl-token): add checked token spans and invariant helpers (#0000)

EffortlessSteven:
### Motivation

- Token byte spans could be malformed (for example `end < start`) and that silently allowed downstream bugs; the buffered token stream also synthesised EOF at position 0 which lost the original source EOF position.
- Make span correctness explicit and provide non-breaking, checked APIs so callers can opt into validation without changing the existing `Token::new` behavior.

### Description

- Introduced a typed `TokenSpan` and `TokenSpanError` to represent/validate byte spans and to return precise errors for inverted or disallowed spans.
- Added checked and helper APIs to `Token`: `try_new`, `new_checked`, `eof_at`, `unknown_at`, `span`, `range`, `with_span`, and `with_kind`, while keeping `Token::new` source-compatible and preserving saturating `len()` semantics.
- Updated `TokenStream` in `perl-parser-core` to track the last token end position (`eof_pos`) and synthesise EOF with `Token::eof_at(...)` so buffered-mode EOF preserves source position.
- Added focused tests in `crates/perl-token/tests/span_invariant_tests.rs` that exercise inverted spans, empty-span policy, EOF positioning, synthetic unknown tokens, and span/range helpers.

### Testing

- Ran `cargo xtask fmt` successfully to apply formatting across the workspace.
- Ran `cargo test -p perl-token` and all token tests passed (including the new span invariant tests).
- Ran `cargo test -p perl-parser-core token_stream` and the token-stream related tests passed.
- Ran `cargo check --workspace --all-targets` which failed due to a pre-existing, unrelated format-string compile error in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs`; this failure is not caused by the changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb77cb117c8333a2cec0aee42bbcea)

### DIFF
--- a/crates/perl-parser-core/src/tokens/token_stream.rs
+++ b/crates/perl-parser-core/src/tokens/token_stream.rs
@@ -59,6 +59,7 @@ enum TokenStreamInner<'a> {
 /// and statement-boundary state management used by the recursive-descent parser.
 pub struct TokenStream<'a> {
     inner: TokenStreamInner<'a>,
+    eof_pos: usize,
     peeked: Option<Token>,
     peeked_second: Option<Token>,
     peeked_third: Option<Token>,
@@ -69,6 +70,7 @@ impl<'a> TokenStream<'a> {
     pub fn new(input: &'a str) -> Self {
         TokenStream {
             inner: TokenStreamInner::Lexer(PerlLexer::new(input)),
+            eof_pos: 0,
             peeked: None,
             peeked_second: None,
             peeked_third: None,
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Closing as REDUNDANT ensemble peer of #6411. Same 4-shot task cluster; winner has equivalent or superior coverage. Thank you.

PR #6409, opened 2026-04-24
feat(perl-token): add checked token spans and invariant helpers (#0000)

EffortlessSteven:
### Motivation

- Make token span correctness explicit to avoid silent malformed spans while preserving the existing `Token::new` API for source compatibility.
- Ensure downstream parser, diagnostics, semantic tokens, and incremental parsing rely on explicit, checked span semantics.

### Description

- Add a small typed span helper `TokenSpan` and an error type `TokenSpanError` with checked construction (`TokenSpan::new`).
- Add non-breaking checked token constructors and helpers: `Token::try_new`, `Token::new_checked`, `Token::eof_at`, `Token::unknown_at`, `Token::span`, `Token::range`, `Token::with_span`, and `Token::with_kind`, while leaving `Token::new` unchanged.
- Define invariants so `start <= end` is enforced and empty spans are rejected for non-synthetic/non-EOF tokens, with `len()` still using saturating subtraction for compatibility.
- Update `TokenStream::from_vec`/buffered path to synthesise EOF at the last observed token end instead of `0`, and add a test asserting the behaviour.
- Add focused tests in `crates/perl-token/tests/span_invariant_tests.rs` and extend `crates/perl-parser-core/tests/token_stream_tests.rs` to cover synthesized EOF positioning.

### Testing

- Ran `cargo test -p perl-token` and all token tests passed successfully.
- Ran `cargo test -p perl-parser-core token_stream` and the token-stream tests (including the new EOF-position test) passed successfully.
- Ran `cargo xtask fmt` and `cargo check --workspace --all-targets`; the workspace check failed due to a pre-existing unrelated compile error in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs` (format string mismatch) and not because of the changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb77cb117c8333a2cec0aee42bbcea)

### DIFF
--- a/crates/perl-parser-core/src/tokens/token_stream.rs
+++ b/crates/perl-parser-core/src/tokens/token_stream.rs
@@ -50,7 +50,7 @@ enum TokenStreamInner<'a> {
     /// Live lexer producing tokens on demand from source text.
     Lexer(PerlLexer<'a>),
     /// Pre-lexed token buffer; used by [`TokenStream::from_vec`].
-    Buffered(VecDeque<Token>),
+    Buffered { tokens: VecDeque<Token>, eof_pos: usize },
 }
 
 /// Token stream that wraps perl-lexer or a pre-lexed token buffer.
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
Closing as REDUNDANT ensemble peer of #6411. Same 4-shot task cluster; winner has equivalent or superior coverage. Thank you.

PR #6416, opened 2026-04-24
feat(perl-token): centralize keyword and operator mapping helpers (#0000)

EffortlessSteven:
### Motivation

- Parser-core contained large ad-hoc string→`TokenKind` match tables for keywords/operators which are easy to get out of sync as `TokenKind` grows.
- `perl-token` should be the single source of truth for canonical spellings so other crates reuse the same mapping logic.
- Centralizing these mappings preserves case-sensitivity and keeps parser-core-specific contextual exceptions local to parser-core.

### Description

- Added canonical mapping helpers to `TokenKind` in `crates/perl-token/src/lib.rs`: `from_keyword`, `from_operator`, `from_delimiter`, and `from_sigil`, all returning `Option<TokenKind>` and preserving case-sensitivity.
- Replaced the large inline keyword/operator match tables in `crates/perl-parser-core/src/tokens/token_stream.rs` with calls to the new `TokenKind` helpers while retaining parser-core contextual behavior (e.g. `qw` remains treated as an identifier, `no` and `*` are handled as special contextual fallbacks, and delimiter fallbacks are preserved).
- Added exhaustive mapping unit tests in `crates/perl-token/tests/token_kind_mapping_tests.rs` to cover keywords, operators, delimiters, and sigils and their negative cases.
- Added parser-core conformance tests in `crates/perl-parser-core/tests/token_stream_conformance.rs` to verify lexer→parser token conversion now uses the canonical mappings and that quote-like keywords remain identifiers.

### Testing

- Ran `cargo test -p perl-token` and all token crate tests passed (including new mapping tests).
- Ran `cargo test -p perl-parser-core token_stream` and the parser-core token stream tests passed, and the new conformance test suite passed via `cargo test -p perl-parser-core --test token_stream_conformance`.
- Ran `cargo test -p perl-lexer` to ensure lexer behavior remains unchanged and it passed.
- Ran `cargo check --all-targets -p perl-token -p perl-parser-core` which passed; a full `cargo check --workspace --all-targets` was attempted but reported a pre-existing unrelated failure in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs` (format string mismatch) outside the scope of these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb77d1dea08333be77e707831301f9)

### DIFF
--- a/crates/perl-parser-core/src/tokens/token_stream.rs
+++ b/crates/perl-parser-core/src/tokens/token_stream.rs
@@ -343,121 +343,19 @@ impl<'a> TokenStream<'a> {
     fn convert_lexer_token(token: LexerToken) -> Token {
         let kind = match &token.token_type {
             // Keywords
-            LexerTokenType::Keyword(kw) => match kw.as_ref() {
-                "my" => TokenKind::My,
-                "our" => TokenKind::Our,
-                "local" => TokenKind::Local,
-                "state" => TokenKind::State,
-                "sub" => TokenKind::Sub,
-                "if" => TokenKind::If,
-                "elsif" => TokenKind::Elsif,
-                "else" => TokenKind::Else,
-                "unless" => TokenKind::Unless,
-                "while" => TokenKind::While,
-                "until" => TokenKind::Until,
-                "for" => TokenKind::For,
-                "foreach" => TokenKind::Foreach,
-                "return" => TokenKind::Return,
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
Closing as REDUNDANT ensemble peer of #6419. Same 4-shot task cluster; winner has equivalent or superior coverage. Thank you.

PR #6418, opened 2026-04-24
feat(perl-token): centralize keyword and operator mapping helpers (#0000)

EffortlessSteven:
### Motivation

- Consolidate the canonical mapping between spellings and `TokenKind` into `perl-token` so the parser no longer duplicates string-to-kind logic.
- Reduce subtle bugs from ad-hoc match tables as `TokenKind` grows by making `perl-token` the single source of truth.
- Preserve contextual lexer/parser decisions (case sensitivity, quote-like operators, identifier-position sigils) at the call site rather than baking them into the mapping helpers.

### Description

- Add mapping helpers on `TokenKind` in `crates/perl-token/src/lib.rs`: `from_keyword`, `from_operator`, `from_delimiter`, and `from_sigil` (case-sensitive and limited to canonical spellings).
- Replace large ad-hoc match tables in `crates/perl-parser-core/src/tokens/token_stream.rs::convert_lexer_token` with calls to the new `TokenKind` helpers while preserving parser-core special-cases for `qw` and bare sigil identifier behaviour.
- Add exhaustive mapping unit tests in `crates/perl-token/tests/token_kind_mapping_tests.rs` that cover every canonical keyword/operator/delimiter/sigil spelling and verify contextual rules (e.g. `BEGIN` case, `qw` not treated as a plain keyword).
- Add token-stream conformance tests in `crates/perl-parser-core/tests/token_stream_conformance.rs` that assert the refactor preserved runtime classification for representative sequences.

### Testing

- Ran `cargo test -p perl-token` and it passed (token mapping unit tests and existing `perl-token` suites succeeded).
- Ran `cargo test -p perl-parser-core token_stream` and the token-stream tests passed.
- Ran `cargo test -p perl-lexer` and the lexer test suite passed.
- Ran `cargo xtask fmt` to ensure formatting; it completed successfully.
- Ran `cargo check --workspace --all-targets` which failed due to an existing, unrelated compile error in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs` (format string argument missing); this failure is not caused by the mapping refactor.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb77d1dea08333be77e707831301f9)

### DIFF
--- a/crates/perl-parser-core/src/tokens/token_stream.rs
+++ b/crates/perl-parser-core/src/tokens/token_stream.rs
@@ -343,121 +343,22 @@ impl<'a> TokenStream<'a> {
     fn convert_lexer_token(token: LexerToken) -> Token {
         let kind = match &token.token_type {
             // Keywords
-            LexerTokenType::Keyword(kw) => match kw.as_ref() {
-                "my" => TokenKind::My,
-                "our" => TokenKind::Our,
-                "local" => TokenKind::Local,
-                "state" => TokenKind::State,
-                "sub" => TokenKind::Sub,
-                "if" => TokenKind::If,
-                "elsif" => TokenKind::Elsif,
-                "else" => TokenKind::Else,
-                "unless" => TokenKind::Unless,
-                "while" => TokenKind::While,
-                "until" => TokenKind::Until,
-                "for" => TokenKind::For,
-                "foreach" => TokenKind::Foreach,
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
Closing as REDUNDANT ensemble peer of #6419. Same 4-shot task cluster; winner has equivalent or superior coverage. Thank you.

PR #6429, opened 2026-04-24
perf(perl-token): add borrowed token view for allocation-sensitive paths

EffortlessSteven:
### Motivation

- Reduce allocations on hot token-processing paths by providing a borrowed view that avoids constructing `Arc<str>`-backed `Token` values for temporary inspection and classification.
- Preserve the existing `Token` API and make conversion to an owned `Token` explicit so callers opt into allocation only when needed.

### Description

- Add `TokenRef<'src>` as a non-breaking borrowed token view with fields `kind`, `text: &'src str`, `start`, and `end`, plus helpers `new`, `to_owned_token`, `len`, `is_empty`, `span`, and `display_name` and `impl From<TokenRef<'_>> for Token`.
- Add ergonomic helpers on `Token`: `span()`, `display_name()`, and `as_ref_token()` which returns a `TokenRef<'_>` without allocating.
- Add allocation-focused unit tests at `crates/perl-token/tests/borrowed_token_tests.rs`, a Criterion bench `crates/perl-token/benches/token_borrowed_scorecard.rs`, update `crates/perl-token/Cargo.toml`, and document `TokenRef<'src>` and the bench in `crates/perl-token/README.md`.

### Testing

- Ran `cargo test -p perl-token`, which passed (all new and existing `perl-token` tests succeeded).
- Ran `cargo bench -p perl-token --bench token_borrowed_scorecard`, which passed and reported an allocation scorecard showing `borrowed=0` vs `owned=16384` over the sample set.
- Ran `cargo check --all-targets -p perl-token`, which passed; `cargo check --workspace --all-targets` fails due to a pre-existing unrelated compile error in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs:3919` and is not caused by these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb77d93b5c8333a32e3f440e7f53c7)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1896,6 +1896,9 @@ version = "0.12.4"
 [[package]]
 name = "perl-token"
 version = "0.12.4"
+dependencies = [
+ "criterion",
+]
 
 [[package]]
 name = "perl-uri"
--- a/crates/perl-token/Cargo.toml
+++ b/crates/perl-token/Cargo.toml
@@ -16,8 +16,15 @@ categories = ["parsing", "text-processing"]
 [dependencies]
 # No external dependencies for now (Arc is std)
 
+[dev-dependencies]
+criterion.workspace = true
+
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [lints]
 workspace = true
+
+[[bench]]
+name = "token_borrowed_scorecard"
+harness = false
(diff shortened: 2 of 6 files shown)


EffortlessSteven:
Closing as REDUNDANT ensemble peer of #6432. Same 4-shot task cluster; winner has equivalent or superior coverage. Thank you.

PR #6431, opened 2026-04-24
perf(perl-token): add borrowed token view for allocation-sensitive paths (#0000)

EffortlessSteven:
### Motivation
- Hot paths that only need to inspect tokens should avoid allocating `Arc<str>` for transient classification or diagnostics. 
- Provide a lightweight borrowed view so callers can operate on `&str` and convert to owned `Token` only when necessary.

### Description
- Add `TokenRef<'src>` with fields `kind`, `text: &'src str`, `start`, and `end`, and helper methods `len()`, `is_empty()`, `span()`, and `display_name()` (non-breaking to existing API). 
- Add explicit conversions: `TokenRef::to_owned_token()` and `impl From<TokenRef<'_>> for Token`, plus `Token::as_ref_token()` for borrowing from an owned `Token`. 
- Add focused unit tests in `crates/perl-token/tests/borrowed_token_tests.rs` covering helpers, malformed spans, and conversions. 
- Add a Criterion bench `crates/perl-token/benches/token_scorecard.rs` and a dev-dependency/bench entry in `crates/perl-token/Cargo.toml` to demonstrate allocation difference; update `README.md` and docs with usage example. 

### Testing
- Ran `cargo test -p perl-token`, which passed (all `perl-token` unit tests and doctests succeeded). 
- Ran benchmark `cargo bench -p perl-token --bench token_scorecard -- --noplot`, which executed and reported the expected allocation difference (owned construction shows allocations while `TokenRef` construction reports zero allocations). 
- Ran `cargo check --workspace --all-targets`, which failed due to a pre-existing, unrelated formatting/format-string error in `crates/perl-semantic-analyzer/tests/scope_and_symbol_tests.rs` and not due to changes in `perl-token`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb77d93b5c8333a32e3f440e7f53c7)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1896,6 +1896,9 @@ version = "0.12.4"
 [[package]]
 name = "perl-token"
 version = "0.12.4"
+dependencies = [
+ "criterion",
+]
 
 [[package]]
 name = "perl-uri"
--- a/crates/perl-token/Cargo.toml
+++ b/crates/perl-token/Cargo.toml
@@ -16,6 +16,13 @@ categories = ["parsing", "text-processing"]
 [dependencies]
 # No external dependencies for now (Arc is std)
 
+[dev-dependencies]
+criterion.workspace = true
+
+[[bench]]
+name = "token_scorecard"
+harness = false
+
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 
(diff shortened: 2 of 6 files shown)


EffortlessSteven:
Closing as REDUNDANT ensemble peer of #6432. Same 4-shot task cluster; winner has equivalent or superior coverage. Thank you.

PR #6730, opened 2026-04-25
fix(navigation): qualified sub declarations missing from find-references results (#6748)

EffortlessSteven:
## Summary

- **Bug:** In `find_references_single_file`, the inner `walk` function's `Subroutine` arm compared the full qualified name (e.g. `"Foo::bar"`) directly against `want_name` (the bare name `"bar"`). The comparison always failed for package-qualified subroutine declarations, so `sub Foo::bar { }` was silently dropped from reference results.
- **Fix:** Apply the same `split_qualified_name` pattern already used in the `FunctionCall` arm — split the subroutine name into package + bare, then match both against `want_pkg` / `want_name`.
- **Tests:** Three new regression tests added inline: qualified sub found, bare sub unchanged, cross-package non-confusion.

## Root cause

`references.rs:91` (before fix):
```rust
NodeKind::Subroutine { name: Some(name), .. } if want_kind == "sub" => {
    if name == want_name {   // "Foo::bar" == "bar" → always false
        out.push(...);
    }
}
```

The `FunctionCall` arm two lines above already used `split_qualified_name` correctly; the `Subroutine` arm did not.

## Test plan

- [x] `cargo test -p perl-lsp-rs-core --lib -- providers::navigation::references::tests` — 3/3 pass
- [x] `cargo clippy -p perl-lsp-rs-core --lib` — clean
- [x] `cargo fmt -p perl-lsp-rs-core` — no changes

https://claude.ai/code/session_01X1LhJaciGj7u3myeHEQHR7

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1LhJaciGj7u3myeHEQHR7)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/navigation/references.rs
+++ b/crates/perl-lsp-rs-core/src/providers/navigation/references.rs
@@ -88,7 +88,9 @@ pub fn find_references_single_file(ast: &Node, offset: usize) -> Option<Vec<(usi
                 }
             }
             NodeKind::Subroutine { name: Some(name), .. } if want_kind == "sub" => {
-                if name == want_name {
+                let (pkg, bare) = split_qualified_name(name);
+                let pkg = pkg.unwrap_or("main");
+                if bare == want_name && pkg == want_pkg {
                     out.push((location.start, location.end));
                 }
             }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
standards review: clean. Fix correctly applies the same split_qualified_name pattern already used in the FunctionCall arm two lines above (one-liner + 2-line condition). Three regression tests cover qualified-sub-found, bare-sub-still-found, cross-package-not-confused. Test code uses .expect() and .unwrap() in line with existing style in this crate (e.g., providers/code_actions/code_actions.rs already uses .expect() at lines 461/589/615) — not flagging since it matches established crate convention.

EffortlessSteven:
**Maintainer-PR review: ALIGNED**

Real navigation correctness fix: the `Subroutine` arm of `find_references_single_file` was comparing the full qualified name (`Foo::bar`) directly against the bare lookup name (`bar`), so package-qualified declarations were silently dropped. Fix applies the same `split_qualified_name` pattern already used in the `FunctionCall` arm two lines above. Three regression tests inline (qualified found, bare unchanged, cross-package isolation). Single-file scope. This is the upstream of the broader pattern-fix sweep in #6756. Aligned with v0.13.0 navigation-correctness direction.

EffortlessSteven:
## pr-responded sweep

All bot comments addressed or non-actionable:
- gemini-code-assist: daily quota exhausted — no review delivered
- coderabbitai: per-commit rate limit exceeded — no review delivered

CI status: only failing check was `validate-title` due to PR title missing required `(#NN)` issue ref suffix. Per house convention for autonomous fix-codebase PRs with no upstream issue, applied `(#0000)` placeholder. Title edit re-triggers validate-title; awaiting fresh result. No PR-side substantive issues; mechanical fix only.

— pr-responded

EffortlessSteven:
## Refactor planner: no refactor needed

Diff is small and focused; code reads cleanly within its existing module. No duplication, dead code, type-tightness, naming, or reuse opportunities stand out at this scope.

Apply `refactor-planner-reviewed` AND `green-refactor-reviewed` (sign-off; nothing to refactor).

EffortlessSteven:
## Deep Review — CORRECT

**Verdict: CORRECT**

**Core fix logic:** Before the fix, the `Subroutine` arm in the inner `walk` closure compared `name == want_name` directly — "Foo::bar" == "bar" is always false. The fix applies `split_qualified_name(name)` (already imported and used correctly in both the outer `match` and the `FunctionCall` arm) to extract `(pkg, bare)`, then compares `bare == want_name && pkg == want_pkg`. This is structurally identical to the `FunctionCall` arm treatment.

**Qualified name edge cases verified:**

1. `A::b` → `split_qualified_name` returns `(Some("A"), "b")`, `pkg = "A"`. Correct.
2. `A::B::c` → `rfind("::")` returns the last `::`, giving `(Some("A::B"), "c")`. Matches `want_pkg = "A::B"` set in the outer `match`. Correct.
3. `main::foo` → `(Some("main"), "foo")`. `want_pkg = "main"`. Correct.
4. `foo` (bare sub) → `(None, "foo")`, `pkg = pkg.unwrap_or("main") = "main"`. The outer `match` on a bare `Subroutine { name: "foo" }` also produces `want_pkg = "main"`. Match is symmetric. Correct.
5. `::foo` (top-level sigil form) → `rfind("::") = Some(0)`, `pkg = Some("")` (empty string). `pkg.unwrap_or("main")= ""`. The outer match would likewise extract `pkg = ""`. The match is symmetric — both sides produce the same empty-string package — but `""  != "main"` means bare-sub lookups won't falsely match `::foo` declarations, and `::foo` declarations won't match bare-sub lookups. Correct behavior, though a Perl `::foo` declaration is unusual in practice.

**Substantive observation:** The three new inline tests cover the three important cases (qualified found, bare unchanged, cross-package non-confusion). The `qualified_sub_declaration_found_in_references` test places the cursor at byte offset 4, which in `"sub Foo::bar { 1 } Foo::bar();"` is the `F` of `Foo::bar`. The `find_node_at_offset` traversal will return the innermost node containing offset 4 — this should be the `Subroutine` node itself (or a child if the parser creates one for the qualified name). The test asserts `refs.len() >= 2`, which is correct if both the declaration and the call site are found.

PR #6732, opened 2026-04-25
fix(parser): correct InvalidDelimiter error for tr{abc}xyz paired-delimiter forms (#6748)

EffortlessSteven:
## Summary

- Fix wrong error type returned when `tr{abc}xyz` has an alphanumeric character after a paired search delimiter — was `MissingReplacement`, should be `InvalidDelimiter`
- Remove dead code: the duplicate `else if let Some(repl_delimiter)` branch in `extract_transliteration_parts_strict` was unreachable (the preceding `if let Some(...)` already consumed any non-empty `trimmed` string)

## Root Cause

In `crates/perl-parser-core/src/syntax/quote.rs`, `extract_transliteration_parts_strict`, the `is_paired` branch (line ~441) had two issues:

1. **Wrong error variant**: `tr{abc}xyz` correctly hits the alphanumeric check but returned `TransliterationError::MissingReplacement`. The replacement character (`x`) is present — it's just invalid. The correct error is `InvalidDelimiter('x')`.
2. **Dead code**: The `else if let Some(repl_delimiter) = trimmed.chars().next()` branch could never be reached because the immediately preceding `if let Some(repl_delimiter) = trimmed.chars().next()` already matched the same condition.

## Test plan

- [ ] `cargo test -p perl-parser-core` — all transliteration strict tests pass, no regressions
- [ ] `cargo clippy -p perl-parser-core --lib` — clean
- [ ] `cargo xtask fmt` — clean

https://claude.ai/code/session_01WWbA5Wd8zL7FSMvNnsPUVL

---
_Generated by [Claude Code](https://claude.ai/code/session_01WWbA5Wd8zL7FSMvNnsPUVL)_

PR #6734, opened 2026-04-25
fix(parser): tr{} with alphanumeric replacement emits correct InvalidDelimiter error (#0000)

EffortlessSteven:
## Summary

- Fixes a bug in `extract_transliteration_parts_strict` where `tr{abc}xyz;` (paired bracket delimiter followed by an alphanumeric replacement) emitted `MissingReplacement` instead of `InvalidDelimiter`
- Removes 5 lines of unreachable dead code: the `else if let Some(repl_delimiter) = trimmed.chars().next()` branch had an identical condition to the `if let Some(...)` above it and could never execute

## Root cause

In `crates/perl-parser-core/src/syntax/quote.rs`, the `is_paired` replacement-parsing branch (line 442) checked whether the replacement started with an alphanumeric/whitespace character but returned the wrong error variant. The dead `else if` block below it had the correct variant (`InvalidDelimiter`) but was unreachable.

## Test plan

- [ ] `cargo test -p perl-parser-core` passes (was failing on `transliteration_strict_handles_edge_cases` before this fix)
- [ ] `cargo clippy -p perl-parser-core --lib` is clean

https://claude.ai/code/session_01SvCToS5nQAf1VpX1Hx74Zk

---
_Generated by [Claude Code](https://claude.ai/code/session_01SvCToS5nQAf1VpX1Hx74Zk)_

### DIFF
--- a/crates/perl-parser-core/src/syntax/quote.rs
+++ b/crates/perl-parser-core/src/syntax/quote.rs
@@ -440,17 +440,12 @@ pub fn extract_transliteration_parts_strict(
         let trimmed = rest1.trim_start();
         if let Some(repl_delimiter) = trimmed.chars().next() {
             if repl_delimiter.is_ascii_alphanumeric() || repl_delimiter.is_whitespace() {
-                return Err(TransliterationError::MissingReplacement);
+                return Err(TransliterationError::InvalidDelimiter(repl_delimiter));
             }
             let repl_closing = get_closing_delimiter(repl_delimiter);
             let (body, rest, found_closing) =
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Superseded by #6732: same one-line fix to `extract_transliteration_parts_strict` (changing the alphanumeric replacement-delimiter error from `MissingReplacement` to `InvalidDelimiter`) and the same removal of the unreachable second `else if` branch. #6732 also adds a multi-line comment explaining why the alphanumeric case is an invalid delimiter. Closing as duplicate.

PR #6776, opened 2026-04-26
perf(critic-cache): bound LRU cache and add content-hash validation

EffortlessSteven:
## Summary

- **Bounded LRU eviction**: `CriticConfig::max_cache_entries` (default 512) caps the external perlcritic scan cache. When full, the least-recently-used entry is evicted to admit the new result, preventing unbounded memory growth on large projects.
- **Content-hash validation**: new `analyze_file_with_hash(path, hash)` method re-runs perlcritic whenever the stored hash diverges from the current document content — catching stale entries from external file changes (git checkout, external editor) that bypass the LSP `didChange` notification.
- Both `collect_external_perlcritic_diagnostics` call sites in `diagnostics.rs` now pass the in-memory document text hash, so every diagnostic cycle automatically invalidates stale cache entries.
- `analyze_file` (path-only, used by one-shot CLI callers) retains its existing signature and gains LRU eviction only.

## Test plan

- [ ] `cargo test -p perl-lsp-rs-core --lib -- perl_critic` — 9 new unit tests: cache hit/miss, content-hash staleness, LRU capacity limit, zero-capacity no-op, hash determinism (all pass)
- [ ] `RUST_TEST_THREADS=2 cargo test -p perl-lsp-rs --lib -- --test-threads=2` — 461 existing tests, 0 failures

https://claude.ai/code/session_015EN4ALtXM6bQMpoCcFPi3Q

---
_Generated by [Claude Code](https://claude.ai/code/session_015EN4ALtXM6bQMpoCcFPi3Q)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/tooling/perl_critic/analyzer.rs
+++ b/crates/perl-lsp-rs-core/src/tooling/perl_critic/analyzer.rs
@@ -164,6 +164,15 @@ impl CriticAnalyzer {
         self.cache.len()
     }
 
+    /// Returns `true` if the cache contains an entry for the given path string.
+    ///
+    /// Intended for tests — allows assertions on which specific entries were
+    /// evicted by the LRU policy.
+    #[cfg(test)]
+    pub fn cache_contains(&self, path: &str) -> bool {
+        self.cache.contains_key(path)
+    }
+
     /// Convert violations to diagnostics
     #[cfg(feature = "lsp-compat")]
     pub fn to_diagnostics(&self, violations: &[Violation]) -> Vec<lsp_types::Diagnostic> {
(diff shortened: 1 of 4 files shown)


EffortlessSteven:
standards review: clean — substantive perf work on perl_critic analyzer: bounded LRU eviction (default 512 entries) plus content_hash-based stale-entry detection via the new analyze_file_with_hash() entry point. The original analyze_file() is preserved for path-only callers. Diagnostics caller in runtime/diagnostics.rs is updated to compute hash_content(doc_text) and route through the new path. Tests cover hit-on-same-hash, miss-on-different-hash, LRU eviction at capacity, zero-disabled cache, and invalidate_cache. Banned-pattern scan clean. Minor: make_analyzer_with_output(_output) test helper accepts a parameter that's unused (the comment admits MockSubprocessRuntime always returns empty stdout) — cosmetic, not blocking.

EffortlessSteven:
## Maintainer-PR review — ALIGNED (with one non-blocking title note)

Adds bounded LRU eviction (`max_cache_entries`, default 512) to the perlcritic external scan cache, plus content-hash validation via new `analyze_file_with_hash(path, hash)` so stale entries from external file changes (git checkout, external editor) are caught even when LSP `didChange` doesn't fire. Both `collect_external_perlcritic_diagnostics` call sites in `diagnostics.rs` now pass the in-memory document hash. 4 files, 211/15.

Aligns with the "polish for release" phase: unbounded caches on large projects are exactly the failure mode that bites users at scale. The content-hash invalidation path closes a real correctness gap (external mutation bypassing the LSP notification stream). 9 new unit tests cover hit/miss, staleness, capacity, zero-cap no-op, hash determinism — appropriate test density for a caching layer.

Architecture is right: the cache lives in `perl-lsp-rs-core/src/tooling/perl_critic/`, the diagnostic-cycle integration in `perl-lsp-rs/src/runtime/diagnostics.rs` — correct layering. Backward-compatible: `analyze_file` retains its signature for one-shot CLI callers.

**Non-blocking title note**: title is `perf(critic-cache): bound LRU cache and add content-hash validation` — missing the `(#NNNN)` issue ref that `validate-title` CI check requires (per `feedback_validate_title_issue_ref.md`). 1-line fix before merge.

Routing: maintainer-pr-reviewed.

EffortlessSteven:
## Deep Review — CORRECT (fix pushed)

**#6776** (bounded LRU critic cache with content-hash validation)

Core logic is correct:
- `insert_entry` guard `self.cache.len() >= max && !self.cache.contains_key(&path)` correctly avoids eviction when updating an existing key at capacity.
- `CriticAnalyzer` is always accessed through a `Mutex` in production (`diagnostics.rs` line ~69), so no concurrent mutation race.
- `DefaultHasher` is appropriate — within-process only, doc comment says so, no stability-across-restarts requirement.
- The `analyze_file` / `content_hash: 0` sentinel: `analyze_file` does a path-only lookup (no hash comparison), so the sentinel never causes a false cache validation miss. The only theoretical issue (`hash_content` returning 0) is astronomically unlikely and consequence-free (correct violations returned regardless).

**Fix pushed (direct fix-forward):**

Two test gaps filled in commit `d08e1ac`:
1. `lru_eviction_respects_max_cache_entries` — now re-accesses file1/file2 before the 4th insert (making file0 definitively the LRU), then asserts with `cache_contains` which entry was evicted vs. which survive. The previous version only checked `cache_len == 3` — correct but doesn't verify *identity* of evicted entry.
2. New test `lru_update_existing_key_at_capacity_does_not_over_evict` — regression test for the edge case where updating an existing key at capacity must not trigger eviction.

Added `cache_contains` helper (cfg(test) only) to enable identity assertions.

Verdict: **CORRECT**. Applying `deep-reviewed`.

PR #5381, opened 2026-04-23
test(perl-lsp-ux-tests): expand live-edit UX harness coverage (#0000)

EffortlessSteven:
### Motivation

- Provide first-class live-edit coverage in the UX harness so scenarios can exercise `textDocument/didChange` and validate diagnostics + navigation responsiveness in a BDD-style flow.
- Make the harness more ergonomic for SRP-style scenarios by surfacing a simple API for full-document edits and deterministic versioning.

### Description

- Added `UxClient::did_change` to send `textDocument/didChange` notifications with explicit `version` and `contentChanges` in `crates/perl-lsp-ux-tests/src/client.rs`.
- Tracked per-document versions in `UxHarness` via a `document_versions: Mutex<HashMap<String,i32>>` and added `change_file_full` plus `bump_document_version` in `crates/perl-lsp-ux-tests/src/lib.rs` to drive full-document edits.
- Made diagnostics polling in `wait_for_diagnostics` prefer the newest matching `publishDiagnostics` event by scanning buffered events in reverse to improve post-edit assertions.
- Added a BDD-style scenario `ux_scenario_19_live_edit_feedback.rs` exercising undeclared-variable diagnostics on open and navigation responsiveness after a `didChange`, and registered the scenario in `fixtures/editor_ux_fixture_matrix.json`.

### Testing

- Ran `cargo xtask fmt` which completed successfully.
- Initial `cargo test -p perl-lsp-ux-tests` in this environment hit failures due to the `perl-lsp` binary not being discovered (expected in some CI setups); after building the binary with `cargo build -p perl-lsp-rs --bin perl-lsp` this was resolved.
- Verified the new scenario and matrix with `PERL_LSP_BIN=/workspace/perl-lsp/target/debug/perl-lsp cargo test -p perl-lsp-ux-tests --test ux_scenario_19_live_edit_feedback --test editor_ux_fixture_matrix` which passed.
- Also ran `cargo check --all-targets -p perl-lsp-ux-tests` and `cargo clippy -p perl-lsp-ux-tests`, both of which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea0bc5454c8333b5f6bbf79a7bf059)

PR #6851, opened 2026-04-26
test(perl-lsp-ux-tests): expand live-edit UX harness coverage [rebased] (#6851)

EffortlessSteven:
Rebased recovery of #5381 onto fresh-root master per the cherry-pick recovery batch (per `feedback_fresh_root_master_rebuild.md`). Original PR was stranded with no merge-base after master became a fresh root commit.

## Summary
- Adds `UxClient::did_change` for `textDocument/didChange` notifications with explicit version/contentChanges.
- Tracks per-document versions in `UxHarness` via `document_versions: Mutex<HashMap<String,i32>>` and adds `change_file_full` plus `bump_document_version`.
- Adds BDD-style `ux_scenario_24_live_edit_feedback.rs` exercising undeclared-variable diagnostics on open and navigation responsiveness after a `didChange`.
- Registers the scenario in `fixtures/editor_ux_fixture_matrix.json` (preserved master's `workspace_folder_addition_freshness` entry alongside the new one).
- Includes the second commit (`f2cf3db2`) that adds a `binary_available` skip guard.

## Test plan
- [x] `cargo build -p perl-lsp-ux-tests` (clean)
- [x] `cargo check -p perl-lsp-ux-tests --tests` (clean)
- [ ] CI runs (full ux scenario suite)

Original PR: #5381

### DIFF
--- a/crates/perl-lsp-rs/src/features/implementation_provider.rs
+++ b/crates/perl-lsp-rs/src/features/implementation_provider.rs
@@ -317,7 +317,8 @@ impl ImplementationProvider {
         match &node.kind {
             NodeKind::Subroutine { name: Some(name), .. } => {
                 let (_, name_bare) = perl_parser::qualified_name::split_qualified_name(name);
-                let (_, method_bare) = perl_parser::qualified_name::split_qualified_name(method_name);
+                let (_, method_bare) =
+                    perl_parser::qualified_name::split_qualified_name(method_name);
                 if name_bare == method_bare {
                     let target_uri = parse_uri(uri);
                     results.push(LocationLink {
--- a/crates/perl-lsp-rs/src/runtime/language/hover.rs
+++ b/crates/perl-lsp-rs/src/runtime/language/hover.rs
@@ -1725,7 +1725,8 @@ impl LspServer {
                 }
             }
             NodeKind::Method { name: method_name, .. } => {
-                let (_, method_bare) = perl_parser::qualified_name::split_qualified_name(method_name);
+                let (_, method_bare) =
+                    perl_parser::qualified_name::split_qualified_name(method_name);
                 let (_, name_bare) = perl_parser::qualified_name::split_qualified_name(name);
                 if method_bare == name_bare {
                     return Some(node);
--- a/crates/perl-lsp-rs/src/runtime/symbol_extraction.rs
+++ b/crates/perl-lsp-rs/src/runtime/symbol_extraction.rs
@@ -314,7 +314,10 @@ impl LspServer {
             }
 
             NodeKind::FunctionCall { name, args } => {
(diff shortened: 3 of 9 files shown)


EffortlessSteven:
**standards review: BLOCKER**

This PR is titled "expand live-edit UX harness coverage" but contains a hidden semantic change in shared infrastructure that affects ALL callers of `wait_for_diagnostics()`.

Specifically `crates/perl-lsp-ux-tests/src/lib.rs` line 463:
```diff
-                for ev in &events {
+                for ev in events.iter().rev() {
```

This changes `wait_for_diagnostics` from returning the FIRST published diagnostic to returning the LAST. Every other UX scenario that calls this helper now receives different data. The change has no test, no docstring update, and no rationale in the PR description.

If the goal is to surface the most recent diagnostic for the live-edit scenario, that semantics change should be:
1. Done in a separate, named helper (e.g. `wait_for_latest_diagnostics`), OR
2. Documented in the docstring with a callout that earlier diagnostics are now ignored, AND
3. Verified to be a no-op for all existing scenarios (since `peek_events` returns events in arrival order, the FIRST diagnostic for a URI is normally the one tests want to assert against).

Other observations (advisory, not blocking):
- Branch base `d31334d93` is stale; current master `034982798` includes the fmt cascade from #6849. After updating the branch the noisy fmt-only diff in DAP/lexer/lsp-rs/semantic-analyzer/perl-symbol files will collapse to just the unique scope.

Please remove the `wait_for_diagnostics` change OR justify it with a docstring + targeted test before this re-enters the pipeline.

EffortlessSteven:
**Builder fix applied** — commit 659dc97ad

**Root cause**: The standards reviewer flagged that `wait_for_diagnostics` was changed from `events.iter()` (first published) to `events.iter().rev()` (most recent) as a hidden semantic change affecting ALL callers. This broke the behavior-preserving contract for tests that rely on "first published" semantics (e.g., initial diagnostics on file open).

**Fixes applied** (in `crates/perl-lsp-ux-tests/`):

1. **`src/lib.rs`**: Reverted `wait_for_diagnostics` to `events.iter()` (first published; original behavior). Updated docstring to document this explicitly. Added new `wait_for_latest_diagnostics` method using `events.iter().rev()` (most recently published, for post-edit state). Includes cross-reference doc links between the two.

2. **`tests/ux_scenario_24_live_edit_feedback.rs`**: Updated the `_post_edit_diagnostics` call (after `change_file_full`) to use `wait_for_latest_diagnostics` — this is the call that genuinely needs post-edit state. The initial `before` diagnostics calls correctly remain on `wait_for_diagnostics`.

3. **`tests/ux_scenario_24_live_edit_feedback.rs`**: Fixed `has_global_symbol_diagnostic` to require BOTH `code.contains("Global symbol") && message.contains(symbol)` in the code branch — the old version matched on code alone without checking the `symbol` argument, making the predicate vacuous for code-based matches.

**Verified**: `cargo check -p perl-lsp-ux-tests --tests` compiles cleanly.

Removing `needs-builder-fix` label.

Issue #7015, opened 2026-04-27
ci: xtask gates fail in detached HEAD state on PR runs ("HEAD does not point to a branch") — blocks #6893, #6887 (#0000)


EffortlessSteven:
## Master-bit-rot incident

Per the conveyor doctrine (`project_conveyor_doctrine.md`): 3+ PRs failing the same gate identically with disjoint diffs = queue-level incident, not per-PR debugging.

## Symptom

CI Gate (Merge-Blocking) FAILURE on multiple PRs with identical error:
```
fatal: Needed a single revision
fatal: HEAD does not point to a branch
```
(5+ occurrences in the failed-step logs)

## Affected PRs (verified disjoint diffs)

- **#6893** feat(parser-ratchet): select parser gate from scope/risk — touches `.ci/scope.d/`, fixtures, docs
- **#6887** feat(generated): generated-file ownership check — touches `xtask/`, `.ci/generated-files.toml`, schema, docs
- (likely additional candidates — pr-responder identified 2 from a sample of 4; full sweep needed)

Both PRs were ensemble-curator winners specifically picked for being clean. Their failure is identical — same git-state error, not PR-attributable.

## Diagnosis

xtask gate code (probably parser-ratchet or generated-file gate, or a shared helper) runs `git rev-parse <something>` or similar that requires HEAD to be on a branch. GitHub Actions checks out PRs in detached HEAD by default. The gate works on master push (HEAD is master) but breaks on `pull_request` event.

## Fix candidates

1. **Use `${{ github.head_ref }}` / `${{ github.base_ref }}`** environment variables instead of `git rev-parse HEAD` in xtask gates
2. **OR** in workflow: `git checkout -b ci-tmp` before invoking xtask, to give it a branch name
3. **OR** harden the xtask gate's git invocations: detect detached HEAD and use `git rev-parse HEAD~0` or similar that doesn't require a branch

## Acceptance criteria

- [ ] Identify which xtask gate(s) emit the `HEAD does not point to a branch` error
- [ ] Fix in the xtask source (preferred — affects all workflows that invoke it)
- [ ] Verify fix on a test PR (open a one-line PR, confirm the gate runs cleanly)
- [ ] Cascade-update affected PRs (#6893, #6887, plus any others surfaced by sweep)

## Next-action sweep

```
gh pr list --state open --json number,title --jq '.[] | .number' | head -50 | while read pr; do
[...]

PR #6907, opened 2026-04-26
feat(parser-ratchet): add perl-corpus concept floors (#6847)

EffortlessSteven:
### Motivation

- Base-vs-candidate parser ratchet checks can pass even when the repo-owned corpus is already weak, so we need absolute, repo-owned concept floors to guarantee minimal parser behavior.
- Provide a small, reviewable set of deterministic fixtures and a harness to fail CI on missing concepts or hard parse regressions.

### Description

- Added `xtask/src/tasks/parser_concept_floor.rs` implementing `cargo xtask parser-ratchet concept-floors` which scans `tests/perl-corpus/**/*.pl` + `*.meta.toml` and enforces required concept buckets and expected outcomes (`parse_clean`, `recover_without_panic`, `allow_errors`, `ast_shape`).
- Wired the new command into the xtask CLI (`xtask/src/main.rs`) and exported the task module in `xtask/src/tasks/mod.rs`.
- Added minimal repo-owned fixtures in `tests/perl-corpus/*.pl` with sidecar `*.meta.toml` describing `concepts`, `profile`, and `expected` values for initial buckets (regex, interpolation, heredoc, package, subroutine, lexical, references, hash_array_deref, pod, data_section, recovery).
- Emit a structured receipt (`concepts_required`, `concepts_hit`, `missing_concepts`, `violations`) and added a JSON schema at `.ci/receipts/schemas/parser-concept-floor.schema.json`; documentation for the command and fixture format was added at `docs/ci/perl-corpus-concepts.md`.

### Testing

- Ran `cargo xtask fmt` with no blocking issues.
- Ran `cargo check -p xtask` which completed successfully to validate build integration.
- Ran `cargo test -p xtask parser_concept_floor -- --nocapture` and both unit tests passed (`fails_when_required_concepts_missing` and `passes_when_all_required_concepts_present`).
- Exercised the new command end-to-end: `cargo xtask parser-ratchet concept-floors --manifest <subset>` correctly detected missing concepts (non-zero exit) and using a manifest listing all new fixtures produced a passing receipt (zero exit), validating detection/pass/fail scenarios.

Refs #6847

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69edd09fc4e88333a851e2af7b6b1321)

PR #6893, opened 2026-04-26
feat(parser-ratchet): select parser gate from scope and risk tags (#6847)

EffortlessSteven:
### Motivation

- Refs #6847 — Parser Ratchet must always emit a receipt but only run expensive parser corpus work when changes could plausibly affect parser behavior. This adds a scope-selection policy to make that decision deterministically.

### Description

- Add a dedicated scope policy file ` .ci/scope.d/parser-ratchet.toml` that declares `always_report = true`, receipt keys (`selected`, `selection_reason`, `reason`), parser-relevant `path_globs`, `risk_tags`, and a `cargo_lock_requires_parser_relevant_deps` caveat.
- Add JSON fixtures under `xtask/tests/fixtures/ci-scope/parser-ratchet/` covering requested scenarios: docs-only (no selection), parser crate change, lexer/token change, `ci_scope.rs` change, and workflow change.
- Add documentation `docs/ci/parser-ratchet-scope.md` describing that this PR implements scope-selection only and listing selector inputs, expected receipt shapes, examples that should no-op, and out-of-scope items.
- No production code behavior or gate execution was changed in this PR; scope selection policy, fixtures, and docs only.

### Testing

- Ran `cargo test -p xtask --test ci_scope_tests` and the test binary completed successfully (`7 passed`), verifying `ci-scope` CLI behavior and schema v2 contract.
- Validated all new fixture JSON files parse successfully using a small Python check (`python` JSON loader), confirming fixture integrity (5 fixtures loaded).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69edd0935b8c833390b358794836183f)

PR #6887, opened 2026-04-26
feat(generated): add generated-file ownership check (#6853)

EffortlessSteven:
### Motivation

- Prevent hand-edits of generated docs/status files from bypassing the generating tool by requiring a matching generator receipt for declared generated paths.
- Keep checks narrow to explicitly-declared paths so hand-authored forensics/docs are not blocked and generators are not run automatically.

### Description

- Add a manifest `.ci/generated-files.toml` and a JSON Schema `.ci/receipts/schemas/generated-files.schema.json` to declare owned generated paths and the expected generator command/owner. 
- Implement `cargo xtask generated-files` with `list` and `check` flows in `xtask/src/tasks/generated_files.rs`, including git-based changed-file detection, fixture-support (`--fixture`), generator-receipt ingestion (`--generator-receipt`), explicit override (`--allow-manual-edits`), and receipt emission (`verdict`, `changed_files`, `expected_command`, `missing_receipts`).
- Wire the command into the xtask CLI (`xtask/src/main.rs`, `xtask/src/tasks/mod.rs`) and add the `glob` dependency to `xtask/Cargo.toml` for pattern matching.
- Add policy documentation at `docs/ci/generated-file-policy.md` and fixture-backed integration tests plus fixtures at `xtask/tests/generated_files_cli.rs` and `xtask/tests/fixtures/generated-files/*.json`.

### Testing

- Ran `cargo check -p xtask` which completed successfully. 
- Ran the integration tests `cargo test -p xtask --test generated_files_cli` and both fixture cases (changed-without-receipt fails, changed-with-receipt passes) passed. 
- Exercised the CLI manually: `cargo xtask generated-files list` and `cargo xtask generated-files check --receipt target/receipts/generated-files.json` both succeeded and produced a receipt with `verdict: pass`. 
- Formatting and lint gates passed: `cargo xtask fmt` and `cargo clippy -p xtask -- -D warnings` completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69edd088a28083339e22afb3d2e8fec9)

PR #7016, opened 2026-04-27
fix(xtask): handle detached HEAD in PR runs (#7015)

EffortlessSteven:
## Summary

GitHub Actions PR jobs check out code by SHA (`actions/checkout` with `ref: sha`), putting git into detached HEAD state. Two functions in `xtask/src/tasks/gates.rs` were leaking git error messages to CI output on every PR run:

- **`is_latest_commit`**: called `git rev-parse --abbrev-ref --symbolic-full-name @{upstream}` without suppressing stderr. In detached HEAD, git prints `fatal: HEAD does not point to a branch` to stderr. The function already handled the `Err` case correctly (returning `true`), but the error leaked into CI logs.

- **`select_scope_base`**: iterated through ref candidates including `HEAD~1`, calling `git rev-parse --verify` on each without suppressing stderr. In a shallow PR clone (the merge-gate job uses depth 1 by default), `HEAD~1` doesn't exist and git prints `fatal: Needed a single revision` for each attempt — appearing 5 times per run.

Additionally, `collect_metadata` was recording `git_branch` as the literal string `"HEAD"` in detached HEAD state (the output of `git rev-parse --abbrev-ref HEAD` when detached), which is uninformative in receipts. The fix prefers `GITHUB_HEAD_REF` (PR source branch) or `GITHUB_REF_NAME` (any CI ref) when running in GitHub Actions.

## Root cause

```
# In detached HEAD, stderr leaks for both:
git rev-parse --abbrev-ref --symbolic-full-name @{upstream}  # → "fatal: HEAD does not point to a branch"
git rev-parse --verify HEAD~1                                 # → "fatal: Needed a single revision" (shallow clone)
```

## Fix

- Add `.stderr_null()` to both probing git calls — callers already handle `Err` gracefully
- In `collect_metadata`, prefer `GITHUB_HEAD_REF`/`GITHUB_REF_NAME` env vars before falling back to `git rev-parse --abbrev-ref HEAD`

## Test plan

- [ ] `cargo test -p xtask -- tasks::gates` passes (9 tests)
- [ ] `cargo xtask fmt --check` passes
- [ ] `cargo clippy -p xtask --bins` passes
- [ ] Detached HEAD smoke test: `git checkout --detach HEAD && cargo xtask gates --tier pr-fast --gate fmt` emits no "fatal:" lines


EffortlessSteven (commented):
## Combined fast-pass review: review-reviewed + maintainer-pr-reviewed + deep-reviewed

**Verdict: CLEAN across all three gates. Labels applied.**

### Standards check (review-reviewed)

- No banned patterns in new code. The two `unwrap_or_else(|_| ...)` in the fallback chain are the allowed form (not `.unwrap()`).
- `stderr_null()` is already established pattern in this codebase (used in `ci_scope.rs`, `dead_code.rs`, `targeted_checks.rs`).
- Both new `.stderr_null()` calls follow the same pattern as 5+ existing call sites.
- No scope drift: one file, three localized hunks, all in `xtask/src/tasks/gates.rs`.

### Maintainer fit (maintainer-pr-reviewed)

Surgical CI infrastructure fix with zero production behavior change. The `git_branch` field in receipts is purely diagnostic metadata — making it informative rather than "HEAD" in PR runs is a direct quality improvement. Unblocking #6893, #6887, #6907, #6872 via cascade is the right outcome for infrastructure PRs.

### Deep correctness (deep-reviewed)

**`is_latest_commit` (hunk 1):** `.stderr_null()` redirects stderr only; stdout capture via `.read()` is unaffected. The error path (`Err(_) => return true`) was already correct — `stderr_null()` makes it silent. No behavioral change on the success path.

**`select_scope_base` (hunk 2):** Same pattern. `.stderr_null()` suppresses the `"fatal: Needed a single revision"` noise when `HEAD~1` doesn't exist in a shallow clone. `.run().is_ok()` still correctly returns `false` on failure. The iteration continues to the next candidate. No regression.

**`collect_metadata` git_branch (hunk 3):** Env-var fallback chain analyzed across all invocation scenarios:
- PR run (`pull_request` event): `GITHUB_HEAD_REF` = source branch name (non-empty) → used. Correct.
- Push to master: `GITHUB_HEAD_REF` = `""` (filtered by `.filter(|s| !s.is_empty())`), `GITHUB_REF_NAME` = `"master"` → used. Correct.
- `merge_group` event: `GITHUB_HEAD_REF` = empty, `GITHUB_REF_NAME` = merge-group ref → used. Acceptable (better than `"HEAD"`).
- Local run: both absent → falls back to `git rev-parse --abbrev-ref HEAD`. On a branch = branch name; in local detached HEAD = `"HEAD"`. Acceptable.
- The `.filter(|s| !s.is_empty())` guards on both env vars prevent empty-string misuse.

**`git_branch` consumer check:** Field name unchanged. Only used in: struct definition (line 255), assignment (line 1191), display (line 1290), and three test JSON literals (lines 1737, 1799, 1865). All use the literal string `"work"` as test data — no consumer reads the env-var behavior. No breaking change.

**Test coverage:** Existing 9 tests continue to cover `failure_guidance`, `blocking_failure_gate_names`, `determine_overall_status`, and receipt round-trips. The new code in `collect_metadata` is stdlib-level `Option` chaining that is exercised by the builder's detached-HEAD smoke test. A unit test for the env-var chain would be achievable (set env var, call helper) but would require extracting the 3-line chain into a named function — acceptable as a follow-up issue, not a blocker.

**No issues found. No pushes needed. Labels applied.**

EffortlessSteven:
## CI Gate failure classification: INFRA_FLAKE / TRANSIENT

**Failed run:** [24988639043 / job 73169265798](https://github.com/EffortlessMetrics/perl-lsp/actions/runs/24988639043/job/73169265798) — head SHA `7c946194`

**Reported failures (from receipt):**
- `parser_corpus_ratchet` — `required: false` (NOT blocking)
- `cpan_corpus_ratchet` — `required: false` (NOT blocking)
- `lsp_tier_b` — `required: true`, single test `test_completion_gold_corpus` failed (15 passed / 1 failed in `editor_intelligence_scorecard`)

**Only `lsp_tier_b` is the blocking failure.** `receipt.summary.blocking_failures = ["lsp_tier_b"]`.

### Why this is a flake, not a PR bug

1. **PR diff scope is unrelated.** This PR only touches `xtask/src/tasks/gates.rs` (+21/-6). `xtask` is not a dependency of `perl-lsp-rs` tests, so the PR's diff cannot cause `test_completion_gold_corpus` to fail.

2. **Master at the PR's base SHA (52367fb27c) had `lsp_tier_b` passing.** Per [run 24979575958](https://github.com/EffortlessMetrics/perl-lsp/actions/runs/24979575958) on master 52367fb27c: `merge_gate: 15 passed, 2 failed, 1 skipped` — the 2 failing gates were `parser_corpus_ratchet` and `cpan_corpus_ratchet` (matching this PR), but `lsp_tier_b` PASSED. Same exact PR base.

3. **Variance between runs at the same head SHA.** The earlier run on this same SHA ([24951389268](https://github.com/EffortlessMetrics/perl-lsp/actions/runs/24951389268), 2026-04-26 08:10) reported `5 gate(s) failed: clippy_full, unit_full, parser_corpus_ratchet, cpan_corpus_ratchet, lsp_tier_b`. Different failure set on identical bytes = transient.

4. **`parser_corpus_ratchet` and `cpan_corpus_ratchet` are non-blocking** — they currently fail on green master runs too, so they have no bearing on CI Gate result.

### Action taken

- Re-running failed jobs on run 24988639043 via `gh run rerun --failed`. Status: queued.
- Applying `needs-ci-fix` to track that the gate is currently red while the rerun is in flight.
- Not applying `ci-green` (gate is not green yet).
- Not applying `needs-builder-fix` (the PR's diff is not the cause).

If the rerun stays red on `lsp_tier_b`, escalate to a separate investigation of `test_completion_gold_corpus` flakiness — file under "transient editor_intelligence_scorecard failures on PRs based on master 52367fb27c". The PR itself is innocent and should not be re-spun.

EffortlessSteven:
Promoted to merge-ready by orchestrator.

**Verification:**
- CI Gate (Merge-Blocking) on rerun = SUCCESS (run 24988639043 completed)
- Original failure was `lsp_tier_b/test_completion_gold_corpus` — confirmed INFRA_FLAKE (PR's xtask diff cannot reach that test path)
- Sign-off chain: review-reviewed + maintainer-pr-reviewed + deep-reviewed + diff-audited ✓

Labels: stripped `needs-ci-fix`, applied `ci-green` and `merge-ready`. Ready for ops merge once master fmt+test cascade fix #7031 lands and this PR is rebased.

PR #5544, opened 2026-04-23
fix(workspace): accept string workspace folder URIs (#0000)

EffortlessSteven:
### Motivation
- Some LSP clients send `workspaceFolders` entries as plain URI strings instead of object entries with a `uri` field, which caused those folders to be ignored during initialization.

### Description
- Update `extract_workspace_folder_uris` in `crates/perl-workspace-index/src/folder/mod.rs` to accept both `Value::String` entries and object entries with a `uri` field. 
- Extend the unit test `extracts_workspace_uris` to include a string-form folder URI to ensure mixed payloads are handled. 
- Extend the property-test input generator in `crates/perl-workspace-index/tests/workspace_folder_prop.rs` to produce string-form workspace folder entries for fuzz coverage.

### Testing
- Ran `cargo test -p perl-workspace` and the crate's unit/property tests passed. 
- Ran `cargo check --all-targets -p perl-workspace` which completed successfully. 
- Ran `cargo clippy -p perl-workspace --all-targets -- -D warnings` which failed on pre-existing lint issues in `crates/perl-workspace-index/tests/workspace_scorecard.rs` unrelated to this change. 
- Ran `cargo xtask fmt` which failed due to pre-existing `xtask` compile errors in `xtask/src/tasks/metrics/lsp_stats.rs` unrelated to this change. 
- `just pr-fast` was unavailable in this environment (`just: command not found`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69ea1fc8b9908333a3343ff7054dce83)

Issue #4901, opened 2026-04-23
fix + test + refactor: Codex wave 2026-04-23 late-PM (36 distinct items across SRP split, test coverage, fixes)


EffortlessSteven:
Omnibus tracker for the 36 winners from the 46-PR Codex batch of 2026-04-23 late-PM. 6 duplicates closed separately.

## Language SRP refactor
- #4843 — extract document-link handlers into SRP module
- #4846 — extract moniker handler into SRP module

## Devex (pr-fast improvements)
- #4847 — README heading structure guard in pr-fast
- #4848 — auto-detect default branch for local checks
- #4849 — ready preflight command
- #4850 — split nextest checks from base tool gate

## Diataxis docs
- #4851 — Diataxis authoring guide
- #4852 — docs routing guidance
- #4853 — documentation taxonomy clarification

## Test coverage (14 PRs)
- #4856 — LSP capability snapshot coverage (6 files)
- #4858 — parser snapshot coverage
- #4859 — LSP pull diagnostic snapshots
- #4860-#4863 — fuzz targets (declarations, symbol query, include-path, parser integration)
- #4864, #4867 — position-tracking proptests (ByteSpan + general)
- #4865 — perl-module path proptests
- #4866 — perl-dap watchpoint proptests
- #4868 — parser BDD given/when
- #4880 — lexer unicode identifier
- #4896 — stabilize workspace proptests

## Fix work
- #4881 — workspace-folder file URI fallback handling
- #4887 — workspace-folder fallback rootPath URI encoding
- #4882 — dap-platform quoted PATH normalization
- #4883 — lsp-runtime timed-out command termination
- #4884 — color deduplicate overlapping detections
- #4886 — dap stale inline helper removal
- #4888 — dap inline quote-like masking (tr/s/y/m ops)
- #4889 — workspace-index legacy separators in dependency matching (read-path; complements merged #4876's index-write path)
- #4892 — ci-hygiene multiline C raw string state tracking
- #4893 — ci-hygiene in_block_comment context
- #4894 — workspace qw extraction hardening
- #4898 — workspace-index `use constant +{...}` indexing

## Semantic
- #4891 — builtin regex/errno hash vars

## Also
- #4872 from the earlier batch (needs rebase; will land in this wave)

EffortlessSteven:
[...]

PR #7177, opened 2026-04-28
fix(workspace): accept string workspace folder URIs (#4901)

EffortlessSteven:
## Summary

- Extend `extract_workspace_folder_uris` to accept plain `Value::String` URI entries in addition to object entries with a `uri` or `path` field — matches real LSP client behavior where some clients send raw URI strings instead of `{"uri": "..."}` objects
- Preserve the existing `path`-object support from master (merged in parallel)
- Add string-form URI to unit test and property test generator for fuzz coverage

## Root Cause

The old PR #5544 was on a fresh-root strand (no common ancestor with master), causing all CI checks to fail. This is a salvage cherry-pick to a clean branch off current master.

## Files Changed

- `crates/perl-workspace-index/src/folder/mod.rs` — pattern-match on `Value::String` | `Value::Object` | other
- `crates/perl-workspace-index/tests/workspace_folder_prop.rs` — extend `uri_entry_strategy` with string-form variant

## Test plan

- [x] `cargo test -p perl-workspace --lib` — 253 passed
- [x] `cargo test -p perl-workspace -- folder` — property tests pass
- [x] `cargo clippy -p perl-workspace --lib -- -D warnings` — no issues
- [x] `cargo fmt --check -p perl-workspace` — clean
- [x] Verified pre-existing failures (`surface_workspace_parity` tests) exist on master, not introduced by this PR

Closes #5544 (stale strand, salvaged here)

### DIFF
--- a/crates/perl-workspace-index/src/folder/mod.rs
+++ b/crates/perl-workspace-index/src/folder/mod.rs
@@ -107,12 +107,14 @@ fn parse_file_uri_fallback(workspace_folder: &str) -> Option<PathBuf> {
 pub fn extract_workspace_folder_uris(workspace_folders: &[Value]) -> Vec<String> {
     workspace_folders
         .iter()
-        .filter_map(|folder| {
-            folder
+        .filter_map(|folder| match folder {
+            Value::String(uri) => Some(uri.clone()),
+            Value::Object(_) => folder
                 .get("uri")
                 .and_then(Value::as_str)
                 .map(std::string::ToString::to_string)
-                .or_else(|| folder.get("path").and_then(Value::as_str).map(root_path_to_file_uri))
+                .or_else(|| folder.get("path").and_then(Value::as_str).map(root_path_to_file_uri)),
+            _ => None,
         })
         .collect()
 }
@@ -221,10 +223,11 @@ mod tests {
             json!({"uri": "file:///one"}),
             json!({"uri": "file:///two"}),
             json!({"path": "/three"}),
+            json!("file:///four"),
             json!({"name": "invalid"}),
         ];
         let uris = extract_workspace_folder_uris(&entries);
-        assert_eq!(uris, vec!["file:///one", "file:///two", "file:///three"]);
+        assert_eq!(uris, vec!["file:///one", "file:///two", "file:///three", "file:///four"]);
     }
 
     #[test]
--- a/crates/perl-workspace-index/src/folder/mod.rs
+++ b/crates/perl-workspace-index/src/folder/mod.rs
@@ -231,6 +231,31 @@ mod tests {
     }
 
     #[test]
+    fn string_form_uri_passes_through_without_normalization() {
+        // Value::String arm passes the string through as-is, matching the behavior
+        // of the Value::Object{"uri": ...} arm which also does not normalize.
+        let entries = vec![json!("file:///a/b/c"), json!("file:///C:/Users/foo")];
+        let uris = extract_workspace_folder_uris(&entries);
+        assert_eq!(uris, vec!["file:///a/b/c", "file:///C:/Users/foo"]);
+    }
+
+    #[test]
+    fn non_file_and_non_object_entries_are_dropped() {
+        // Null, arrays, booleans, and numbers should all be silently skipped.
+        let entries = vec![json!(null), json!(42), json!(true), json!([])];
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
## Salvage note

This PR salvages #5544 (closed). The old branch had no common ancestor with current master (fresh-root strand) — all CI checks were failing because the base was a completely different git root, not because of any code issue.

**What changed vs the old PR**: The conflict resolution merged two concurrent evolutions of `extract_workspace_folder_uris`:
- This PR adds: `Value::String(uri)` case (bare string URI acceptance)
- Master already had: `{"path": "..."}` object form via `root_path_to_file_uri`

The merged implementation handles all three input shapes: string URI, object-with-uri, object-with-path.

**Prior reviews from #5544 carried over**: The implementation was reviewed by maintainer-pr and deep-reviewer on the old PR and found ALIGNED/clean. The fix logic is identical.

EffortlessSteven:
## Review-deep verdict: APPROVED with fix-forward additions

All 4 gates run. Labels applied: `review-reviewed`, `maintainer-pr-reviewed`, `diff-audited`, `deep-reviewed`.

---

### Gate 1 — review-reviewed (standards)

Clean. No banned patterns (`unwrap`, `expect`, `panic`, `todo`). Two-file scope. Formatting correct after `cargo fmt`. Clippy `-D warnings` clean. Commit title uses conventional commit format. The PR title `(#4901)` satisfies validate-title.

---

### Gate 2 — maintainer-pr-reviewed (project fit)

ALIGNED. This is a targeted compatibility fix for non-conformant LSP clients that send raw URI strings in `workspaceFolders`. The LSP 3.18 spec defines `WorkspaceFolder` as `{uri, name}` (always an object), but real-world clients (vim-lsp, some plugin authors) send plain strings. Accepting them at the parsing boundary — and letting the downstream path-resolution chain (`source_path_from_uri`) handle the rest — is exactly the right place to absorb client variance. The scope is minimal: 2 files, 1 crate, 1 function.

---

### Gate 3 — diff-audited (coherence)

CLEAN. The cumulative diff is 2 files, single crate `perl-workspace`, single function `extract_workspace_folder_uris`. No artifact directories (no `.spec/`, `.hermes/`). Pre-existing integration test failures (`surface_workspace_parity`) confirmed present on master — not introduced by this PR. The salvage cherry-pick framing is accurate.

---

### Gate 4 — deep-reviewed (correctness)

**Q: Does the `Value::String` handler validate URI shape or accept anything?**
No validation. The string is returned as-is, matching the existing `{"uri": "..."}` arm which also passes through un-validated. This is consistent: the function is a parser, not a validator. Downstream, `source_path_from_uri` already handles malformed URIs gracefully (returns `None` → folder has no path → no file scanning).

**Q: Does it normalize (e.g., Windows path conversion)?**
No normalization for the string arm — same as the `uri` key arm. Normalization only happens for the `path` key arm via `root_path_to_file_uri`. This is correct and consistent.

**Q: Edge cases: empty string, null, file:// without authority, mixed array**
- `json!(null)` → `Value::Null` → `_ => None` → dropped. Correct.
- `json!(42)`, `json!(true)`, `json!([])` → `_ => None` → dropped. Correct.
- `json!("")` → `Value::String("")` → `Some("".clone())` returned. Downstream: `source_path_from_uri("")` → `Path::new("").is_absolute()` = false → `None`. The folder gets a `uri=""` but `path=None`. Low severity — no files scanned. Consistent with `{"uri": ""}` behavior in master.
- `json!("file:/foo")` (single-slash) → returned as-is, downstream `Url::parse` handles it correctly.
- `json!("http://example.com")` → accepted by this layer, rejected downstream by `source_path_from_uri` (non-file scheme returns `None`). Harmless.

**Q: Does it preserve precedence with the legacy `path` key?**
Yes. The `uri > path` precedence within `Value::Object` entries is fully preserved in the `Value::Object(_)` arm. The string arm is orthogonal.

**Fix-forward additions pushed:** 3 unit tests pinning edge case behavior:
- `string_form_uri_passes_through_without_normalization` — documents the pass-through contract
- `non_file_and_non_object_entries_are_dropped` — pins null/number/bool/array all become `None`
- `object_uri_key_takes_precedence_over_path_key` — pins `uri > path` precedence in the object arm

256 tests pass (was 253 before the new edge case tests). Fmt clean, clippy clean.

Issue #4816, opened 2026-04-23
ci: local just pr-fast parity with Tier-1 scoped-deep lane


EffortlessSteven:
Parent: #4706. Related: ci-doctor. Developers running just pr-fast locally must get the same verdict Tier-1 draft CI gives on the same diff. Both should delegate to cargo xtask ci-scope --base origin/master for lane selection, invoke the same lane runner. Test by running on a known-failing PR locally — same failures surface. Integration test in xtask covers parity.

EffortlessSteven:
## Analysis: Local pr-fast Parity with Tier-1

**Status**: Ready for implementation

### Current state:
- `just pr-fast` exists locally
- `ci.yml` has `pr-smoke` job with ci-scope integration
- **Gap**: Local `just pr-fast` doesn't use ci-scope for lane selection

### Implementation path:

1. **Update `justfile`**:
   - Modify `pr-fast` recipe to call `cargo xtask ci-scope --base origin/master`
   - Use scope output for lane selection
   - Fall back to `clippy-core` / `test-core` on ci-scope failure

2. **Parity verification**:
   - Run on known-failing PR locally → same failures should surface
   - Integration test in xtask covers parity

3. **Current ci-scope integration in CI** (for reference):
   ```yaml
   - name: Compute CI scope
     id: scope
     run: |
       SCOPE_JSON=$(cargo xtask ci-scope --base origin/master --format json)
       # ... extracts direct_crates + architecture_wideners
   ```

### Related:
- Parent: #4706 (tiered CI)
- See `justfile` for current `pr-fast` implementation
- See `ci.yml` pr-smoke job for CI-side ci-scope usage

**Suggested assignee**: Good first issue — mostly justfile + xtask integration.

alceops:
I took a focused pass on #4816 and prepared a local `just pr-fast` parity patch: the recipe delegates to a script that computes `cargo xtask ci-scope --base origin/master --format json`, runs scoped clippy/lib tests/integration-test compile for selected crates, skips Rust lanes for prose/docs/CI-config diffs like PR Smoke, and falls back to the existing core lanes if scope detection fails. `bash -n` passes locally; full `just pr-fast` needs a Rust 1.92 toolchain.

Happy to open the focused PR if that shape matches the intended Tier-1/local parity path.

EffortlessSteven:
@alceops Yes, please open it.

[...]

PR #7181, opened 2026-04-29
ci: align pr-fast with ci-scope (#4816)

alceops:
Problem: local `just pr-fast` and the Tier-1 PR Smoke workflow had separate ci-scope implementations for #4816, so the two paths could drift.

Fix: move PR-fast planning into the shared `cargo xtask gates --tier pr-fast --receipt` runner. The runner now uses the existing Rust `ci-scope` classifier directly, includes direct crates + reverse-dep closure + architecture wideners in scoped package args, skips Rust build lanes for prose/docs/CI-config-only diffs, falls back to the core clippy/test baseline if scope detection fails, and emits the normal gate receipt at `target/receipts/receipt.json`. `just pr-fast` and PR Smoke both delegate to this same xtask runner.

Verification: `git diff --check` passed. Full compile/gate verification is deferred to CI because this cron runner has Cargo 1.75/no `just`, while this workspace requires the current edition2024 toolchain.

### DIFF
--- a/.ci/gate-policy.yaml
+++ b/.ci/gate-policy.yaml
@@ -131,6 +131,188 @@ gates:
     tags:
       - formatting
       - style
+    planning:
+      role: always_on
+
+  - name: check_conflict_markers
+    tier: pr_fast
+    description: "Prevent committed merge conflict markers"
+    required: true
+    command: just check-conflict-markers
+    timeout_seconds: 30
+    retry_count: 0
+    budgets:
+      max_duration_ms: 10000
+    quarantine: false
+    tags:
+      - hygiene
+      - conflict-markers
+    planning:
+      role: always_on
+
+  - name: release_history
+    tier: pr_fast
+    description: "Validate release-history surfaces"
+    required: true
+    command: just ci-release-history
+    timeout_seconds: 60
+    retry_count: 0
+    budgets:
+      max_duration_ms: 30000
+    quarantine: false
+    tags:
+      - release
+      - hygiene
+    planning:
+      role: always_on
+
+  - name: readme_heading_check
+    tier: pr_fast
+    description: "Prevent duplicate README level-2 headings"
+    required: true
+    command: just readme-heading-check
+    timeout_seconds: 30
+    retry_count: 0
+    budgets:
+      max_duration_ms: 10000
+    quarantine: false
+    tags:
+      - docs
+      - hygiene
+    planning:
+      role: always_on
+
+  - name: publish_closure
+    tier: pr_fast
+    description: "Check publish-closure transitive dependencies"
+    required: true
+    command: just ci-publish-closure
+    timeout_seconds: 60
+    retry_count: 0
+    budgets:
+      max_duration_ms: 30000
+    quarantine: false
+    tags:
+      - publish
+      - hygiene
+    planning:
+      role: always_on
+
+  - name: publish_manifest_check
+    tier: pr_fast
+    description: "Check publish manifest allowlist drift and license metadata"
+    required: true
+    command: just ci-publish-manifest-check
+    timeout_seconds: 60
+    retry_count: 0
+    budgets:
+      max_duration_ms: 30000
+    quarantine: false
+    tags:
+      - publish
+      - manifest
+    planning:
+      role: always_on
+
+  - name: layer_check
+    tier: pr_fast
+    description: "Enforce crate dependency layering"
+    required: true
+    command: just ci-layer-check
+    timeout_seconds: 60
+    retry_count: 0
+    budgets:
+      max_duration_ms: 30000
+    quarantine: false
+    tags:
+      - architecture
+      - hygiene
+    planning:
+      role: always_on
+
+  - name: published_crate_count_pr_fast
+    tier: pr_fast
+    description: "Ratchet published crate count"
+    required: true
+    command: just ci-published-crate-count
+    timeout_seconds: 30
+    retry_count: 0
+    budgets:
+      max_duration_ms: 10000
+    quarantine: false
+    tags:
+      - publish
+      - ratchet
+    planning:
+      role: always_on
+
+  - name: release_history_check
+    tier: pr_fast
+    description: "Check release-history drift"
+    required: true
+    command: just ci-release-history-check
+    timeout_seconds: 60
+    retry_count: 0
+    budgets:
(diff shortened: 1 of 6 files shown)


EffortlessSteven (changes requested):
Thanks for putting this together. The behavior you targeted is right: consult `ci-scope`, skip Rust lanes for non-Rust diffs, run scoped lanes for code diffs, fall back safely when scope detection fails. The implementation needs to move, though.

#4816 isn't asking local `just pr-fast` to copy current Tier-1 PR Smoke behavior. It's asking both to invoke the same lane runner. A new shell/Python script gives us a third CI implementation: local, Tier-1 YAML, and the script. All three can drift independently.

What we want instead:

```text
just pr-fast
  -> cargo xtask gates --tier pr-fast --base ${CI_SCOPE_BASE:-origin/master}

Tier-1 PR Smoke
  -> same xtask runner
```

For a revised PR:

1. Add or extend an `xtask` runner for the `pr-fast` tier.
2. Use the existing Rust `ci-scope` logic directly, not shell/Python JSON parsing.
3. Preserve the intended behavior: always run non-Rust quick checks; skip Rust lanes for `prose_only`, `docs_as_code`, and `ci_config`; run scoped clippy/test/check for code diffs; include direct crates, reverse-dep closure, and architecture wideners; fall back to core lanes when scope detection fails.
4. Make `just pr-fast` a thin delegate to that runner.
5. Emit the normal gate receipt to `target/receipts/receipt.json` using the existing schema so local and CI evidence are comparable.
6. Add `xtask` tests: prose/docs/CI-config diffs skip Rust lanes, code diffs select scoped lanes, reverse deps included, architecture wideners included, scope failure falls back cleanly, local and Tier-1 planning agree for the same fixture diff.

You don't need to redesign all of CI. Move the scoped `pr-fast` behavior into `xtask`, delegate to it from `justfile` and CI, and you're done.

The behavior this PR validated is useful. The mergeable version puts it behind the shared runner.

alceops:
Revised the PR around the shared-runner direction: removed the separate shell script, moved PR-fast scope planning into `cargo xtask gates --tier pr-fast --receipt`, changed `just pr-fast` and PR Smoke to delegate to that runner, and kept the scope behavior to ci-scope direct crates + reverse-dep closure + architecture wideners with core fallback on scope failure. Local verification here: `git diff --check` passes; full gate compile is left to CI because this runner has Cargo 1.75/no just while the repo requires the current edition2024 toolchain.

alceops:
Followed up on the latest review-bot findings in `36555de`: PR-fast receipt upload now matches the merge-gate artifact behavior with explicit `if-no-files-found: warn` and `retention-days: 7`, and `cargo xtask gates --tier pr-fast --list` now lists configured gates without invoking ci-scope planning. Local checks here: `git diff --check` plus static assertions; `cargo fmt --check` is unavailable in this runner because the installed Cargo lacks rustfmt.

EffortlessSteven:
Thanks for sticking with this.

Your revision got the key seam right: local `pr-fast` and PR Smoke on the same `xtask` runner instead of parallel shell/YAML logic. That's what #4816 needed.

I carried the branch through the repo-specific integration work:

- old local `pr-fast` had a broader quick-check envelope that had to move into gate policy rather than disappear
- local and CI needed an explicit shared base so they compare the same diff
- receipts needed to surface the selected/skipped/fallback plan in `target/receipts/receipt.json`
- planner tests needed to cover skip/scoped/fallback behavior
- scoped `--lib` lanes needed package target filtering to avoid selecting packages without lib targets
- your artifact-upload and `--list` follow-ups are preserved

For future CI/tooling PRs: read the existing xtask before picking an implementation shape. In a Rust repo, that's usually where the answer lives. The pattern generalizes: thin UX command, shared `xtask`/policy implementation, same local/CI evidence, focused planner tests.

Thanks for pushing this forward!

alceops:
Followed up on the latest review-bot pass in `c10eba75`:

- PR-fast now uses the actual pull request base (`origin/${{ github.base_ref }}`) and falls back to the default gate-runner base outside pull_request events.
- Removed the duplicate PR-fast release-history gate so `scripts/check_release_history.sh` only runs once in the shared gate plan.
- Rechecked the receipt schema concern against the current code: `agent_plan_receipt()` only serializes `name`, `role`, and `reason` for selected/skipped plan entries, so no `gate` payload is emitted and the schema stays unchanged.

Verification here: `git diff --check upstream/master..HEAD`, JSON schema parse, YAML parse for `.ci/gate-policy.yaml`/`.github/workflows/ci.yml`, and static assertions for the base-ref and duplicate release-history changes.

alceops:
Handled the latest review-bot pass in `580ce67`:

- kept `clippy_scoped` library-only (`cargo clippy --locked --lib {package_args}`), matching the fallback lane surface;
- stopped local `just pr-fast` from defaulting to `origin/master` by only passing `--base` when `CI_SCOPE_BASE` is explicit, so the xtask base selector can choose the appropriate repo base;
- expanded the xtask base selector fallback list to include `origin/main`, `origin/HEAD`, and local `main` for main-branch-only clones.

Verification here: `git diff --check`, YAML parse/static assertions, and extracted `pr-fast` shell syntax check passed. Full Cargo/rustfmt still needs the repo toolchain (this runner is Cargo 1.75 and has no cargo-fmt).

alceops:
Followed up on the latest CodeRabbit pass in 0f11240f8: PR-fast now fetches/uses the runtime base ref in CI; receipt upload has explicit retention/no-file behavior; the duplicate release-history PR-fast gate is removed; scoped clippy is library-only; non-Rust diff classification is centralized on ScopeOutput; the stale package-skip fallback path is removed; Windows shell execution now uses the same duct macro style. Verified locally: git diff --check, cargo fmt --check -p xtask, cargo check -p xtask.

EffortlessSteven:
Thanks, the library-only scoped clippy and local base-selection change are both right.

Taking the branch from here to finish the CI/review cleanup. Could you hold off on further pushes so we don't race each other or reset in-flight checks?

No action needed on your side.

EffortlessSteven (approved):
Approved.

This lands the intended #4816 shape: local `just pr-fast` and Tier-1 PR Smoke share the same `xtask` runner, use an explicit CI base, preserve the `pr_fast` gate envelope, and emit the normal receipt with plan details.

PR #7211, opened 2026-04-29
fix(perl-uri): canonicalize local file URIs in normalize_uri (#0000)

EffortlessSteven:
### Motivation
- `normalize_uri` returned already-parsed `file:` URLs unchanged which left local legacy forms like `file://localhost/...` or `file://C:/...` uncanonicalized and could produce inconsistent keys and comparisons.

### Description
- Route parsed `file` URLs through `uri_to_fs_path` followed by `fs_path_to_uri` before returning so local authorities normalize to the canonical `file:///` form while preserving non-local authorities and non-file schemes.
- Added a regression test `test_normalize_uri_canonicalizes_localhost_authority` to assert `normalize_uri("file://localhost/tmp/test.pl") == "file:///tmp/test.pl"`.
- Change is implemented in `crates/perl-uri/src/lib.rs` and keeps non-file URIs returned unchanged.

### Testing
- Ran `cargo test -p perl-uri` and all unit and doctests passed (all tests in the crate succeeded).
- The newly added regression test passed as part of the test run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c436ff48333af565e4a9cf1ba1a)

### DIFF
--- a/crates/perl-uri/src/lib.rs
+++ b/crates/perl-uri/src/lib.rs
@@ -278,6 +278,7 @@ pub fn normalize_uri(uri: &str) -> String {
         // forms like `file://C:/...` normalize to `file:///c:/...` on Windows
         // and `file:///tmp/...` on Unix while preserving non-local authorities.
         if url.scheme() == "file"
+            && url.host_str() == Some("localhost")
             && let Some(fs_path) = uri_to_fs_path(uri)
             && let Ok(normalized) = fs_path_to_uri(&fs_path)
         {
--- a/crates/perl-uri/src/lib.rs
+++ b/crates/perl-uri/src/lib.rs
@@ -274,7 +274,17 @@ pub fn normalize_uri(uri: &str) -> String {
 
     // Try to parse as URL first
     if let Ok(url) = Url::parse(uri) {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review fix applied (reviewer-deep)**

The `normalize_uri` change applied the `uri_to_fs_path` → `fs_path_to_uri` round-trip to ALL `file://` URIs, not just localhost ones. On Windows, `uri_to_fs_path("file:///tmp/test.pl")` successfully resolves via `windows_rooted_file_uri_to_path` to `H:\tmp\test.pl`, so `normalize_uri("file:///tmp/test.pl")` returned `"file:///H:/tmp/test.pl"` instead of the original — breaking 2 tests:
- `test_normalize_uri_valid`
- `test_normalize_uri_canonicalizes_localhost_authority`

Fix: add `&& url.host_str() == Some("localhost")` guard so the round-trip only fires for `file://localhost/...` URIs (the intended case), not canonical `file:///...` URIs. Fix pushed as commit e566f70d.

All tests pass (204/204). Clippy clean. Fmt clean.

PR #7212, opened 2026-04-29
fix(perl-uri): make special-scheme fallback case-insensitive (#0000)

EffortlessSteven:
### Motivation
- The fallback branch of `is_special_scheme` only matched lowercase prefixes for unparseable URIs, so uppercase variants (e.g. `UNTITLED:`) could be misclassified.

### Description
- Switch the fallback prefix checks in `is_special_scheme` to ASCII case-insensitive comparisons using `get(..).is_some_and(|p| p.eq_ignore_ascii_case(...))` to recognize prefixes regardless of case.
- Add a regression unit test `detects_special_schemes_case_insensitive_fallback` in `crates/perl-uri/src/classify.rs` to ensure uppercase invalid URIs are treated as special schemes.
- Change is confined to `crates/perl-uri/src/classify.rs` and does not alter other behavior.

### Testing
- Ran `cargo test -p perl-uri` and the entire test suite for the crate passed.
- The new test `detects_special_schemes_case_insensitive_fallback` executed and passed as part of the run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c436ff48333af565e4a9cf1ba1a)

### DIFF
--- a/crates/perl-uri/src/classify.rs
+++ b/crates/perl-uri/src/classify.rs
@@ -125,11 +125,11 @@ pub fn is_special_scheme(uri: &str) -> bool {
     if let Ok(url) = Url::parse(uri) {
         url.scheme() != "file"
     } else {
-        uri.starts_with("untitled:")
-            || uri.starts_with("git:")
-            || uri.starts_with("vscode-notebook:")
-            || uri.starts_with("vscode-notebook-cell:")
-            || uri.starts_with("vscode-vfs:")
+        uri.get(..9).is_some_and(|p| p.eq_ignore_ascii_case("untitled:"))
+            || uri.get(..4).is_some_and(|p| p.eq_ignore_ascii_case("git:"))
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review fix applied (reviewer-deep)**

Found off-by-one errors in the `is_special_scheme` fallback prefix bounds:
- `get(..17)` for `"vscode-notebook:"` (len=16) → corrected to `get(..16)`
- `get(..22)` for `"vscode-notebook-cell:"` (len=21) → corrected to `get(..21)`

Both errors cause the fallback to match one char beyond the scheme prefix, so equality checks against the exact scheme string would always fail for any input needing the fallback path.

The fallback is currently dead code (all tested inputs parse successfully via `Url::parse`), so existing tests still passed, but the logic was wrong. Fix pushed as commit 0ff44a20.

All tests pass (209/209). Clippy clean. Fmt clean.

PR #7213, opened 2026-04-29
fix(perl-workspace): reject unsafe git discovery paths (#0000)

EffortlessSteven:
### Motivation

- `git ls-files -z` can include absolute paths or `..` components which, when naively joined to the workspace root, allow traversal-style entries to escape the intended repository root and be treated as workspace files.

### Description

- Add `is_safe_relative_git_path` which rejects absolute paths and any path containing a `ParentDir` (`..`) component.
- Validate each NUL-delimited `git ls-files` entry with `is_safe_relative_git_path` in `parse_git_ls_files_output` and skip unsafe entries before joining with the workspace root.
- Import `std::path::Component` and add a unit test `parse_git_output_excludes_parent_directory_components` that verifies parent-directory entries are excluded while valid relative paths are accepted.

### Testing

- Ran `cargo test -p perl-workspace parse_git_output_excludes_parent_directory_components`, which passed.
- Ran `cargo check --all-targets -p perl-workspace`, which completed successfully.

------

### DIFF
--- a/crates/perl-workspace-index/src/discovery/mod.rs
+++ b/crates/perl-workspace-index/src/discovery/mod.rs
@@ -11,6 +11,7 @@ use crate::ignore::{is_skipped_dir_name, path_contains_skipped_component};
 use perl_parser_core::source_file::is_perl_source_path;
 use std::collections::HashSet;
 use std::ffi::OsString;
+use std::path::Component;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, Instant};
 use walkdir::{DirEntry, WalkDir};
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

**What was verified:**

- `is_safe_relative_git_path` correctly uses `Path::is_absolute()` and `Component::ParentDir` matching. The component-based check is the right abstraction — it handles both forward-slash and backslash separators on Windows correctly.
- `lib/../../etc/passwd` yields `ParentDir` components and is rejected correctly.
- The guard fires before `path_contains_skipped_component` so traversal paths don't get a second chance via the skip list.

**Fix-forward applied** (2 commits pushed directly):

1. Added `parse_git_output_excludes_absolute_paths` (Unix-only via `#[cfg(unix)]` — Windows root-relative `/`-prefixed paths are not considered absolute by `Path::is_absolute()` and have a different threat model).
2. Added `parse_git_output_excludes_embedded_parent_directory_traversal` — tests `lib/../../etc/passwd` which is the embedded `..` case not covered by the original test.

**No further changes needed.** The implementation is correct and the test coverage is now complete for the stated attack class.

PR #7214, opened 2026-04-29
fix(perl-lsp-rs): avoid executing binaries during command probe (#0000)

EffortlessSteven:
### Motivation
- On Unix the existing `command_exists` implementation launched `<command> --version`, which could execute an attacker-controlled binary when probing untrusted command names.

### Description
- Change Unix probing to call `which <command>` instead of running the candidate binary in `crates/perl-lsp-rs/src/execute_command/provider.rs`.
- Add a Unix-only security regression test `test_command_exists_does_not_execute_candidate_binary` in `crates/perl-lsp-rs/tests/execute_command_security_tests.rs` that places a fake executable on `PATH` and verifies discovery without execution.
- Update the test to call the public `command_exists` helper and add a `use std::path::Path` import.

### Testing
- Ran `cargo test -p perl-lsp-rs --test execute_command_security_tests test_command_exists_does_not_execute_candidate_binary` and it passed.
- The test verifies that the probe reports the candidate as discoverable and that no marker file was created, confirming the probe does not execute the candidate binary.

------

### DIFF
--- a/crates/perl-lsp-rs/src/execute_command/provider.rs
+++ b/crates/perl-lsp-rs/src/execute_command/provider.rs
@@ -954,8 +954,8 @@ pub fn command_exists(command: &str) -> bool {
         cmd.arg(command);
         cmd
     } else {
-        let mut cmd = std::process::Command::new(command);
-        cmd.arg("--version");
+        let mut cmd = std::process::Command::new("which");
+        cmd.arg(command);
         cmd
     };
     crate::util::run_command_with_timeout(cmd, 2)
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

**What was verified:**

- The standalone `command_exists` function previously ran `command --version`, which would execute the candidate binary. The fix correctly replaces this with `which <command>` / `where <command>` (Unix/Windows respectively), which only does PATH lookup without execution.
- The instance method `self.command_exists` already used `which`/`where` before this PR, so the fix brings the standalone function into alignment.
- All callers of both `command_exists` variants use hardcoded strings (`"yath"`, `"prove"`, `"perlcritic"`) — no user-controlled input reaches either function.
- The test creates a marker-file script and verifies it is NOT executed after `command_exists` is called on it. This is a correct behavioral test for the fix.

**Edge case — `which` availability:** `which` is universally available on Unix/macOS. `where.exe` is a built-in on Windows Server 2003+. No platform compatibility concern.

**No changes needed.** Logic is correct, test is sound, scope is tight.

PR #7215, opened 2026-04-29
fix(perl-subprocess-runtime): reject invalid command inputs (#0000)

EffortlessSteven:
### Motivation
- Prevent malformed or attacker-controlled command payloads (empty program names and NUL bytes) from reaching OS process creation APIs and causing unexpected behavior or potential injection vectors.

### Description
- Add `validate_command_input` and call it at the start of `OsSubprocessRuntime::run_command` to reject empty/whitespace-only program names and any NUL bytes in the program or arguments in `crates/perl-subprocess-runtime/src/lib.rs`.
- Add unit tests that assert empty program names and NUL-containing program/argument values are rejected.

### Testing
- Ran `cargo test -p perl-subprocess-runtime` which passed; `cargo check --all-targets -p perl-subprocess-runtime` and `cargo clippy -p perl-subprocess-runtime -- -D warnings` also passed.

------

### DIFF
--- a/crates/perl-subprocess-runtime/src/lib.rs
+++ b/crates/perl-subprocess-runtime/src/lib.rs
@@ -129,6 +129,8 @@ impl SubprocessRuntime for OsSubprocessRuntime {
         use std::io::Write;
         use std::process::{Command, Stdio};
 
+        validate_command_input(program, args)?;
+
         let (resolved_program, resolved_args) = resolve_command_invocation(program, args);
         let mut cmd = Command::new(&resolved_program);
         cmd.args(resolved_args.iter().map(String::as_str));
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

**What was verified:**

- `validate_command_input` is called as the first operation in `OsSubprocessRuntime::run_command`, before `resolve_command_invocation` or `Command::spawn`. The guard fires before any OS-level process creation.
- The `#[cfg(not(target_arch = "wasm32"))]` gate on `validate_command_input` is consistent with the impl block for `OsSubprocessRuntime` — the function can only be called from code that's already under the same cfg.
- NUL byte (`\0`) is the real POSIX injection vector: POSIX `execve` treats arguments as NUL-terminated C strings; a NUL embedded in a Rust `&str` argument can silently truncate the argument or cause undefined behavior at the C layer.
- The whitespace-only program name check correctly uses `.trim().is_empty()` — covers spaces, tabs, and other whitespace.

**Checked for bypass paths:**

- The `MockSubprocessRuntime` in tests does not call `validate_command_input`, which is intentional — mocks skip real validation. No production bypass exists via the mock.
- No other impl of `SubprocessRuntime` that could bypass the check was found.

**No changes needed.** The validation is placed at the right chokepoint and tests are complete.

PR #7216, opened 2026-04-29
fix(perl-dap): cap debugger identifiers to harden injection surface (#0000)

EffortlessSteven:
### Motivation

- The DAP handlers accepted arbitrarily long identifier strings for `setVariable` and `setFunctionBreakpoints`, which increases the attack surface for command-processing and resource-amplification from untrusted clients. 
- Limit identifier length to mitigate injection/amplification risks and reduce the surface area for malformed or adversarial inputs.

### Description

- Add a `const MAX_DEBUGGER_IDENTIFIER_LEN: usize = 512` to `crates/perl-dap/src/debug_adapter/mod.rs` to centralize the identifier length cap. 
- Enforce the length cap in `is_valid_set_variable_name` and `is_valid_function_breakpoint_name` by requiring `name.len() <= MAX_DEBUGGER_IDENTIFIER_LEN` in addition to existing regex checks. 
- Add a regression test `test_set_function_breakpoints_rejects_overlong_name` in `crates/perl-dap/tests/dap_non_regression_tests.rs` that verifies overlong function breakpoint names are rejected (unverified). 

### Testing

- Ran `cargo test -p perl-dap --test dap_non_regression_tests test_set_function_breakpoints_rejects_overlong_name`, and the targeted test passed (1 passed). 
- Ran `cargo check --all-targets -p perl-dap`, which completed successfully. 
- The change is limited to `crates/perl-dap` and the added test verifies the intended hardening behavior.

------

### DIFF
--- a/crates/perl-dap/src/debug_adapter/mod.rs
+++ b/crates/perl-dap/src/debug_adapter/mod.rs
@@ -111,6 +111,7 @@ const RECENT_OUTPUT_MAX_LINES: usize = 2048;
 const DEBUG_SESSION_TERMINATE_WAIT_MS: u64 = 250;
 const DEBUGGER_QUERY_WAIT_MS: u64 = 75;
 const DEBUGGER_FRAME_POLL_MS: u64 = 10;
+const MAX_DEBUGGER_IDENTIFIER_LEN: usize = 512;
 
 fn context_re() -> Option<&'static Regex> {
     CONTEXT_RE
@@ -362,7 +363,8 @@ fn set_variable_name_re() -> Option<&'static Regex> {
 
 /// Validate DAP setVariable names (e.g. `$x`, `%ENV`, `$Package::value`) for safe passthrough.
 fn is_valid_set_variable_name(name: &str) -> bool {
-    set_variable_name_re().is_some_and(|re| re.is_match(name))
+    name.len() <= MAX_DEBUGGER_IDENTIFIER_LEN
+        && set_variable_name_re().is_some_and(|re| re.is_match(name))
 }
 
 fn function_breakpoint_name_re() -> Option<&'static Regex> {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

**What was verified:**

- `is_valid_set_variable_name` and `is_valid_function_breakpoint_name` are defined in `debug_adapter/mod.rs` but ARE called in the real request handlers: `breakpoints.rs` (line 154 — function breakpoints) and `variables.rs` (lines 279 and 279 — set variable). Both use `use super::*` so `MAX_DEBUGGER_IDENTIFIER_LEN` is visible.
- The length check short-circuits before the regex evaluation. This prevents potential ReDoS from the `(?:::[A-Za-z_][A-Za-z0-9_]*)*` repeated group in `SET_VARIABLE_NAME_RE` on pathologically long inputs.
- 512 bytes is appropriate for Perl identifiers: a valid qualified name like `Foo::Bar::Baz::...::X` at 512 chars allows ~170 package components. No legitimate debugger identifier approaches that length.
- The test `test_set_function_breakpoints_rejects_overlong_name` correctly exercises the end-to-end path through `handle_set_function_breakpoints` → `is_valid_function_breakpoint_name`.

**Pre-existing gap identified (out of scope for this PR — file as follow-up):**

The `setVariable` handler validates `name` via `is_valid_set_variable_name` but the `value` field only has a newline check. A value like `1 ; x /etc/passwd` could inject semicolon-separated commands into the Perl debugger's `p $name = $value` command sequence. The correct fix is to apply an expression validator similar to `validate_safe_expression` used in the `evaluate` handler. This requires a separate PR targeting `debug_adapter/variables.rs`.

**No blocking issues.** This PR correctly hardens the identifier surface. The value injection gap is pre-existing and tracked as a follow-up.

PR #7228, opened 2026-04-29
fix(perl-lsp-perltidy): clamp zero timeout for OS runtime (#0000)

EffortlessSteven:
### Motivation
- Prevent the language server from panicking when a client supplies `timeout_secs = 0` because `OsSubprocessRuntime::with_timeout(0)` panics; the change ensures misconfigured clients cannot crash the process.

### Description
- Clamp `config.timeout_secs` to at least `1` in `PerlTidyFormatter::with_os_runtime` and add a non-WASM regression test `with_os_runtime_clamps_zero_timeout` that constructs a formatter with `timeout_secs: 0`.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-perltidy` and it passed.
- Ran `cargo clippy -p perl-lsp-perltidy -- -D warnings` and it passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c70cfd08333a09bf28f4d4352aa)

### DIFF
--- a/crates/perl-lsp-perltidy/src/lib.rs
+++ b/crates/perl-lsp-perltidy/src/lib.rs
@@ -197,7 +197,9 @@ impl PerlTidyFormatter {
     #[must_use]
     pub fn with_os_runtime(config: PerlTidyConfig) -> Self {
         use perl_subprocess_runtime::OsSubprocessRuntime;
-        let timeout = config.timeout_secs;
+        // OsSubprocessRuntime::with_timeout panics on zero; clamp to 1s so
+        // misconfigured clients do not crash the language server process.
+        let timeout = config.timeout_secs.max(1);
         Self::new(config, Arc::new(OsSubprocessRuntime::with_timeout(timeout)))
     }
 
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

**Logic:** CORRECT. `config.timeout_secs.max(1)` correctly prevents the `assert!(timeout_secs > 0)` panic in `OsSubprocessRuntime::with_timeout`. The `u64::max(1)` operation is a no-op for any caller-supplied value >= 1. The comment added explains WHY the clamp exists, which is important for future maintainers.

**Tests:** IMPROVED. Added a companion boundary test `with_os_runtime_accepts_minimum_valid_timeout` (timeout_secs=1) to anchor the invariant: any value >= 1 must construct without panic. This documents the boundary explicitly.

**Regression risk:** LOW. The only callers of `with_os_runtime` are in the LSP server initialization path. A well-configured client will have timeout_secs >= 1, so the clamp is a no-op for them.

**Security:** CLEAN. No shell injection surface; the timeout is a numeric bound only.

**Edge cases:** The `u64::MAX` case continues to work (large timeout simply means perltidy runs until it exits naturally). Pre-existing theoretical overflow in `Instant::now() + Duration::from_secs(u64::MAX)` is out of scope for this PR.

PR #7229, opened 2026-04-29
fix(perl-lsp-perltidy): preserve extra args when profile is set (#0000)

EffortlessSteven:
### Motivation
- `PerlTidyConfig::to_args()` dropped `extra_args` whenever a `profile` was configured, preventing users from combining a `.perltidyrc` profile with explicit CLI flags.

### Description
- Append `extra_args` when `profile` is set so `to_args()` emits `--profile=...` and still includes any explicit `extra_args`.
- Update unit tests in `crates/perl-lsp-perltidy/tests/config_tests.rs` to assert the profile flag is present and that explicit `extra_args` are preserved while other derived flags remain suppressed.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-perltidy` and it completed successfully.
- Ran `cargo clippy -p perl-lsp-perltidy -- -D warnings` and there were no lints reported.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c70cfd08333a09bf28f4d4352aa)

### DIFF
--- a/crates/perl-lsp-perltidy/src/lib.rs
+++ b/crates/perl-lsp-perltidy/src/lib.rs
@@ -110,6 +110,7 @@ impl PerlTidyConfig {
 
         if let Some(profile) = &self.profile {
             args.push(format!("--profile={profile}"));
+            args.extend(self.extra_args.clone());
             return args;
         }
 
--- a/crates/perl-lsp-perltidy/tests/config_tests.rs
+++ b/crates/perl-lsp-perltidy/tests/config_tests.rs
@@ -53,6 +53,19 @@ fn config_with_profile_ignores_other_settings_but_keeps_extra_args() {
 }
 
 #[test]
+fn config_with_profile_and_no_extra_args_produces_only_profile_flag() {
+    // Profile without extra_args should still yield exactly one arg.
+    let config = PerlTidyConfig {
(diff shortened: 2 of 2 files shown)


EffortlessSteven:

**Logic:** CORRECT. The fix `args.extend(self.extra_args.clone())` correctly appends extra_args after the `--profile=...` flag. When `extra_args` is empty (the common case), `extend` is a no-op, so the fix is backward-compatible. The one-line production change is minimal and well-placed.

**Tests:** IMPROVED. The renamed tests (`config_with_profile_uses_profile_flag_and_extra_args`, `config_with_profile_ignores_other_settings_but_keeps_extra_args`) now accurately reflect the new semantics. Added boundary test `config_with_profile_and_no_extra_args_produces_only_profile_flag` to confirm the zero-extra-args case still produces exactly one arg. This prevents the "test passes vacuously" failure mode if a future change broke the empty-case.

**Regression risk:** LOW. The only behavior change is that extra_args now survive when profile is set — previously they were silently dropped. No callers of `to_args()` that previously relied on extra_args being dropped.

**Note on #7233:** PR #7233 makes the identical production code change. It keeps old test names that are now misleading after the behavioral change. #7229 is the preferred implementation. #7233 should be closed as a duplicate once #7229 merges.

PR #7230, opened 2026-04-29
fix(perl-lsp-perltidy): preserve trailing newline behavior in builtin formatter (#0000)

EffortlessSteven:
### Motivation
- The built-in Rust fallback formatter always appended a trailing newline, which introduced spurious EOF diffs when `perltidy` is unavailable.
- Preserve the original document shape by keeping the presence or absence of a trailing newline unchanged by the fallback formatter.

### Description
- Update `BuiltInFormatter::format` to collect `lines`, detect `had_trailing_newline`, and only append a newline for the final output line when the input originally ended with one.
- Add a regression test `builtin_formatter_preserves_missing_trailing_newline` in `crates/perl-lsp-perltidy/tests/config_tests.rs` to assert newline-less input remains newline-less after formatting.
- Change is isolated to the perltidy fallback formatter and its unit tests with no public API changes.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-perltidy` and it passed.
- Ran `cargo clippy -p perl-lsp-perltidy --all-targets -- -D warnings` and it passed.
- Ran `cargo xtask fmt` to ensure formatting, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c78b3508333ab44296416f4c9c2)

### DIFF
--- a/crates/perl-lsp-perltidy/src/lib.rs
+++ b/crates/perl-lsp-perltidy/src/lib.rs
@@ -352,13 +352,15 @@ impl BuiltInFormatter {
     pub fn format(&self, code: &str) -> String {
         let mut result = String::new();
         let mut indent_level: i32 = 0;
+        let lines: Vec<&str> = code.lines().collect();
+        let had_trailing_newline = code.ends_with('\n');
         let indent_str = if self.config.tabs.unwrap_or(false) {
             "\t".to_string()
         } else {
             " ".repeat(self.config.indent_columns.unwrap_or(4) as usize)
         };
 
-        for line in code.lines() {
+        for (index, line) in lines.iter().enumerate() {
             let trimmed = line.trim();
             let leading_closers = count_leading_closers(trimmed) as i32;
             indent_level = indent_level.saturating_sub(leading_closers);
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

**Logic:** CORRECT. The fix correctly detects trailing newlines before iterating and conditionally appends `\n` only when it's not the last line OR when the original had a trailing newline. All edge cases check out:
- Empty input `""`: `lines()` empty, result stays empty. Correct.
- Single `"\n"` only: `lines()` yields one empty string, `had_trailing_newline=true`, loop pushes `\n`. Correct.
- `"a\nb"` (no trailing newline): last line doesn't get `\n`. Correct.
- `"a\n"` (with trailing newline): last line gets `\n`. Correct.
- Multi-line with blank lines: `"\n\n"` yields two empty strings, both get `\n`. Correct.

The use of `index + 1 == lines.len()` is the right idiom since `lines.len()` is precomputed once.

**Tests:** IMPROVED. Added four boundary tests:
- `builtin_formatter_preserves_trailing_newline_when_present` — verifies input with `\n` produces output with `\n`
- `builtin_formatter_single_line_no_trailing_newline` — verifies single-line input without `\n` stays clean
- `builtin_formatter_single_line_with_trailing_newline` — verifies round-trip for single-line + `\n`

**Regression risk:** LOW. Affects `BuiltInFormatter::format` only, which is the fallback used when perltidy is not installed.

**Performance:** CLEAN. One extra boolean (`had_trailing_newline`) added before the loop. Allocation pattern unchanged.

PR #7233, opened 2026-04-29
fix(perl-lsp-perltidy): keep extra args with profile mode (#0000)

EffortlessSteven:
### Motivation

- `PerlTidyConfig::to_args()` returned early when `profile` was set, which silently dropped `extra_args` and prevented callers from forwarding additional perltidy flags.

### Description

- Append `extra_args` when `profile` is present by calling `args.extend(self.extra_args.clone())` before returning in `to_args()`.
- Add a focused test `config_with_profile_keeps_extra_args` and clarify an existing test comment to cover the profile + extra args behavior.

### Testing

- Ran `cargo test -p perl-lsp-perltidy`, and all tests passed (unit/integration tests for the crate completed successfully).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c78b3508333ab44296416f4c9c2)

### DIFF
--- a/crates/perl-lsp-perltidy/src/lib.rs
+++ b/crates/perl-lsp-perltidy/src/lib.rs
@@ -110,6 +110,7 @@ impl PerlTidyConfig {
 
         if let Some(profile) = &self.profile {
             args.push(format!("--profile={profile}"));
+            args.extend(self.extra_args.clone());
             return args;
         }
 
--- a/crates/perl-lsp-perltidy/tests/config_tests.rs
+++ b/crates/perl-lsp-perltidy/tests/config_tests.rs
@@ -43,9 +43,23 @@ fn config_with_profile_ignores_other_settings() {
     };
     let args = config.to_args();
 
-    // Only profile flag should be present; all others suppressed
+    // Only profile flag should be present from primary settings; all others suppressed
     assert_eq!(args.len(), 1);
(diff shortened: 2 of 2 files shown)


EffortlessSteven:
## Deep Review — DUPLICATE (prefer #7229)

This PR makes the same production code change to `to_args()` in `lib.rs` as #7229: adding `args.extend(self.extra_args.clone())` in the profile branch.

**Status:** #7229 is the preferred implementation because:
1. It renames the existing tests to accurately reflect the new behavior (`config_with_profile_uses_profile_flag_and_extra_args` vs the misleading old name `config_with_profile_uses_only_profile_flag`).
2. #7229 includes a boundary test for profile-with-no-extra-args to prevent vacuous assertions.
3. This PR keeps the old test names (`config_with_profile_uses_only_profile_flag`, `config_with_profile_ignores_other_settings`) which will still pass (since the default empty `extra_args` means `extend` is a no-op), but their names no longer reflect the contract — a future reader would assume extra_args are always suppressed with profile, which is now false.

**Recommendation:** Close this PR in favor of #7229.

No `deep-reviewed` label applied — this PR is a duplicate.

EffortlessSteven:
Closing as duplicate of #7229 (keep extra args with profile mode is the same fix as #7229's preserve extra args).

PR #7234, opened 2026-04-29
fix(vscode-extension): resolve relative perlcritic profiles in health check (#0000)

EffortlessSteven:
### Motivation

- The onboarding health check treated relative `perl-lsp.perlcritic.profile` values as current-working-directory paths, which caused valid workspace-relative profiles to produce false “profile not found” warnings.

### Description

- Resolve non-absolute Perl::Critic profile paths against the primary workspace root via a new `resolvePerlcriticProfilePath` helper and use the resolved path for existence checks in `checkPerlcriticSetup` (`vscode-extension/src/onboarding.ts`).
- Update the health-check warning message to include the resolved profile path when a profile is not found.
- Add a regression unit test that exercises workspace-relative profile resolution by populating `vscode.workspace.workspaceFolders` and mocking the extension configuration (`vscode-extension/src/test/onboarding.test.ts`).
- Restore mocked state after the test to avoid cross-test leakage.

### Testing

- Ran `npm test -- onboarding.test.ts` in `vscode-extension`, which executed the modified test file but the run failed due to repository TypeScript/Jest global type configuration errors (multiple `TS2304: Cannot find name 'jest'` / test-runner globals), not due to the new test logic.
- No Rust or Cargo tests were modified or executed as part of this change.

------

### DIFF
--- a/vscode-extension/src/onboarding.ts
+++ b/vscode-extension/src/onboarding.ts
@@ -263,7 +263,11 @@ export class OnboardingManager {
     }
 
     const primaryWorkspace = vscode.workspace.workspaceFolders?.[0]?.uri.fsPath;
-    return primaryWorkspace ? path.resolve(primaryWorkspace, profile) : profile;
+    // When no workspace folder is open we cannot resolve a relative path to a
+    // meaningful absolute location (CWD of the extension host is not the user's
+    // project directory).  Return `undefined` so the caller skips the existence
+    // check rather than silently probing the wrong location.
+    return primaryWorkspace ? path.resolve(primaryWorkspace, profile) : undefined;
   }
 
   /**
--- a/vscode-extension/src/onboarding.ts
+++ b/vscode-extension/src/onboarding.ts
@@ -8,6 +8,7 @@
  */
 
 import * as fs from 'fs';
+import * as path from 'path';
 import { execFile } from 'child_process';
 import * as vscode from 'vscode';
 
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

**Verdict: Approved with bug fix pushed.**

### Bug found and fixed

`resolvePerlcriticProfilePath` had a correctness bug when no workspace folder is open:

```typescript
// BEFORE (buggy)
return primaryWorkspace ? path.resolve(primaryWorkspace, profile) : profile;
```

When `primaryWorkspace` is `undefined`, the function returned the raw relative string `profile` (e.g. `"config/perlcriticrc"`). The caller then called `fs.existsSync("config/perlcriticrc")`, which resolves relative to the **extension host's CWD** — not the user's project directory. This produces either false-positive "profile not found" warnings or false-negatives depending on what happens to be in CWD.

**Fix:** Return `undefined` when no workspace folder is available, causing the caller to skip the existence check entirely.

```typescript
// AFTER (correct)
return primaryWorkspace ? path.resolve(primaryWorkspace, profile) : undefined;
```

### Additional edge case in `~` handling

This PR does NOT handle `~/` tilde expansion. If a user has both a relative path and no workspace, the fix-forward is now correct. However, if a user configures `profile = "~/.perlcriticrc"` and PR #7236 is not merged alongside this PR, the `~/` prefix will be treated as a relative path, resolved to `undefined` (with this fix) or incorrectly to `path.resolve(workspace, "~/.perlcriticrc")` (a wrong path). **These two PRs must be merged together or sequenced** — #7234 first, then #7236 rebased on top.

### What was verified

- Added regression test for no-workspace case: verifies no "profile not found" warning when `workspaceFolders` is `undefined` and profile is relative.
- Absolute paths still return directly without workspace lookup.
- Relative paths with workspace resolve correctly (existing test).

### Title issue

PR title contains `(#0000)` — needs a real issue number before merge.

### Merge ordering note

This PR and PR #7236 both modify `vscode-extension/src/onboarding.ts`. They must be merged sequentially (#7234 first, then #7236 rebased on top) or combined into one PR.

PR #7236, opened 2026-04-29
fix(vscode-extension): support ~/ perlcritic profiles in onboarding (#0000)

EffortlessSteven:
### Motivation
- The onboarding health check treated configured `perlcritic.profile` paths literally so `~/...` home-relative profiles were reported as missing even when the file existed, causing a poor UX when using home-shorthand paths.

### Description
- Import `os` and `path` and add a `resolveUserPath` helper to expand `~` and `~/` into the user home directory and trim whitespace.
- Use `resolveUserPath` when reading `perlcritic.profile` in `checkPerlcriticSetup` so filesystem checks operate on normalized paths.
- Add a regression unit test `OnboardingManager.checkPerlcriticSetup` that writes a temp profile to `~` and verifies `checkPerlcriticSetup` accepts `~/...` profiles (file: `vscode-extension/src/test/onboarding.test.ts`).

### Testing
- Ran the extension test command `npm test -- onboarding.test.ts`, which failed due to TypeScript/Jest global-type resolution errors (unresolved `jest`, `describe`, `expect`) that are pre-existing in the workspace, so the automated test run did not complete successfully.
- The change is limited to `vscode-extension/src/onboarding.ts` and its test and is small and self-contained; the added regression test should pass in a correctly-configured Jest/TypeScript test environment.

------

### DIFF
--- a/vscode-extension/src/onboarding.ts
+++ b/vscode-extension/src/onboarding.ts
@@ -8,6 +8,7 @@
  */
 
 import * as fs from 'fs';
+import * as os from 'os';
 import * as path from 'path';
 import { execFile } from 'child_process';
 import * as vscode from 'vscode';
@@ -96,6 +97,19 @@ type ExecCheckFn = (
 interface ExecInvocation {
   command: string;
   args: string[];
+}
+
+function resolveUserPath(rawPath: string): string {
+  const trimmedPath = rawPath.trim();
+  if (trimmedPath === '~') {
+    return os.homedir();
+  }
+
+  if (trimmedPath.startsWith('~/') || trimmedPath.startsWith('~\\')) {
+    return path.join(os.homedir(), trimmedPath.slice(2));
+  }
+
+  return trimmedPath;
 }
 
 // ---------------------------------------------------------------------------
@@ -211,7 +225,10 @@ export class OnboardingManager {
     const label = 'perlcritic';
     const perlcriticConfig = vscode.workspace.getConfiguration('perl-lsp').get<any>('perlcritic', {});
     const enabled = Boolean(perlcriticConfig?.enabled);
-    const profile = typeof perlcriticConfig?.profile === 'string' ? perlcriticConfig.profile.trim() : '';
+    const profile =
+      typeof perlcriticConfig?.profile === 'string'
+        ? resolveUserPath(perlcriticConfig.profile)
+        : '';
 
     if (!enabled) {
       return {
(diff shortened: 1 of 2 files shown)


EffortlessSteven:

**Verdict: Approved with test hardening pushed.**

### What was verified

`resolveUserPath` logic:
- `"~"` alone → `os.homedir()` ✓
- `"~/foo"` → `path.join(os.homedir(), "foo")` ✓
- `"~\foo"` (Windows) → `path.join(os.homedir(), "foo")` ✓
- Plain string with whitespace → trimmed and returned as-is ✓

The `checkPerlcriticSetup` integration: profile is resolved via `resolveUserPath` (replacing `.trim()`) before being passed to `fs.existsSync`. This means `~/...` paths are correctly expanded to absolute paths and existence-checked correctly.

### Missing interaction with PR #7234

This PR does NOT handle workspace-relative profile paths — it only handles `~/`. PR #7234 adds workspace-relative resolution. Neither PR handles both cases together. The correct complete solution needs both PRs combined or sequenced:

1. `~/` tilde → expand to home dir (this PR)
2. Relative path → resolve against workspace root (PR #7234)

These two PRs both modify the same lines in `checkPerlcriticSetup`. **They must be sequenced or combined before merge** to avoid a merge conflict that silently drops one fix.

### Bug in PR #7236: the `~/` branch doesn't also handle workspace-relative

After `resolveUserPath` expands `"~/.perlcriticrc"` to an absolute path, the profile is checked with `fs.existsSync(profile)` (not `resolvedProfile` as in #7234). This is correct for tilde-expanded paths since they're now absolute. But if a user has a plain relative path (no `~`), `resolveUserPath` just trims it and returns it as-is — then `fs.existsSync("config/perlcriticrc")` still probes CWD wrongly. This is PR #7234's bug, not #7236's.

### Test hardening pushed

The original test used a fixed file name `.perlcritic-onboarding-test.rc` in `os.homedir()`. In CI with parallel shards, two shards could try to write/delete the same file simultaneously.

**Fix:** Use `Date.now() + process.pid` for a unique file name per test run. Also guard `rmSync` with a `profileWritten` flag so cleanup is skipped if the initial write failed (e.g., read-only home dir in restricted CI).

### Title issue

PR title contains `(#0000)` — needs a real issue number before merge.

### Merge ordering note

This PR and PR #7234 both modify `vscode-extension/src/onboarding.ts` at overlapping lines. They must be merged sequentially or combined. Recommended order: #7234 first (relative path fix), then #7236 (tilde expansion) rebased on top.

PR #7237, opened 2026-04-29
fix(perlcritic): use doc-scoped workspace root for profile resolution (#0000)

EffortlessSteven:
### Motivation
- Perl::Critic profile discovery could pick the global `root_path` in multi-root workspaces, causing incorrect relative profile resolution and .perlcriticrc discovery for documents living outside the primary folder.

### Description
- Add `workspace_root_for_doc` helper in `crates/perl-lsp-rs/src/runtime/diagnostics.rs` and use it instead of directly reading `root_path` when resolving configured profiles and walking up for `.perlcriticrc` in both the pull and push perlcritic diagnostic code paths.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-rs` which completed successfully (with unrelated compile-time warnings from test helpers).

------

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/diagnostics.rs
+++ b/crates/perl-lsp-rs/src/runtime/diagnostics.rs
@@ -34,6 +34,14 @@ fn resolve_configured_profile_path(
 }
 
 #[cfg(not(target_arch = "wasm32"))]
+fn workspace_root_for_doc(server: &LspServer, uri: &str) -> Option<std::path::PathBuf> {
+    server
+        .folder_for_doc_uri(uri)
+        .and_then(|folder| folder.path.or_else(|| source_path_from_uri(&folder.uri)))
+        .or_else(|| server.root_path.lock().clone())
+}
+
+#[cfg(not(target_arch = "wasm32"))]
 fn find_workspace_perlcritic_profile(
     workspace_root: Option<&std::path::Path>,
     file_path: &std::path::Path,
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

**Verdict: Approved.** Logic is correct. Pushed one improvement commit.

### What was verified

- `workspace_root_for_doc` helper correctly falls back to `server.root_path.lock().clone()` when no folder matches.
- `folder_for_doc_uri` + `source_path_from_uri` are both non-wasm32-gated, so calling them inside the `#[cfg(not(target_arch = "wasm32"))]` helper is valid.
- `build_context` inlines the same logic rather than calling the helper — this is intentional: `build_context` must compile on all targets including wasm32, while the helper only needs to exist on native targets. Previously this was undocumented and could be mistaken for an oversight.
- Both new tests (`build_context_uses_doc_scoped_workspace_root` and `build_context_falls_back_to_root_path_when_no_folder_matches`) pass locally (2 passed, 2830 filtered out).
- `analyze_file_with_hash` and `hash_content` APIs exist and are properly re-exported from `perl-lsp-rs-core::tooling::perl_critic`.

### Fix pushed

Added a comment in `build_context` explaining why the workspace root logic is inlined rather than delegated to `workspace_root_for_doc`. Prevents future readers from mistaking this for a missed refactor.

### Title issue

PR title contains `(#0000)` — a placeholder that will fail `validate-title` CI. Needs to be updated with a real issue number before merge.

PR #7238, opened 2026-04-29
fix(perl-lsp-rs-core): ignore commented use pragmas in built-in critic (#0000)

EffortlessSteven:
### Motivation

- The built-in Perl::Critic replacement was treating commented `use strict`/`use warnings` lines as real pragmas and thus suppressed real violations.
- Ensure the built-in analyzer only recognizes uncommented pragma directives so diagnostics reflect actual code, not comments.

### Description

- Replaced the naive `content.contains("use {feature}")` check with a new `has_use_statement(content, feature)` helper that inspects only the uncommented portion of each line by splitting on `#` before searching.
- Updated `missing_use_statement_violation` to use `has_use_statement` for correct pragma detection.
- Added two focused unit tests `builtin_analyzer_flags_missing_strict_when_only_comment_mentions_it` and `builtin_analyzer_flags_missing_warnings_when_only_comment_mentions_it` to prevent regressions.
- Changes are in `crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs` and are limited to the built-in critic policy logic and tests.

### Testing

- Ran `cargo test -p perl-lsp-rs-core builtin_analyzer_flags_missing_` and the new regression tests passed (2 passed; 0 failed).
- Full crate test run completed with no failures for the modified crate's unit tests during the session.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c84f08c8333a3d1ba07ffd487e8)

### DIFF
--- a/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
+++ b/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
@@ -191,7 +191,7 @@ fn missing_use_statement_violation(
     feature: &str,
     explanation: &str,
 ) -> Vec<Violation> {
-    if content.contains(&format!("use {feature}")) {
+    if has_use_statement(content, feature) {
         return Vec::new();
     }
 
@@ -203,6 +203,15 @@ fn missing_use_statement_violation(
         range: insertion_range(),
         file: String::new(),
     }]
+}
+
+
+fn has_use_statement(content: &str, feature: &str) -> bool {
+    let needle = format!("use {feature}");
+    content.lines().any(|line| {
+        let code = line.split('#').next().unwrap_or_default();
+        code.contains(&needle)
+    })
 }
 
 fn find_bareword_open_filehandles(content: &str) -> Vec<Range> {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review: closing as inferior duplicate of #7241**

Both PRs fix the same root cause (commented `use strict`/`use warnings` lines falsely suppressing violations) and originate from the same Codex task.

**Why #7241 is strictly better:**

1. **`use strictures` false positive** — PR #7238's `has_use_statement` uses `code.contains(&needle)` where `needle = "use strict"`. The string `"use strictures;"` contains `"use strict"`, so `use strictures;` would incorrectly suppress a `RequireUseStrict` violation. PR #7241's token-based `has_use_statement_line` does an exact module name comparison (`module.trim_end_matches(';') == feature`) which correctly rejects `strictures != strict`.

2. **Better separation of concerns** — #7241 uses a named helper `has_use_statement_line` making the logic unit-testable in isolation.

3. **Token precision** — #7241 verifies the first token is exactly `"use"` and the second token exactly matches the feature name, handling leading whitespace and inline comments correctly without substring false positives.

A regression test for the `use strictures` edge case has been added to #7241 (commit `0870fc4`). Closing this PR in favor of #7241.

PR #7239, opened 2026-04-29
fix(perl-lsp-rs-core): support fully-qualified perlcritic aliases (#0000)

EffortlessSteven:
### Motivation
- Diagnostics emitted by `perlcritic` can include fully-qualified policy names like `Perl::Critic::Policy::InputOutput::RequireThreeArgOpen`, which did not match existing quick-fix routing that expected the short policy names.

### Description
- Normalize incoming diagnostic codes by stripping the `Perl::Critic::Policy::` prefix before match routing in `CodeActionsProvider::get_code_actions` so both short and fully-qualified names map to the same handlers.
- Fix a pattern-match variable name for `parse-error-*` guards to remain consistent with surrounding matches.
- Add a regression test `test_fully_qualified_perlcritic_policy_aliases_produce_quick_fixes` that verifies fully-qualified policy codes produce the same quick fixes for bareword filehandles and three-argument `open()` conversion.

### Testing
- Ran `cargo check --all-targets -p perl-lsp-rs-core` and it passed.
- Ran `cargo test -p perl-lsp-rs-core test_perlcritic_policy_aliases_produce_quick_fixes -- --exact` and it passed.
- Ran `cargo test -p perl-lsp-rs-core test_fully_qualified_perlcritic_policy_aliases_produce_quick_fixes -- --exact` and it passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c84f08c8333a3d1ba07ffd487e8)

PR #7241, opened 2026-04-29
fix(perl-lsp-rs-core): harden built-in use statement detection (#0000)

EffortlessSteven:
### Motivation
- The built-in Perl::Critic replacement treated commented lines like `# use strict;` / `# use warnings;` as real `use` declarations, which incorrectly suppressed required diagnostics.

### Description
- Replace the raw substring check in `missing_use_statement_violation` with `has_use_statement` that scans lines and ignores comments via `has_use_statement_line` to token-match `use <module>`.
- Add a regression test `comments_do_not_satisfy_use_statement_requirements` to ensure comment-only `use` lines do not satisfy the policies.
- Keep existing quick-fix logic and analyzer API unchanged while hardening detection to be comment-aware.

### Testing
- Ran `cargo test -p perl-lsp-rs-core comments_do_not_satisfy_use_statement_requirements -- --exact` and the test passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c84f08c8333a3d1ba07ffd487e8)

### DIFF
--- a/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
+++ b/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
@@ -583,4 +583,21 @@ eval $code;
         assert!(violations.iter().any(|v| v.policy == "TestingAndDebugging::RequireUseWarnings"));
         Ok(())
     }
+
+    #[test]
+    fn use_strictures_does_not_satisfy_use_strict_requirement() -> TestResult {
+        // "use strictures;" must NOT suppress RequireUseStrict — it is a different module.
+        // A substring check ("use strict" in "use strictures") would false-negative here.
+        let source = "use strictures;\nuse warnings;\nmy $x = 1;\n";
+        let mut parser = Parser::new(source);
+        let ast = parser.parse()?;
+        let analyzer = BuiltInAnalyzer::new();
+        let violations = analyzer.analyze(&ast, source);
+
+        assert!(
+            violations.iter().any(|v| v.policy == "TestingAndDebugging::RequireUseStrict"),
+            "use strictures should not satisfy RequireUseStrict — they are different modules"
+        );
+        Ok(())
+    }
 }
--- a/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
+++ b/crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs
@@ -191,7 +191,7 @@ fn missing_use_statement_violation(
     feature: &str,
     explanation: &str,
 ) -> Vec<Violation> {
-    if content.contains(&format!("use {feature}")) {
+    if has_use_statement(content, feature) {
         return Vec::new();
     }
 
@@ -203,6 +203,25 @@ fn missing_use_statement_violation(
         range: insertion_range(),
         file: String::new(),
     }]
+}
+
+fn has_use_statement(content: &str, feature: &str) -> bool {
+    content.lines().any(|line| has_use_statement_line(line, feature))
+}
+
+fn has_use_statement_line(line: &str, feature: &str) -> bool {
+    let code_portion = line.split('#').next().unwrap_or_default();
+    let mut tokens = code_portion.split_whitespace();
+    let Some(first) = tokens.next() else {
+        return false;
+    };
+    if first != "use" {
+        return false;
+    }
+    let Some(module) = tokens.next() else {
+        return false;
+    };
+    module.trim_end_matches(';') == feature
 }
 
 fn find_bareword_open_filehandles(content: &str) -> Vec<Range> {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
**Deep review: approved with improvements**

**What was verified:**
- Root cause correctly identified: `content.contains("use strict")` checked raw content including comment lines
- Fix approach (token-based `has_use_statement_line`) is correct and strictly better than the alternative substring approach in #7238
- Comment stripping via `split('#').next()` is consistent with existing `is_stringy_eval_line` pattern in the same file
- No regressions to existing callers — `missing_use_statement_violation` is a private function with two call sites (RequireUseStrict, RequireUseWarnings)

**What was fixed forward:**
- Added `use_strictures_does_not_satisfy_use_strict_requirement` test (commit `0870fc4`): the token-based implementation correctly rejects `use strictures;` as not satisfying `RequireUseStrict`, but a substring-based implementation would silently false-negative. The test documents and guards this distinction.
- `cargo fmt` confirmed: no formatting changes needed to existing code

**Edge cases checked:**
- `use strict;` — matches (correct)
- `use strictures;` — does NOT match (correct, guarded by new test)
- `# use strict;` — does not match due to comment strip (the PR's main fix)
- `use strict 'refs';` — matches on module name token (correct)
- `use\tstrict;` — `split_whitespace` handles tab correctly

**Logic:** CORRECT
**Tests:** IMPROVED (use strictures edge case added)
**Regression risk:** LOW
**Security:** CLEAN
**Performance:** CLEAN
**Research verification:** SKIPPED (no external claims)

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-7241",
  "current_version": {
    "sha": "0870fc4c80e58d812030f28a18922367029f9886",
    "updated_at": "2026-04-29T03:20:55Z"
  },
  "label_bindings": [
    {
      "label": "deep-reviewed",
      "bound_at_version": "0870fc4c80e58d812030f28a18922367029f9886",
      "bound_at_timestamp": "2026-04-29T03:21:23Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Deep Reviewer Wrapup

**Summary:** Deep reviewer completed 3-PR correctness review. Closed 1 inferior duplicate (#7238), fixed 2 logic/variable bugs, added 1 edge case test, applied `deep-reviewed` to 2 PRs.

### Ending State

- PR #7241 (`codex/improve-perl-critic-rust-tooling-d7n44k`): **approved**, `deep-reviewed` label set
  - Fixed forward: added `use_strictures_does_not_satisfy_use_strict_requirement` test (commit `0870fc4`) — proves the token-based approach correctly rejects `use strictures;` where a substring check would false-negative
  - 992 tests passing in `perl-lsp-rs-core`

- PR #7239 (`codex/improve-perl-critic-rust-tooling-7n5rvw`): **approved with fix**, `deep-reviewed` label set
  - Fixed: `parse-error-*` arm passed outer `code: &String` instead of bound variable `c` (the normalized `policy_code`). Behaviorally equivalent in practice since `parse-error-*` codes never carry the `Perl::Critic::Policy::` prefix, but semantically wrong and inconsistent with every other arm. Fixed to pass `c` (commit `ae99af0`).
  - 991 tests passing in `perl-lsp-rs-core`

- PR #7238 (`codex/improve-perl-critic-rust-tooling`): **closed** as inferior duplicate of #7241
  - Root issue: `has_use_statement` used `code.contains(&needle)` where `needle = "use strict"`, which falsely accepts `use strictures;` (a different module). PR #7241's token-based `has_use_statement_line` fixes this correctly.

### Retrospective

**Harder than expected:** Identifying that the three PRs were duplicates from the same Codex task took careful cross-checking — the branch names and PR titles were superficially different enough that a quick scan wouldn't have caught it. Worth checking the Codex task URL in the PR body when multiple PRs touch the same file in the same crate.

**Surprise:** PR #7239's `parse-error-*` bug was subtle — `code` (the `&String`) and `c` (the `policy_code &str`) are deref-compatible so the compiler accepted both, and the test passed because `parse-error-*` codes never have the `Perl::Critic::Policy::` prefix. The bug only matters if a future caller sends a contrived fully-qualified parse-error code. Caught by variable-consistency reasoning, not by test failure.

**What would have helped:** A pre-check that compares all open PRs from the same Codex task ID before starting the review would have surfaced the duplicate immediately. The Codex task URL in the PR body is the signal.

- Logical next: `green-ci` on PRs #7241 and #7239 (CI freshness gate before `diff-auditor`)
- Note: both PR titles contain `(#0000)` — not a real issue number. The `validate-title` CI check may flag this. Route to `pr-responder` if CI fails on title validation.
- `has_use_statement` / `has_use_statement_line` both live in `crates/perl-lsp-rs-core/src/tooling/perl_critic/built_in.rs`. The same `split('#').next()` comment-stripping pattern is used in `is_stringy_eval_line` — consistent.
- Gotcha: `missing_use_statement_violation` is a private function; its two callers are `RequireUseStrict::analyze` and `RequireUseWarnings::analyze`. Any change to the detection logic affects both.
- Confidence for next: **high**

PR #7251, opened 2026-04-29
fix(perl-uri): normalize malformed localhost Windows URIs (#0000)

EffortlessSteven:
### Motivation
- Some editors/clients emit malformed Windows `file://localhost/<drive>...` URIs which were not normalized by `uri_key`, causing inconsistent lookup keys across clients.

### Description
- Update `normalize_legacy_windows_uri` in `crates/perl-uri/src/classify.rs` to strip a malformed `localhost/` (case-insensitive) prefix before Windows-path normalization so inputs like `file://localhost/C:\...` become canonical `file:///c:/...` keys.
- Add regression tests `normalizes_legacy_localhost_windows_uri_variants` covering backslash and forward-slash variants and uppercase `LOCALHOST` to verify the new behavior.

### Testing
- Ran `cargo test -p perl-uri`, and all unit tests and doctests in the crate passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14c93984483338ebcaab2989b2ad3)

### DIFF
--- a/crates/perl-uri/src/classify.rs
+++ b/crates/perl-uri/src/classify.rs
@@ -90,10 +90,8 @@ fn normalize_legacy_windows_uri(uri: &str) -> Option<String> {
 
     // Accept malformed localhost authorities commonly emitted by some clients,
     // e.g. `file://localhost/C:\dir\file.pl` and `file://localhost/C:/dir/file.pl`.
-    let path = path
-        .strip_prefix("localhost/")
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Fixed: fmt drift at crates/perl-uri/src/classify.rs:90 corrected via `cargo xtask fmt`. The issue was improper line wrapping in the `path` assignment. Commit e20a6c229 pushed with tests passing (73 passed in perl-uri). Fmt check now passes.

PR #7265, opened 2026-04-29
fix(tree-sitter-perl-rs): remove expect from cursor path resolution (#0000)

EffortlessSteven:
### Motivation
- Prevent possible panics from `expect` in tree cursor path resolution and comply with the repository quality rule forbidding `unwrap`/`expect` in production code.

### Description
- Rewrote `resolve_path` to avoid `expect`: it now uses `if let Some(child)` to advance, emits a `debug_assert!` when the invariant is violated, and returns `root` as a safe fallback instead of panicking.

### Testing
- Ran `cargo test -p tree-sitter-perl-rs`, which passed (all unit, integration, and doctests succeeded). 
- Ran `cargo check --all-targets -p tree-sitter-perl-rs && cargo clippy -p tree-sitter-perl-rs`, which completed successfully with only unrelated warnings.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f14ca0f3d483338a2840e3fb12052a)

### DIFF
--- a/crates/tree-sitter-perl-rs/src/lib.rs
+++ b/crates/tree-sitter-perl-rs/src/lib.rs
@@ -661,14 +661,18 @@ fn collect_visible_use_imports(
     node.for_each_child(|child| collect_visible_use_imports(child, source, offset, out));
 }
 
-// Invariant: TreeCursor path is constructed by the traversal that just yielded this
-// index, so child_at is guaranteed valid.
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Superseded by PR #7266 which was merged. Both PRs addressed the same issue (removing expect from cursor path resolution) and the merged version is now in master.

PR #7168, opened 2026-04-28
test(perl-lsp-ux-tests): cover diagnostics edit lifecycle in harness (#5333)

EffortlessSteven:
## Summary

Salvage-cherry-pick of PR #5333 content onto current master with the green-tdd-flagged bug fixed.

**Bug fixed**: `wait_for_no_diagnostics` was permanently flaky — it only returned `true` if the server sent an explicit empty `publishDiagnostics` notification. If the server clears diagnostics silently (common behavior), the original test would always timeout and return `false`, causing `assert!(cleared, ...)` to fail.

**Fix**: The scenario test uses a two-phase approach:
1. Drain the pre-edit event queue via `collect_notifications()` to baseline the post-edit state.
2. After `change_file_full`, poll for EITHER an explicit empty notification (server clears explicitly) OR accept silence (no new non-empty notifications within timeout) as "cleared" — matching both server behaviors.

**Why cherry-pick**: The original PR #5333 branch predates the 2026-04-24 master fresh-root rebuild and cannot be rebased. Per the salvage-classifier doctrine and the ops comment on that PR, this is the correct recovery path.

## Changes

- `crates/perl-lsp-ux-tests/src/diagnostics.rs` — new `DiagnosticsTracker` helper with `latest_for_uri` / `wait_for_uri_matching`, with unit tests for both "most recent" semantics and "no match" cases.
- `crates/perl-lsp-ux-tests/src/lib.rs` — added `pub mod diagnostics`, `DiagnosticsTracker` re-export, and `wait_for_no_diagnostics` with a doc note about silent-clear behavior.
- `crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs` — robust BDD scenario that handles both explicit-clear and silence-clear server behaviors.
- `crates/perl-lsp-ux-tests/fixtures/editor_ux_fixture_matrix.json` — added `diagnostics_lifecycle_editing` entry.

## Test plan

- [ ] `cargo test -p perl-lsp-ux-tests --lib diagnostics` — 2 unit tests pass
- [ ] `cargo test -p perl-lsp-ux-tests --lib` — all 33 lib tests pass
- [ ] `cargo clippy -p perl-lsp-ux-tests --all-targets` — clean (only pre-existing warning in unrelated test)
- [ ] `cargo xtask fmt` — clean
- [ ] Integration test `ux_scenario_19_diagnostics_lifecycle` compiles; skips gracefully when binary not available

PR #7294, opened 2026-04-29
fix(perl-lsp-ux-tests): fix scenario_19 pre-fix event leak causing CI panic (#7293)

### LINKED ISSUE #7293: fix: scenario_19_diagnostics_clear_after_fix flaky on CI due to pre-fix event leak
## Problem

`scenario_19_diagnostics_clear_after_fix` at `crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs:111` panics on CI intermittently.

## Root cause

The test used `peek_notifications()` in the post-fix verification loop, which reads **all** accumulated events from the queue without removing them. After the two-pass drain (settle + second drain), a race window exists between the final drain call and the `change_file_full` send: the stdout reader thread can deliver pre-fix diagnostic events into the queue during that gap. These events are then visible to `peek_notifications()` and appear as post-fix events.

When the server never sends an explicit empty `publishDiagnostics` for the fixed content (which is a valid LSP behavior — silence = cleared), the leaked pre-fix events make `has_new_errors = true`, failing the assertion at line 111.

## History

- Test added in #7168
- First stabilization attempt (#7277): added 300ms settle window + second drain — reduced flakiness but did not eliminate the race
- This is the second stabilization attempt

## Fix (this issue)

1. Added a third drain immediately before `change_file_full` to close the settle→send race window
2. Changed from `peek_notifications()` (reads all accumulated events) to a periodic `collect_notifications()` drain pattern that accumulates ONLY events arriving after the final pre-fix drain into a local `post_fix_events` list
3. The post-fix verification loop and the silence-check both use `post_fix_events`, which is guaranteed to contain only post-fix events

## Files

- `crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs`

EffortlessSteven:
## Summary

- Closes #7293
- `scenario_19_diagnostics_clear_after_fix` was panicking on CI intermittently due to pre-fix diagnostic events leaking into the post-fix verification window
- Root cause: `peek_notifications()` reads all accumulated events without draining; events that raced into the queue between the settle drain and the `change_file_full` send appeared as post-fix events
- Fix: switch to periodic `collect_notifications()` drain pattern with a local `post_fix_events` accumulator; add a third drain immediately before `change_file_full` to close the settle→send race window

## Root cause detail

After `#7277`'s two-pass flush (drain → 300ms settle → drain), there's still a tiny race window between the final drain and the `change_file_full` notification send. The stdout reader thread can deliver pre-fix diagnostic events during this gap. `peek_notifications()` reads the entire queue so these events appear as post-fix, and when the server uses the silent-clear path (no explicit empty `publishDiagnostics`), `has_new_errors` becomes true — failing the assertion.

## Fix

1. Added a third `collect_notifications()` immediately before `change_file_full` to close the gap
2. Changed the post-fix loop from `peek_notifications()` (reads all) to periodic `collect_notifications()` drain appended to a local `post_fix_events` list — guaranteed to contain only post-fix events
3. The silence-check at the end also uses `post_fix_events` instead of `peek_notifications()`

## History

- Test added in #7168
- First stabilization (#7277): 300ms settle + second drain — improved but still flaky
- This PR: structural fix removing the data source for false positives

## Test plan

- [x] `cargo check -p perl-lsp-ux-tests` passes
- [x] `cargo fmt -p perl-lsp-ux-tests -- --check` passes
- [x] `cargo clippy -p perl-lsp-ux-tests` — no issues
- [ ] CI: `scenario_19_diagnostics_clear_after_fix` no longer panics on master

### DIFF
--- a/crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs
+++ b/crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs
@@ -54,14 +54,20 @@ fn scenario_19_diagnostics_clear_after_fix() {
 
     // Drain the event queue so post-fix checks only see new notifications.
     //
-    // Two-pass flush: the stdout reader thread may still be delivering in-flight
-    // events from the broken-content analysis (async diagnostics pipeline).  A
-    // single drain races with those deliveries — the queue can appear empty, then
-    // the late events arrive and get misclassified as "post-fix".  Waiting a
-    // brief settle window (≥ one server round-trip) and draining again eliminates
-    // that window reliably.
+    // Three-pass flush: the stdout reader thread may still be delivering in-flight
+    // events from the broken-content analysis (async diagnostics pipeline).
+    // - First drain: remove any already-buffered events.
+    // - Settle window: wait for in-flight events from the server to arrive.
+    // - Second drain: remove events that arrived during the settle window.
+    // - Final drain immediately before didChange: close the race window between
+    //   the settle and the send.  Events that slip in during this last tiny gap
+    //   are cleared before we record the post-fix baseline.
     harness.collect_notifications();
     std::thread::sleep(Duration::from_millis(300));
+    harness.collect_notifications();
+
+    // Final drain immediately before the fix — eliminates events that snuck in
+    // between the settle drain and the didChange notification.
     harness.collect_notifications();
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Fix merged into PR #7295

Issue #7297, opened 2026-04-29
test(perl-lsp-ux-tests): scenario_19_diagnostics_clear_after_fix flaky on CI (#7178)


EffortlessSteven:
## Problem

The test `scenario_19_diagnostics_clear_after_fix` in `crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs` is structurally flaky in CI despite three stabilization attempts:

- **#7168**: Test added
- **#7277**: First stabilization (300ms settle + second drain) → improved but still flaky  
- **#7294**: Second stabilization (structural fix with `post_fix_events` accumulator + third drain) → still fails in CI despite passing locally

## Root cause hypothesis

The panic line has moved across stabilization attempts (102 → 111 → 126), suggesting a **structural race in the test harness itself**, not the test logic:
- The settle-drain-send race window persists despite successive closure attempts
- `peek_notifications()` reads all queued events without draining; late-arriving pre-fix diagnostics appear as post-fix events
- Event ordering is not guaranteed between the settle window and the file change send

## Escalation

Three independent fix attempts have not resolved the underlying race. The test harness architecture may need redesign.

## Tracking

Parent issue: #7178 (UX Regression Tests baseline drift)

## Test affected

`scenario_19_diagnostics_clear_after_fix` in `crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs`

EffortlessSteven:
Duplicate of #7296 (filed earlier with the same diagnostic history and stabilization next-steps). Closing in favor of the existing issue.

PR #7298, opened 2026-04-29
test(perl-lsp-ux-tests): mark scenario_19_diagnostics_clear_after_fix as flaky (#7178)

EffortlessSteven:
## Summary

After three independent stabilization attempts (#7168, #7277, #7294), `scenario_19_diagnostics_clear_after_fix` remains flaky in CI despite passing locally. The panic line has moved across attempts (102 → 111 → 126), indicating a **structural race in the test harness**, not the test itself.

## Changes

- Marked test with `#[ignore = "flaky in CI; tracked in #7297"]`
- Created follow-up issue #7297 with diagnostic history and root-cause hypothesis

## Root cause

The settle-drain-send race window in the test harness persists despite successive closure attempts:
- `peek_notifications()` reads all queued events without draining
- Late-arriving pre-fix diagnostics from the async pipeline appear as post-fix events
- Event ordering is not guaranteed between settle window and file change send

The harness architecture may need redesign to handle async event delivery guarantees.

## Test plan

- [x] `cargo check -p perl-lsp-ux-tests` passes
- [x] `cargo clippy -p perl-lsp-ux-tests --tests` passes
- [x] `cargo xtask fmt --check` passes

## Tracking

- Parent: #7178 (UX Regression Tests baseline drift)
- Details: #7297 (test harness race analysis)

### DIFF
--- a/crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs
+++ b/crates/perl-lsp-ux-tests/tests/ux_scenario_19_diagnostics_lifecycle.rs
@@ -30,6 +30,7 @@ const FIXED_SOURCE: &str = "use strict;\nmy $x = 1;\n";
 ///   1. Broken content → diagnostics appear.
 ///   2. Fixed content → diagnostics clear (either explicitly or by silence).
 #[test]
+#[ignore = "flaky in CI; tracked in #7297"]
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Duplicate of #7295 (filed earlier with the same ignore-attribute fix). Closing in favor of the existing PR.

PR #7314, opened 2026-04-29
feat(perl-semantic-facts): add neutral semantic fact foundation (#0000)

EffortlessSteven:
### Motivation

- Provide a neutral, serializable vocabulary for semantic facts so parser, analyzer, and workspace layers can share a stable API. 
- Keep this crate limited to typed stable IDs, facts, provenance, and confidence metadata without parsing, LSP behavior, or workspace storage responsibilities. 
- Make the initial slice reviewable and testable before any consumer migration.

### Description

- Added a new crate `crates/perl-semantic-facts` with `Cargo.toml`, `README.md`, and `src/lib.rs` implementing typed IDs (`FileId`, `ScopeId`, `EntityId`, `AnchorId`, `OccurrenceId`, `EdgeId`, `DiagnosticId`) and fact records (`AnchorFact`, `EntityFact`, `OccurrenceFact`, `EdgeFact`, `DiagnosticFact`).
- Implemented core enums `EntityKind`, `OccurrenceKind`, `EdgeKind`, `Provenance`, and `Confidence` and derived `Serialize`/`Deserialize` for serialization support via `serde`.
- Added crate-level docs and a README that state the crate is a neutral facts vocabulary and does not parse Perl or own workspace storage.
- Wired the crate into the workspace `Cargo.toml` and updated lockfile entries so the new crate builds independently.

### Testing

- Ran `cargo test -p perl-semantic-facts` and the unit tests passed (3 tests: JSON roundtrip, deterministic `Debug`, pretty-JSON stability).
- Ran `cargo check -p perl-semantic-facts` which finished successfully.
- Ran `cargo check --workspace --all-targets` which completed; it reported pre-existing workspace warnings unrelated to this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f1ae494e2083338bae2e2d542805f3)

PR #7341, opened 2026-04-29
feat(perl-symbol): adapt SymbolDecl into semantic facts (#0000)

EffortlessSteven:
### Motivation

- Provide a parser-near adapter that translates `SymbolDecl` into the neutral `perl-semantic-facts` vocabulary so downstream layers can consume canonical `AnchorFact`, `EntityFact`, and `EdgeFact::Defines` records.
- Ensure outputs are deterministic and test-friendly by producing stable IDs and by explicitly reporting unsupported declaration kinds rather than silently dropping them.

### Description

- Added a focused adapter module `crates/perl-symbol/src/surface/facts.rs` with `symbol_decls_to_semantic_facts` that emits `AnchorFact`, `EntityFact`, and `EdgeFact::Defines` records from `SymbolDecl` inputs and returns `SymbolDeclSemanticFacts` and `UnsupportedDeclFact` entries for unsupported kinds.
- Uses a deterministic FNV-like `stable_id` function to synthesize `AnchorId`, `EntityId`, and `EdgeId` values so facts are stable across runs and suitable for golden tests.
- Maps supported `SymbolKind` variants (package, class, subroutine, method, variable, constant, label, format) to `EntityKind` and treats `Role`, `Import`, and `Export` as unsupported with explicit reporting.
- Exposed the adapter and types through `surface/mod.rs` and crate `api.rs`, added `perl-semantic-facts` to `crates/perl-symbol/Cargo.toml`, and added unit and snapshot-style tests in `crates/perl-symbol/src/surface/facts.rs` and `crates/perl-symbol/tests/surface_decl_facts.rs`.

### Testing

- Ran `cargo test -p perl-symbol` and all tests passed, including the new deterministic adapter tests (`adapter_is_deterministic_for_mixed_decls`) and the snapshot-style projection test (`adapter_snapshot_for_symbol_decl_projection`).
- Ran `cargo test -p perl-semantic-facts` and the crate tests passed.
- Ran `cargo check --workspace --all-targets` which completed successfully (workspace warnings unrelated to this change remain).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f1cec052888333b2d70415c254ed24)


EffortlessSteven:

**HEAD after fix:** 903990f88

### What I verified

**FNV-1a stable_id**
Constants are correct (offset\_basis `0xcbf29ce484222325`, prime `0x00000100000001b3`). Separator byte `0xff` is a sound choice (not a valid UTF-8 continuation byte). The algorithm is deterministic within a process and stable across runs for fixed inputs. Cross-arch variance (usize is 4 vs 8 bytes) is inherited from the `(usize, usize)` span types already in the codebase — not introduced here.

**EntityKind mapping (8 supported, 3 unsupported)**
All 8 supported kinds (`Package`, `Class`, `Subroutine`, `Method`, `Variable(_)`, `Constant`, `Label`, `Format`) map to existing `EntityKind` variants in `perl-semantic-facts`. The three unsupported (`Role`, `Import`, `Export`) have no matching `EntityKind` in the vocabulary crate — deferring them is correct. `Role` in particular maps to `EntityKind::Role` in the facts crate, so that is a future upgrade path, not a current gap.

**Unsupported reporting**
The `UnsupportedDeclFact` approach emits explicit diagnostic records instead of silently dropping. Both the entity-kind path and the missing-container path emit into `unsupported`. No silent data loss.

**Test assertion quality**
The three existing tests are not vacuous — they would each fail if the function returned empty facts or wrong counts. The snapshot test checks structure (JSON key presence, entity count, edge count) but not specific edge identities. That is acceptable for a projection test; the new regression test covers identity correctness.

### Bug found and fixed

**Container resolution: `ends_with` was a raw byte-suffix match, not a segment-boundary match.**

Original logic (line 85):
```rust
if prefix.ends_with(container) {
```

`"FooBar".ends_with("Bar")` is `true`. When a package `FooBar` exists in `entity_by_name` and a sibling `Bar` is the intended container, the wrong entity received the `Defines` edge.

Fix applied (commit 903990f88):
```rust
let is_segment_match =
    prefix == container.as_str() || prefix.ends_with(&format!("::{container}"));
```

A regression test `container_resolution_uses_segment_boundary_not_byte_suffix` was added that explicitly asserts `edge.from_entity_id == bar_entity.id` (not FooBar's id). This test would have failed on the original code.

`cargo xtask fmt` applied; all 306 tests in `perl-symbol` pass.

### Routing

No external Perl/LSP/crate claims requiring `research-verifier` dispatch. No attribution claims. Regression risk is LOW — change is fully contained in `crates/perl-symbol/src/surface/facts.rs`, a new module with no existing callers in the workspace.

Next: **green-ci** (verify CI on 903990f88) → **diff-auditor** → ops merge.

EffortlessSteven:
```json
{
  "schema_version": "1.0",
  "artifact_id": "pr-7341",
  "current_version": {
    "sha": "e607b8d73ee8239ce16b2cb2d485b2c1eb7bbf6e",
    "updated_at": "2026-04-29T18:22:26Z"
  },
  "label_bindings": [
    {
      "label": "deep-reviewed",
      "bound_at_version": "e607b8d73ee8239ce16b2cb2d485b2c1eb7bbf6e",
      "bound_at_timestamp": "2026-04-29T18:22:49Z",
      "valid": true
    }
  ]
}
```

EffortlessSteven:
## Deep Reviewer Wrapup

**Summary:** reviewer-deep completed correctness gate on PR #7341 (perl-symbol SymbolDecl→semantic-facts adapter). Fixed 1 logic bug, added 3 edge-case tests, all 308 tests green.

### Ending state

- PR improved: https://github.com/EffortlessMetrics/perl-lsp/pull/7341
- Changes made: 2 commits pushed to PR branch
  - `903990f88` — logic fix: segment-boundary check in `from_candidate` container resolution
  - `83221cdde` — 2 additional edge-case tests (missing container, deeply nested multi-level qualifier)
- Files touched:
  - `crates/perl-symbol/src/surface/facts.rs` (fix + 3 inline tests)
  - `crates/perl-symbol/tests/surface_decl_facts.rs` (fmt only, reformatted by `cargo xtask fmt`)
- Tests passing: 308 passed, 0 failed (`cargo test -p perl-symbol`)
- `deep-reviewed` label: set
- `needs-deep-review` label: removed
- Receipt comment: https://github.com/EffortlessMetrics/perl-lsp/pull/7341#issuecomment-4346397252
- Status: approved

### What was found

**Bug (fixed):** `prefix.ends_with(container)` on line 85 of `facts.rs` was a raw byte-suffix match. `"FooBar".ends_with("Bar")` evaluates to `true`, so when a package `FooBar` was present in `entity_by_name` and `Bar` was the intended container, the wrong entity received the Defines edge. Fixed to:
```rust
let is_segment_match =
    prefix == container.as_str() || prefix.ends_with(&format!("::{container}"));
```
The maintainer had specifically flagged this exact interaction ("interaction between decl.qualified_name and the container field during prefix matching warrants one correctness pass") — the flag was correct.

**Tests added:**
1. `container_resolution_uses_segment_boundary_not_byte_suffix` — directly asserts `edge.from_entity_id == bar_entity.id`, would have failed on original code
2. `missing_container_produces_unsupported_entry` — exercises the missing-container fallback path with a specific reason-string assertion
3. `deeply_nested_qualified_name_resolves_correct_container` — verifies `ends_with("::C")` for A→A::B→A::B::C→A::B::C::D chain

**No security or performance issues found.** No external Perl/LSP/crate claims requiring research-verifier dispatch.

### Retrospective

The maintainer's pre-flagged concern ("interaction between decl.qualified_name and the container field") was perfectly targeted — that is exactly where the bug lived. When a maintainer flags a specific interaction in the PR context, treat it as a near-certain find, not a maybe. The fix was a one-liner but the test to prove it needed both FooBar and Bar in entity_by_name simultaneously to catch the false-positive match path.

The most useful thing to know going in would have been the full `entity_by_name` BTreeMap population order, since the bug only manifests when a same-suffix sibling package also has a declaration. The test corpus (snapshot test + determinism test) used only unambiguous simple names like "Foo", so the bug would never have triggered in the existing tests.

For future adapter-style PRs: always trace the two-pass structure (entity-build pass, then edge-build pass) separately, and specifically ask "does the lookup key ever collide with a different entity's name via substring/suffix?" for any `.ends_with()` or `.contains()` call on qualified names.

### Breadcrumbs for green-ci

- Logical next: green-ci agent verifies CI on HEAD `e607b8d73ee8239ce16b2cb2d485b2c1eb7bbf6e` → diff-auditor → ops merge
- Gotchas: PR branch has a master-merge bump (`25bb9ce28`) from `0.13.0-rc1` prep already rebased in. The `CLAUDE.md` version header changed to `0.13.0-rc1` — this is correct, not drift.
- Traps: `facts.rs` was reformatted by `cargo xtask fmt` (long function signatures split across lines) — don't mistake the fmt diff for scope drift
- Confidence for next: high

EffortlessSteven:
**Admin-merging** — all CI checks SUCCESS, no failures. Full review chain: review-reviewed + maintainer-pr-reviewed + deep-reviewed (with byte-suffix→segment-boundary bug fix + 3 new tests). Wave 2 SymbolDecl adapter — companion to merged #7314 facts foundation.

PR #7357, opened 2026-04-29
feat(perl-workspace): add write-through file fact shards (#0000)

EffortlessSteven:
### Motivation

- Provide Wave 2 with an in-memory write-through place to store canonical semantic facts alongside the existing symbol/reference index without changing public query behavior.
- Keep shard storage observational and lifecycle-managed so upstream APIs like `find_definition` / `find_references` remain unchanged.

### Description

- Added `perl-semantic-facts` as a dependency for `perl-workspace` to model anchors, entities, occurrences, and edges via typed fact records.
- Introduced `FileFactShard` and a `fact_shards: Arc<RwLock<HashMap<String, FileFactShard>>>` field on `WorkspaceIndex` to hold per-file fact shards (URI, file id, content hash, per-category hashes, and fact vectors).
- Wired shard lifecycle into indexing operations by constructing a shard in `index_file` (via `build_fact_shard`), replacing on reindex, removing in `remove_file`, and clearing in `clear`, and added inspection APIs `fact_shard_count` and `file_fact_shard`.
- Added focused tests `tests/fact_shard_lifecycle_tests.rs` that verify index/populate, reindex/replace, remove, and clear lifecycle behavior.

### Testing

- Ran `cargo test -p perl-workspace facts` and the new fact-shard lifecycle tests passed. 
- Ran `cargo test -p perl-workspace` and observed failures in existing `surface_workspace_parity` tests (`surface_workspace_divergence_details_are_actionable` and `surface_workspace_parity_bank`) that are unrelated to the shard lifecycle changes. 
- Ran `cargo check --workspace --all-targets` which progressed with warnings and no compilation errors at the time of reporting.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f1ced1649c8333934e165355457fa8)


EffortlessSteven:
## Deep Review — correctness findings and fixes pushed

Reviewed HEAD `e86c0d0a9` (pre-fixes) → `d412f3beb` (post-fixes).

### Bug 1 — `span_start_byte`/`span_end_byte` stored line numbers (semantic misuse) [FIXED]

In `build_fact_shard`, `AnchorFact.span_start_byte` was set to `symbol.range.start.line` and `span_end_byte` to `symbol.range.end.line`. The field names are `span_*_byte` and are documented as byte offsets in `perl-semantic-facts`. `WorkspaceSymbol.range` uses LSP line/column coordinates (0-based line number + column), not byte offsets. Storing line numbers in byte offset fields would corrupt any downstream consumer using these values as byte offsets.

Fix: zero-initialize both fields with a comment explaining why, pending a future byte-offset source.

### Bug 2 — `anchors_hash`/`entities_hash` stored lengths, not hashes [FIXED]

`anchors_hash: Some(anchors.len() as u64)` stores the count as a "hash". Two files with different anchor content but the same symbol count would produce the same `anchors_hash`, defeating the change-detection purpose of the field.

Fix: compute proper `DefaultHasher`-based hashes over the anchor IDs and entity canonical names — those are the stable distinguishing attributes at this layer.

### Bug 3 — Missing idempotency test for same-content reindex [FIXED]

The maintainer flagged this. The existing `reindex_replaces_stale_fact_shard` only tests different content. The early-exit path (content hash match → `return Ok(())` without rebuilding the shard) was untested. Added `reindex_same_content_preserves_shard` test to `fact_shard_lifecycle_tests.rs`.

### Lock ordering analysis — no ABBA deadlock risk

Traced lock acquisition order across `index_file`, `remove_file`, and `clear`:
- All mutation paths that touch multiple locks acquire `files.write()` first, serializing concurrent callers. No path holds an inner lock (`symbols`, `global_refs`, `fact_shards`) while calling into `index_file` or `remove_file`.
- `parking_lot::RwLock` is used (not `std::sync`), which is non-re-entrant by design but provides no cross-lock cycle detection. The consistent acquisition order (`files → {global_refs, symbols, fact_shards}`) prevents ABBA deadlock.
- `fact_shards.write()` is acquired last inside the `index_file` critical section while `symbols.write()` is still held — this is consistent with `remove_file`'s ordering (both hold `files` as the outer guard). No cycle.

### URI key normalization consistency — existing pattern, not a new bug

`file_fact_shard` uses `DocumentStore::uri_key(&Self::normalize_uri(uri))` while `index_file` uses `DocumentStore::uri_key(&uri_str)` where `uri_str = uri.to_string()` (already a `Url`). This matches the existing pattern used by `file_symbols` and other read paths vs write paths. For well-formed `Url`-derived strings the keys are equivalent. Not introduced by this PR.

### Commits pushed to branch

- `c79a7106e` — test: idempotency test for same-content reindex
- `d412f3beb` — fix: `span_start_byte` misuse + real hashes in `build_fact_shard`

PR #7389, opened 2026-04-29
fix(perl-lsp-rs): avoid PATH-hijacked which probe (#0000)

EffortlessSteven:
### Motivation
- The Unix branch of `command_exists` spawned an external `which` via `Command::new("which")`, which is resolved through `PATH` and can be hijacked by an attacker-controlled `which` binary.
- Tool detection (e.g. for `perltidy`/`perlcritic`) runs during server startup and diagnostics, making this probe reachable without explicit user action and creating an unacceptable local command-execution risk.

### Description
- Replaced the external probe with an in-process lookup by changing `command_exists` to use `which::which(command).is_ok()` so no PATH-resolved helper binary is executed. 
- Promoted the `which` crate from a dev-dependency into normal `[dependencies]` in `crates/perl-lsp-rs/Cargo.toml` because production code now depends on it. 
- Added a Unix regression test `test_command_exists_does_not_execute_path_hijacked_which` which prepends a malicious `which` into `PATH` and asserts it is not executed by `command_exists`.

### Testing
- Ran `cargo test -p perl-lsp-rs --test execute_command_security_tests` and the test binary completed with all tests passing (15 passed, 0 failed). 
- Ran `cargo check --all-targets -p perl-lsp-rs` and it completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f17b674bac8333a431009573aa9f00)

### DIFF
--- a/crates/perl-lsp-rs/Cargo.toml
+++ b/crates/perl-lsp-rs/Cargo.toml
@@ -80,6 +80,7 @@ ropey = { version = "1.6.1" }
 tokio = { workspace = true }
 once_cell.workspace = true
 tracing.workspace = true
+which = "8.0.2"
 
 [build-dependencies]
 
@@ -126,7 +127,6 @@ tempfile.workspace = true
 proptest.workspace = true
 perl-tdd-support = { workspace = true }
 # Wave Final PR B: perl-content-length-framing absorbed into perl-lsp-rs-core::transport::framing
-which = "8.0.2"
 parking_lot.workspace = true
 criterion.workspace = true
 perl-corpus = { workspace = true }
--- a/crates/perl-lsp-rs/src/execute_command/provider.rs
+++ b/crates/perl-lsp-rs/src/execute_command/provider.rs
@@ -949,18 +949,7 @@ pub(crate) fn select_test_runner(
 
 /// Check whether a command exists in the current PATH.
 pub fn command_exists(command: &str) -> bool {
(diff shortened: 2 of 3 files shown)


EffortlessSteven:
## Standards Review Pass ✓

### Scope & Code Quality
- **Scope**: Tight and focused — single-crate security fix (3 files, 45 net changes)
- **Banned patterns**: None detected (no unwrap/expect/panic/todo)
- **Implementation**: 12-line function → 1-line in-process lookup (significant improvement)
  - Replaced spawned external binary with  in-process probe
  - Eliminates PATH hijacking vulnerability in tool detection
  - Dependency legitimately promoted from dev to production

### Test Quality
- Uses `Result<>()` return type per standards
- Marker-file pattern proves binary wasn't executed
- Proper cleanup with `unsafe SAFETY` comments
- Unix-specific gating via `#[cfg(unix)]`
- Clear assertion message

### Security Assessment
- **Threat model**: Local PATH hijacking during startup/diagnostics tool detection
- **Risk**: Unacceptable (reachable without explicit user action)
- **Resolution**: In-process lookup eliminates external binary execution
- **Regression coverage**: Present and well-designed

### Title
- Format: `fix(perl-lsp-rs): avoid PATH-hijacked which probe (#0000)` ✓

**Routing**: Gate clean → route to reviewer-deep for correctness check.

PR #7397, opened 2026-04-29
fix(xtask): enforce per-gate command timeouts (#0000)

EffortlessSteven:
### Motivation
- Gate-level command runs could be terminated by outer workflow timeouts and surface opaque exit codes (e.g., SIGTERM/143), making triage hard and losing which gate actually timed out.

### Description
- Wrap shell-backed gate commands with a per-gate timeout helper (`shell_command_with_timeout`) that uses GNU `timeout` on non-Windows and falls back to unwrapped execution on Windows.
- Detect hard timeouts by recognizing `timeout`'s exit code (`124`) and keep a duration-based fallback to mark a gate as `timeout` when appropriate.
- Enrich the gate `output_summary` when a timeout occurs to include the timeout duration, the exact command, and the gate log path for faster attribution and repro.
- Adjusted the main gate execution path to use the new helper while preserving existing internal-xtask gate dispatch behavior.

### Testing
- Ran `cargo check -p xtask`, which completed successfully with only pre-existing warnings.
- Verified the xtask binary builds locally after the change (`cargo check`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f21d5b01948333b33b5741c74a731f)

### DIFF
--- a/xtask/src/tasks/gates.rs
+++ b/xtask/src/tasks/gates.rs
@@ -1612,7 +1612,8 @@ fn run_single_gate(
     // Run through the platform shell because gate commands are policy strings.
     // Linux CI uses bash; Windows local runs use cmd.exe to keep `cargo`/`just`
     // lookup and simple command chaining available without requiring Git Bash.
-    let result = shell_command(command).stderr_to_stdout().stdout_capture().unchecked().run();
+    let (gate_cmd, timeout_enforced) = shell_command_with_timeout(command, timeout_secs);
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as ARCHITECTURAL-MISFIT — Windows codepath returns `timeout_enforced=false`, silently disabling per-gate timeouts on Windows. Contradicts the motivation 'enforce per-gate timeouts' uniformly. Superseded by #7398 which has cross-platform polling-based timeout enforcement.

PR #7422, opened 2026-04-30
fix(perl-lexer): preserve start checkpoints after edits (#0000)

EffortlessSteven:
### Motivation
- `CheckpointCache::apply_edit` could drop otherwise-valid checkpoints when an edit moved them to byte position `0`, which breaks incremental parsing anchors at start-of-file.
- Add focused edge-case coverage to drive a red->green fix for this condition.

### Description
- Stop removing checkpoints at position `0` by removing the over-eager retention that discarded `CheckpointContext::Normal` entries with `position == 0` in `CheckpointCache::apply_edit` (`crates/perl-lexer/src/checkpoint.rs`).
- Add a regression test `test_checkpoint_cache_apply_edit_preserves_start_checkpoint` in the `tests` module of `crates/perl-lexer/src/checkpoint.rs` that deletes bytes before a checkpoint and asserts `find_before(0)` still returns the moved checkpoint.
- Keep the existing position-resort logic (positions are updated from each checkpoint after `apply_edit`) so cached order and lookup invariants remain intact.

### Testing
- Ran the focused test suite with `cargo test -p perl-lexer checkpoint::tests::` and all checkpoint tests passed (`12 passed; 0 failed`).
- During development an added red test exposed the bug (failing assertion), and after the fix the test became green as above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f29d1c16b483339c288a7edb14166c)

### DIFF
--- a/crates/perl-module/src/resolution/use_lib.rs
+++ b/crates/perl-module/src/resolution/use_lib.rs
@@ -37,8 +37,8 @@ pub enum UseLibAction {
 pub fn extract_use_lib_paths(source: &str) -> Vec<UseLibPath> {
     let mut paths = Vec::new();
 
-    for line in source.lines() {
-        let trimmed = line.trim();
+    for statement in perl_statements(source) {
+        let trimmed = statement.trim();
         if let Some(rest) = strip_use_lib_prefix(trimmed) {
             extract_paths_from_args(rest, &mut paths);
         }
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Merge conflicts with #7420 (both touched checkpoint.rs apply_edit). Bouncing for rebase. Re-apply on top of master and the labels/diff-audit will need to be re-validated.

EffortlessSteven:
Closing — original branch unrebaseable. Patch does not apply cleanly. Tracking issue: https://github.com/EffortlessMetrics/perl-lsp/issues/7508

PR #7450, opened 2026-04-30
fix(xtask): stream update-status command output (#0000)

EffortlessSteven:
### Motivation
- `update-status --write` used buffered subprocess output which can appear silent for long-running steps and trigger CI inactivity kills.

### Description
- Switched `run_cmd` and `run_cmd_merged` to spawn child processes with piped stdout/stderr and stream output via a new `stream_child_output` helper in `xtask/src/tasks/update_status/mod.rs`.
- `stream_child_output` reads stdout/stderr lines on background threads, writes each line immediately to stderr, and accumulates combined output so callers keep the same return semantics.
- Added required imports and a small mpsc/threading-based implementation while keeping the public helper contracts unchanged.

### Testing
- Tried to run `cargo test -p xtask update_status::mod_tests`, but the test run could not complete due to unrelated compile errors in `crates/perl-symbol/src/surface/mod.rs`, so unit tests were blocked and did not finish.
- Committed the change as `fix(xtask): stream update-status command output (#0000)` and verified the file diff (`xtask/src/tasks/update_status/mod.rs`) contains the streaming implementation.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2ea1efcec8333882e540b6a17f51a)

### DIFF
--- a/xtask/src/tasks/update_status/mod.rs
+++ b/xtask/src/tasks/update_status/mod.rs
@@ -15,8 +15,11 @@
 //! Also keeps docs/project/ROADMAP.md compliance table in sync when lsp subsystem runs.
 
 use std::fs;
+use std::io::{BufRead, BufReader};
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
+use std::sync::mpsc;
+use std::thread;
 use std::time::Duration;
 
 use color_eyre::eyre::{Context, Result, eyre};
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as part of cluster triage — keeper is #7453. Same intent (stream xtask update-status output to prevent CI inactivity kills), different variant from same Codex burst.

PR #7452, opened 2026-04-30
fix(xtask): stream update-status subprocess output (#0000)

EffortlessSteven:
### Motivation

- `xtask update-status` used buffered subprocess capture which can produce long stretches of silence on stdout/stderr and trigger GitHub Actions inactivity timeouts. 

### Description

- Replaced buffered `Command::output()` usage in `xtask/src/tasks/update_status/mod.rs` with spawning a piped child and streaming stdout/stderr live while still capturing their contents. 
- Added a `stream_and_capture` helper and switched `run_cmd` to `Command::spawn()` with `Stdio::piped()` and per-stream reader threads, preserving the combined output return value for downstream parsing. 

### Testing

- Ran `cargo test -p xtask update_status -- --nocapture`, which exercised the changed `update-status` runner but could not complete due to a pre-existing compile error in `crates/perl-symbol/src/surface/mod.rs` (duplicate imports), so the test run did not finish successfully. 
- Ran `cargo check -p xtask --all-targets`, which also failed to finish for the same unrelated `perl-symbol` compile error.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2ea1efcec8333882e540b6a17f51a)

### DIFF
--- a/xtask/src/tasks/update_status/mod.rs
+++ b/xtask/src/tasks/update_status/mod.rs
@@ -15,8 +15,9 @@
 //! Also keeps docs/project/ROADMAP.md compliance table in sync when lsp subsystem runs.
 
 use std::fs;
+use std::io::Read;
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::time::Duration;
 
 use color_eyre::eyre::{Context, Result, eyre};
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as part of cluster triage — keeper is #7453. Same intent (stream xtask update-status output to prevent CI inactivity kills), different variant from same Codex burst.

Issue #7404, opened 2026-04-29
ci: update-status --write hits Actions inactivity timeout (cancelled at 5min)


EffortlessSteven:
## Symptom

Post-Merge Status Regeneration workflow run [25116946219](https://github.com/EffortlessMetrics/perl-lsp/actions/runs/25116946219) cancelled at 15:16:51Z while running `xtask update-status --write`. Started at 15:11:04Z (build complete) so cancellation hit at ~5min 47s of stdout silence.

## Root cause

Same pattern as #7334 (gates SIGTERM): `update-status --write` likely buffers stdout via `stdout_capture()` instead of streaming via tee. GitHub Actions inactivity timer kills jobs after ~5min of no output.

## Fix

Replace `stdout_capture()` in update-status implementation with tee-style streaming (same pattern as #7334 for `xtask gates`).

## Impact

Post-Merge Status Regeneration runs after every merge to master; cancellations leave docs/project/status/*.md stale without anyone noticing until a status-drift PR fires.

PR #7527, opened 2026-04-30
fix(xtask): stream update-status progress output (#7404)

EffortlessSteven:
### Motivation

- Long-running `xtask update-status --write` steps produced buffered/silent output and allowed GitHub Actions to cancel post-merge status regeneration due to inactivity. 
- The intent is to stream child process output and emit periodic progress so Actions sees activity during expensive subsystem work.

### Description

- Replaced buffered `Command::output()` usage with a new `run_cmd_streaming` that pipes child `stdout`/`stderr`, forwards child lines to logs in real time, and keeps a combined buffer for existing consumers. 
- `run_cmd_merged` now calls the streaming runner so shell `2>&1` cases maintain ordering while streaming output. 
- Emit heartbeat logs every 30 seconds when a child is quiet and add explicit `subsystem start` / `subsystem complete` messages for `lsp`, `tests`, `parser`, `quality`, `dap`, and `workspace` so progress is visible to CI. 
- Wrap each subsystem invocation with a small closure that augments errors with per-subsystem repro text (for example `cargo run -p xtask -- update-status --write --only parser`).

### Testing

- Attempted `cargo run -p xtask -- update-status --write`, which started building but the run was blocked by a pre-existing unrelated compile error (`E0252` duplicate imports) in `crates/perl-symbol/src/surface/mod.rs`. 
- Attempted `cargo test -p xtask update_status -- --nocapture`, which also failed due to the same unrelated workspace compile error. 
- No automated tests were changed; the update-status coordination tests remain unchanged and local test attempts were prevented by the unrelated compile failure.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2f86bf14883338cbc928690d3c79f)

PR #7529, opened 2026-04-30
fix(xtask): stream update-status quality progress (#7404)

EffortlessSteven:
### Motivation
- Long-running status collectors (notably `cargo test -- --list` in the quality subsystem) can produce long stretches of stdout silence and trigger GitHub Actions inactivity cancellation.  
- The goal is to provide a tee-style/streaming progress heartbeat so Actions receives live output while heavy status generation runs, without changing status document content or weakening `--check` behavior.

### Description
- Add a new `run_cmd_merged_streaming` helper in `xtask/src/tasks/update_status/mod.rs` that spawns a shell, merges `stderr` into `stdout`, streams child output live to `stderr` while capturing the full output, and returns `Result<String>` with detailed failure context including an exact repro command.  
- Use the streaming helper from the quality subsystem by switching `collect_per_crate_test_counts` to call `run_cmd_merged_streaming` and change its signature to return `Result<BTreeMap<...>>` so errors are surfaced with context.  
- Emit explicit subsystem-start progress messages (`eprintln!` like `[update-status] subsystem start: <name>`) before each status subsystem (`lsp`, `tests`, `parser`, `quality`, `dap`, `workspace`) so CI receives a heartbeat before long stages.  
- Preserve existing status generation and file-write semantics; only execution and reporting of child commands were changed (and quality now propagates failures instead of silently returning empty output).

### Testing
- Ran `cargo run -p xtask -- update-status --write --only parser`, which started and printed build progress, but the run could not complete because the workspace currently fails to compile due to an unrelated pre-existing `crates/perl-symbol` duplicate-import error (E0252), so end-to-end `update-status` verification is blocked.  
- Ran `cargo check -p xtask` and attempted `cargo test -p xtask update_status -- --nocapture`, both of which are blocked by the same upstream compile error, so unit tests exercising the new streaming behavior could not be completed in this workspace state.  
- The change is focused and limited to streaming/reporting behavior for the quality collector and subsystem progress logging; it does not alter status document formatting logic and will be effective once the workspace build is green.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2f86bf14883338cbc928690d3c79f)

### DIFF
--- a/xtask/src/tasks/update_status/mod.rs
+++ b/xtask/src/tasks/update_status/mod.rs
@@ -15,8 +15,9 @@
 //! Also keeps docs/project/ROADMAP.md compliance table in sync when lsp subsystem runs.
 
 use std::fs;
+use std::io::{BufRead, BufReader};
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::time::Duration;
 
 use color_eyre::eyre::{Context, Result, eyre};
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as part of cluster triage — keeper is #7527. Same intent (xtask update-status streaming), different variant from same Codex burst.

PR #7537, opened 2026-04-30
fix(xtask): emit update-status subsystem heartbeat progress (#0000)

EffortlessSteven:
### Motivation

- `cargo xtask update-status --write` can run for many minutes without producing output, which causes GitHub Actions to consider the job inactive and cancel it.
- Emit lightweight progress so Actions receives periodic stdout/stderr and long-running subsystem work does not hit the ~5 minute inactivity cancellation window.

### Description

- Add a small `SubsystemProgress` guard that logs subsystem start and spawns a background thread that prints a "still running" heartbeat every 30 seconds while the subsystem executes.
- Wire the progress guard into each status subsystem run (`lsp`, `tests`, `parser`, `quality`, `dap`, `workspace`) so `--write` emits activity before and during each subsystem execution. 
- Preserve existing generator behavior and outputs; this change only adds diagnostic/heartbeat output and does not change generated file contents or check semantics.

### Testing

- Ran `cargo test -p xtask update_status -- --nocapture`, which started a build but the run failed due to pre-existing workspace compile errors in `crates/perl-symbol` (duplicate-import E0252), so xtask tests could not complete. (failed)
- Ran `cargo check -p xtask`, which also failed due to the same unrelated compile errors in `crates/perl-symbol`, preventing full verification of the xtask binary in this snapshot. (failed)
- Ran `rustfmt`/format step and committed the focused change; formatting of unrelated modules reported existing issues but the change itself was formatted before commit.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2f8c6c3248333802412808784fd85)

### DIFF
--- a/xtask/src/tasks/update_status/mod.rs
+++ b/xtask/src/tasks/update_status/mod.rs
@@ -17,6 +17,9 @@
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::thread;
 use std::time::Duration;
 
 use color_eyre::eyre::{Context, Result, eyre};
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as part of cluster triage — keeper is #7527. Same intent (xtask update-status streaming), different variant from same Codex burst.

PR #7561, opened 2026-04-30
fix(ci): upload UX regression receipts automatically (#0000)

EffortlessSteven:
### Motivation

- UX regression failures required manual log inspection because CI did not always produce or upload structured receipts usable for triage.
- The repository already has `cargo xtask ux-regression-receipt`; the CI should invoke it and publish its output so regressions are repairable from artifacts while preserving the gate as a hard failure.

### Description

- Updated `.github/workflows/ci.yml` to capture `just ux-tests` output to `target/receipts/ux-regression.log`, run `cargo xtask ux-regression-receipt` to write `target/receipts/ux-regression.json`, upload both artifacts, emit a short GitHub step summary pointing to the receipt/log, and still fail the job when UX tests fail via an explicit fail step keyed on `steps.ux_tests.outcome`.
- Updated `.github/workflows/ux-regression-gate.yml` to adopt the same `target/receipts/` convention, `tee` the UX test output to `target/receipts/ux-regression.log`, run the receipt generator on `if: always()`, upload the receipt/log artifacts, update the summary parser to read the new paths, and preserve gate semantics by failing when `steps.ux_tests.outcome == 'failure'`.
- Preserved existing UX gate behavior and reporting while adding structured receipts and artifact uploads; no UX test logic or scenarios were changed.

### Testing

- Ran `cargo test -p xtask ux_regression_receipt` to exercise the classifier unit tests but the command failed due to pre-existing unrelated compile errors in `crates/perl-symbol/src/surface/mod.rs` (duplicate re-exports), preventing the unit tests from completing.
- Workflow YAML changes were validated by static inspection and local edits to ensure the `tee`/receipt/upload/summary/fail sequencing meets the requirements (capture to `target/receipts`, run receipt on success/failure, upload receipt/log, and preserve gate failure signal).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2f87b146c8333b27d7fe3fa968cb9)

PR #7562, opened 2026-04-30
fix(ci): upload UX regression receipts from gate workflow (#0000)

EffortlessSteven:
### Motivation

- Make UX regression failures repairable from CI artifacts by emitting a structured receipt in the standard `target/receipts/` location.
- Ensure the UX gate still fails the job on real test failures while allowing receipt generation and artifact upload to run in both pass and fail cases.

### Description

- Capture UX test output with `just ux-tests` piped through `tee` into `target/receipts/ux-regression.log` and create the `target/receipts` directory if missing.
- Run the UX test step with `continue-on-error` semantics and publish the test exit code as `ux_exit_code` via `GITHUB_OUTPUT` so later steps can enforce the gate.
- Always invoke the classifier: `cargo xtask ux-regression-receipt --input target/receipts/ux-regression.log --receipt target/receipts/ux-regression.json --sha "${{ github.sha }}"` so a JSON receipt is produced for both successes and failures.
- Upload both `target/receipts/ux-regression.json` and `target/receipts/ux-regression.log` as artifacts and enforce the UX gate by exiting non-zero when the captured `ux_exit_code` is non-zero.
- Update the GitHub step-summary parsing to read the new `target/receipts/*` paths so the run summary can point to classified failure fields when available.

### Testing

- Attempted to run the classifier unit tests with `cargo test -p xtask ux_regression_receipt -- --nocapture`, but the run was blocked by a pre-existing unrelated compilation error in `crates/perl-symbol` (duplicate imports), so the xtask test could not complete. 
- Verified the workflow changes locally by inspecting the YAML diff and ensuring the new `tee`/receipt/instrumentation and artifact upload steps are present and reference `target/receipts/` as required.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f2f87b146c8333b27d7fe3fa968cb9)

### DIFF
--- a/.github/workflows/ux-regression-gate.yml
+++ b/.github/workflows/ux-regression-gate.yml
@@ -115,17 +115,22 @@ jobs:
       - name: Run UX regression tests
         id: ux_tests
         if: steps.harness_check.outputs.harness_available == 'true'
+        continue-on-error: true
         run: |
+          set +e
           echo "Running UX regression tests..."
-          just ux-tests 2>&1 | tee /tmp/ux-test-output.txt
-          exit ${PIPESTATUS[0]}
+          mkdir -p target/receipts
+          just ux-tests 2>&1 | tee target/receipts/ux-regression.log
+          ux_status=${PIPESTATUS[0]}
+          echo "ux_exit_code=${ux_status}" >> "$GITHUB_OUTPUT"
+          exit 0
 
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing as part of cluster triage — keeper is #7561. Same intent, different variant from same Codex burst.

PR #7599, opened 2026-04-30
fix(xtask): add update-status heartbeat output (#0000)

EffortlessSteven:
### Motivation

- `cargo xtask update-status --write` can run long-running child commands while emitting no stdout, which allows GitHub Actions to cancel the job due to inactivity and leaves status docs stale. 
- The intent is to stream liveliness/progress so Actions sees output even when subprocesses are quiet and to make subsystem repro lines explicit for reproducible failures.

### Description

- Add a heartbeat thread inside `run_cmd` that prints a progress line every 30 seconds while a child process is running, implemented with an `Arc<AtomicBool>` flag to stop the heartbeat when the child exits. 
- Preserve existing subsystem start/completion/failure messages and the repro-line behavior emitted by `run_subsystem`, so failures still name the subsystem and provide a reproduction command. 
- Add a focused unit test that verifies subsystem repro commands are explicit write-only commands (e.g. `cargo xtask update-status --write --only parser`) to keep reproduction guidance accurate.

### Testing

- Ran `./scripts/storage-doctor` which completed successfully in this environment. 
- Attempted the focused unit test via `./scripts/cargo-safe test -p xtask --profile agent --locked update_status::mod_tests::test_update_status_repro_commands_are_write_only` but it was blocked by the environment storage policy (`DENY: /root/.cache/devplane/perl-lsp`), so the test was not executed here. 
- No other automated test failures were observed in this session; CI should run full `xtask` tests and validate runtime behavior on Actions where the heartbeat prevents inactivity cancellations.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f36fb5d9a483338693837377bee002)

### DIFF
--- a/xtask/src/tasks/update_status/mod.rs
+++ b/xtask/src/tasks/update_status/mod.rs
@@ -129,9 +129,7 @@ fn run_cmd(root: &Path, args: &[&str], timeout: Duration) -> String {
         while heartbeat_flag.load(Ordering::Relaxed) {
             std::thread::sleep(Duration::from_secs(30));
             if heartbeat_flag.load(Ordering::Relaxed) {
-                eprintln!(
-                    "[update-status] still running (heartbeat): {command_name}"
-                );
+                eprintln!("[update-status] still running (heartbeat): {command_name}");
             }
         }
     });
--- a/xtask/src/tasks/update_status/mod.rs
+++ b/xtask/src/tasks/update_status/mod.rs
@@ -18,6 +18,8 @@ use std::fs;
 use std::io::{BufRead, BufReader, Read};
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Duration;
 
 use color_eyre::eyre::{Context, Result, eyre};
@@ -120,7 +122,23 @@ fn run_cmd(root: &Path, args: &[&str], timeout: Duration) -> String {
     // the parameter for API compatibility and future improvement.
     let _ = timeout;
 
+    let heartbeat_running = Arc::new(AtomicBool::new(true));
+    let heartbeat_flag = Arc::clone(&heartbeat_running);
+    let command_name = args.join(" ");
+    let heartbeat = std::thread::spawn(move || {
+        while heartbeat_flag.load(Ordering::Relaxed) {
+            std::thread::sleep(Duration::from_secs(30));
+            if heartbeat_flag.load(Ordering::Relaxed) {
+                eprintln!(
+                    "[update-status] still running (heartbeat): {command_name}"
+                );
+            }
+        }
+    });
+
     let status = child.wait();
+    heartbeat_running.store(false, Ordering::Relaxed);
+    let _ = heartbeat.join();
     let mut combined = String::new();
     if let Some(handle) = out_handle {
         combined.push_str(&handle.join().unwrap_or_default());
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
## Accuracy Verification ✓

**File paths, line numbers, function names:** All verified correct.

**Heartbeat implementation:**
- ✓ Imports added correctly (Arc, AtomicBool, Ordering at lines 20-21)
- ✓ Spawned in run_cmd before child.wait() (lines 122-136 on PR branch)
- ✓ Uses Arc<AtomicBool::new(true)> for thread-safe synchronization
- ✓ Sleeps 30s, then prints to stderr if flag still true (prevents race condition)
- ✓ After child.wait(), sets flag to false and joins thread (correct lifecycle)
- ✓ Memory ordering Relaxed is appropriate for simple flag
- ✓ Uses eprintln! which writes to stderr (won't interfere with stdout parsing by subsystems)

**Test verification:**
- ✓ test_update_status_repro_commands_are_write_only() added correctly
- ✓ All 6 subsystem commands verified (lsp, tests, parser, quality, dap, workspace)
- ✓ Subsystem names match StatusSubsystem enum exactly (lines 45-52)
- ✓ Test correctly placed in mod_tests within cfg(test) block
- ✓ Clear error messages for assertion

**Integration points:**
- ✓ Heartbeat applies to all run_cmd calls (all subsystems benefit)
- ✓ Existing subsystem start/completion/failure messages preserved
- ✓ Repro-line behavior unchanged
- ✓ stderr and stdout handled separately (no parsing collision)

**Issue alignment:**
- ✓ PR addresses #7404 (Actions inactivity timeout)
- ✓ 30-second heartbeat prevents 5+ minute silence periods
- ✓ Output format [update-status] still running is GitHub Actions compatible

**No mechanical issues found.** Code compiles on PR branch, test assertions are sound, no flush()-handling concerns (project standard is implicit flushing via eprintln in interactive/pipe contexts).

Labeling .

EffortlessSteven:
## Deep review: SIGN OFF (with fix-forward fmt)

**Reviewer**: reviewer-deep (Opus, no-worktree)
**HEAD before review**: f33c7bbdf36f94369294c0c9e02ae756a14b846c
**HEAD after fix-forward**: 3655ef4d30cf699c4281c21d9ac7aa1c04b6bdd3
**Issue**: #7404 (update-status streaming for GitHub Actions inactivity timeout)

### Correctness analysis

| Concern | Verdict | Notes |
|---|---|---|
| Atomic ordering (Relaxed) | OK | Flag-only synchronization; no other shared data depends on it. `heartbeat.join()` provides the happens-before edge for any post-thread inspection. |
| Thread join after child exits | OK | No orphaned threads. `heartbeat_running.store(false)` precedes `heartbeat.join()`. |
| Race window (flag set→false vs heartbeat read) | Bounded | Worst case: heartbeat prints one trailing "still running" line after child has exited (TOCTOU between inner `if` and `eprintln!`). Cosmetic, not a correctness or panic risk. |
| Join-tail latency | Up to 30s | If child completes mid-sleep, `heartbeat.join()` blocks for up to ~30s waiting for the next loop iteration. Not catastrophic for an `update-status` run that already takes minutes per subsystem; not worth the complexity of a `Condvar`-based interruptible sleep. Documenting as known. |
| Heartbeat goes to stderr | Correct | Matches existing `[update-status]` eprintln prefix; stdout parsers (e.g., test runners) won't see it. |
| 30s vs GitHub's 5-min cancel | 10x margin — good | First heartbeat at +30s; child completing in <30s emits no heartbeat (correct). |
| eprintln! across threads | Safe | `eprintln!` takes the stderr lock; no interleaving risk with other writers in the parent. |
| Test (`test_update_status_repro_commands_are_write_only`) | Tautological but harmless | Hard-codes the 6 repro strings then asserts they begin with the expected prefix. It's a contract literal-test, not a coverage test of `run_subsystem` call sites. Not a regression — same shape as the existing `test_update_status_is_split_into_subsystem_modules` structural assertions. Considered filing as green-tdd hardening but not a deep-review block. |

### Fix-forward applied

`cargo fmt -p xtask --check` flagged the heartbeat `eprintln!` as multi-line where rustfmt collapses it to one line. Pushed fix as commit [3655ef4d](https://github.com/EffortlessMetrics/perl-lsp/commit/3655ef4d30cf699c4281c21d9ac7aa1c04b6bdd3) so the merge gate fmt check passes. Net: additions 37 → 35 (3-line eprintln → 1-line).

### Pre-existing finding (not blocking, not introduced by this PR)

`mod.rs` is now 515 lines after this PR (was 478 on master, threshold in `test_update_status_is_split_into_subsystem_modules` is `loc <= 400`). That structural-LOC test is **already failing on master** (confirmed locally via `cargo test -p xtask test_update_status_is_split_into_subsystem_modules`). It does not run in PR Smoke or CI Gate (xtask tests aren't included in `cargo test --workspace --lib`), so it doesn't block this PR. Worth a follow-up issue to either tighten the gate to actually run in CI or split mod.rs further — separate bit-rot from the heartbeat fix. Not gating sign-off on it.

### Verification

- `cargo check -p xtask` — clean
- `cargo fmt -p xtask --check` — clean (after fix-forward)
- `cargo clippy -p xtask --bin xtask` — only pre-existing collapsible-if warning at line 149 (unrelated)
- Tautology Detection — SKIPPED (nightly tier)

Applying `deep-reviewed`. Routes next to green-ci → diff-auditor → ops per pipeline.

PR #7711, opened 2026-05-01
test(perl-dap): harden regex mutation survivor coverage (#0000)

EffortlessSteven:
### Motivation
- The regex-mutation validator in `crates/perl-dap/src/eval/validator.rs` only inspected the first regex match, allowing an early allowed token (e.g. `$s` or `\s`) to mask a later dangerous `s///`/`tr///` operator.
- Add focused tests to ensure dangerous regex mutation operators are detected even when they appear after allowed identifiers or escape sequences.

### Description
- Update `check_regex_mutation` to iterate all matches using `re.find_iter(...)` and `continue` on allowed cases instead of returning after the first match in `crates/perl-dap/src/eval/validator.rs`.
- Replace early `return Ok(())` paths with `continue` so later matches are still examined.
- Add unit tests `test_regex_mutation_detected_after_allowed_identifier` and `test_regex_mutation_detected_after_allowed_escape_sequence` alongside the existing test suite in `crates/perl-dap/src/eval/validator.rs` to cover the regression.

### Testing
- Added unit tests in `crates/perl-dap/src/eval/validator.rs` that assert `s///` and `tr///` are detected when placed after allowed tokens.
- Attempted to run the validator tests with `./scripts/cargo-safe test -p perl-dap --profile agent --locked eval::validator` but the run was blocked by the environment storage policy (`DENY: /root/.cache/devplane/perl-lsp used=46% free=32G`), so local verification is pending and will be exercised in CI.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f4333691a4833396d4080b5c2c4fee)

### DIFF
--- a/crates/perl-dap/src/eval/validator.rs
+++ b/crates/perl-dap/src/eval/validator.rs
@@ -154,18 +154,18 @@ impl SafeEvaluator {
             return Ok(());
         };
 
-        if let Some(mat) = re.find(expression) {
+        for mat in re.find_iter(expression) {
             let op = mat.as_str();
             let start = mat.start();
 
             // Allow sigil-prefixed identifiers ($s, $tr, $y)
             if is_sigil_prefixed_identifier(expression, start) {
-                return Ok(());
+                continue;
             }
 
             // Allow escape sequences like \s, \y
             if is_escape_sequence(expression, start) {
-                return Ok(());
+                continue;
             }
 
             return Err(ValidationError::RegexMutation(op.trim().to_string()));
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Review refresh for current master.

Verification run locally:
- `cargo fmt --package perl-dap`
- `cargo check -p perl-dap --all-targets --locked` (passes; existing DAP test-support dead-code warnings remain)
- `cargo test -p perl-dap eval::validator --locked` (16 validator tests passed, including the two new regression tests)
- `cargo clippy -p perl-dap --lib --locked -- -D warnings`
- `git diff --check`

Additional strict all-target clippy check attempted:
- `cargo clippy -p perl-dap --all-targets --locked -- -D warnings` is blocked by pre-existing DAP test-support issues in `tests/common/mod.rs` plus an existing `collapsible_match` in `dap_reference_client_conformance_tests.rs`. I did not fold those unrelated test cleanups into this focused validator fix.

PR #7736, opened 2026-05-01
refactor(perl-lexer): split checkpoint impl into submodule (#0000)

EffortlessSteven:
Problem: `perl-lexer` still kept `Checkpointable` wiring in the large crate root after the test module split landed.

Fix: Move the `PerlLexer` checkpoint implementation into `checkpoint_impl.rs` and leave `lib.rs` as the module composition point.

Verification: `cargo fmt --package perl-lexer`, `cargo check -p perl-lexer --all-targets --locked`, `cargo test -p perl-lexer --locked`, `cargo clippy -p perl-lexer --all-targets --locked -- -D warnings`, and `git diff --check origin/master...HEAD` pass.

### DIFF
--- a/crates/perl-lexer/src/checkpoint_impl.rs
+++ b/crates/perl-lexer/src/checkpoint_impl.rs
@@ -1,0 +1,66 @@
+use crate::*;
+
+impl Checkpointable for PerlLexer<'_> {
+    fn checkpoint(&self) -> LexerCheckpoint {
+        use checkpoint::CheckpointContext;
+
+        // Determine the checkpoint context based on current state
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Review update: rebased after #7740 landed. The duplicate unit-test extraction dropped out, so this PR is now only the checkpoint implementation extraction.

Verification:

```bash
cargo fmt --package perl-lexer
cargo check -p perl-lexer --all-targets --locked
cargo test -p perl-lexer --locked
cargo clippy -p perl-lexer --all-targets --locked -- -D warnings
git diff --check origin/master...HEAD
```

PR #7737, opened 2026-05-01
test(perl-parser-core): bound heredoc depth fixture (#0000)

EffortlessSteven:
Problem: PR Smoke unit_core can hang on a heredoc depth-limit fixture that exercises a parenthesized initializer instead of the intended pending-heredoc guard.
Fix: Use the minimal over-limit heredoc declaration shape with bodies and assert the parser reaches the depth-limit diagnostic quickly.
Verification: cargo test -p perl-parser-core test_deeply_nested_heredocs_hit_limit --lib --locked passes; cargo test -p perl-parser -p perl-lexer -p perl-parser-core --lib --locked -- --test-threads=4 passes; cargo fmt -p perl-parser-core --check passes; git diff --check passes.


EffortlessSteven:
Closing as already satisfied on current master. Rebase onto origin/master dropped the PR's only commit as patch contents already upstream, leaving an empty diff.

PR #7738, opened 2026-05-01
refactor(perl-lexer): split core lexer into srp module boundary (#0000)

EffortlessSteven:
### Motivation

- The crate root combined crate wiring and the full `PerlLexer` implementation, creating a large multi-responsibility file that hindered single-responsibility organization and maintainability.
- Introduce a clear SRP module boundary so the crate facade only exposes public types and the heavy implementation lives in a focused `lexer` submodule.

### Description

- Move the full `PerlLexer` implementation from `crates/perl-lexer/src/lib.rs` into a new submodule at `crates/perl-lexer/src/lexer/mod.rs` and keep `lib.rs` as the crate facade with module declarations and public re-exports.
- Update internal imports in the moved implementation to use explicit `crate::` paths (e.g., `crate::token`, `crate::mode`, `crate::unicode`, `crate::checkpoint`, etc.) so the submodule remains self-contained.
- Move the lexer-specific test module `test_format_debug.rs` into `crates/perl-lexer/src/lexer/test_format_debug.rs` so tests live next to the implementation they exercise.
- Preserve public re-exports from `lib.rs` (including `PerlLexer`, token types and utilities) with no behavioral changes to the lexer API.

### Testing

- Ran `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo check -p perl-lexer --all-targets --locked`, which completed successfully for the `perl-lexer` crate.
- Attempted repository-level agent checks: `./scripts/cargo-safe check -p perl-lexer --all-targets --profile agent --locked` was blocked by local environment storage guard, and `just agent-check -p perl-lexer` was not available in this environment, so those agent-style checks could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f4338b68248333b14d2a113d59a1a9)


EffortlessSteven:
Queue hygiene: closing this broad lexer move variant. Keeping the incremental path, starting with #7780 for state/constructors, so lexer refactors land in reviewable slices instead of a full implementation move. Branch preserved for reference.

PR #7773, opened 2026-05-01
refactor(perl-dap): extract server lifecycle from lib facade (#0000)

EffortlessSteven:
### Motivation

- Reduce the crate-root surface area by moving server lifecycle, mode, and config out of `lib.rs` so the facade only answers “what is the public API”.
- Improve Single-Responsibility boundaries so server configuration and lifecycle are owned by a `server` domain rather than the crate root.

### Description

- Added `crates/perl-dap/src/server` with `mod.rs`, `config.rs`, `mode.rs`, and `lifecycle.rs` and moved `DapConfig`, `DapMode`, and `DapServer` into those files.
- Updated the crate root to publish the `server` module and kept compatibility by re-exporting `DapConfig`, `DapMode`, and `DapServer` from `lib.rs` as before.
- Kept all existing behavior unchanged; `DapServer::new`, `run`, and `run_socket` were moved into `server::lifecycle` with the same signatures and semantics.

### Testing

- Ran `cargo check -p perl-dap --all-targets --locked` which completed successfully.
- A full test suite was not executed in this change; no behavioral tests were modified and the public API was preserved via re-exports.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f450d0bb1083338bd1542330813a26)

### DIFF
--- a/crates/perl-dap/src/lib.rs
+++ b/crates/perl-dap/src/lib.rs
@@ -370,6 +370,8 @@ pub mod configuration;
 pub mod debug_adapter;
 /// DAP feature catalog and capability gating helpers.
 pub mod feature_catalog;
+/// DAP server configuration and lifecycle.
+pub mod server;
 
 // Wave H collapsed modules (in DAG order — command_args before platform, platform before shell, value before variables)
 /// Explicit public API re-exports from all collapsed satellite modules.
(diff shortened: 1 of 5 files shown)


EffortlessSteven:
Queue hygiene: closing this duplicate Codex variant. Keeping #7776 as the DAP lifecycle survivor because it covers the same extraction and reports the API stability plus external consumer integration tests. Branch preserved for reference.

PR #7835, opened 2026-05-02
feat(completion): use semantic method candidates (#0000)

EffortlessSteven:
Problem: Method completion still relied on the legacy workspace member list even when semantic method_candidates had confident answers.
Fix: Route workspace method completion through semantic method_candidates when high/medium-confidence semantic candidates cover the legacy method set, with conservative fallback to the existing path.
Verification: cargo check -p perl-lsp-rs-core --all-targets --profile agent --locked; cargo test -p perl-lsp-rs-core --profile agent --locked completion; cargo test -p perl-workspace --profile agent --locked method_candidates; cargo clippy -p perl-lsp-rs-core --profile agent --locked -- -D warnings -A missing_docs; cargo xtask semantic-scorecard --check; cargo xtask semantic-shadow-compare --check; cargo xtask fmt --check.

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion/tests.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion/tests.rs
@@ -1176,6 +1176,103 @@ $this->"#;
 }
 
 #[test]
+fn test_method_completion_semantic_inheritance_detail() -> Result<(), Box<dyn std::error::Error>> {
+    let index = Arc::new(WorkspaceIndex::new());
+    index.index_file(
+        Url::parse("file:///workspace/Parent.pm")?,
+        r#"package Parent;
+sub inherited_method { }
+1;
+"#
+        .to_string(),
+    )?;
+    index.index_file(
+        Url::parse("file:///workspace/Child.pm")?,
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Addressed the ancestor-shadowing review point in `9607c20e`: semantic completion candidates now have to match the legacy-resolved defining package for each method name before the provider cuts over. Added `test_method_completion_prefers_nearest_ancestor` to cover `Child -> Parent -> Base` where both ancestors define the same method.

Local verification included:
- `cargo test -p perl-lsp-rs-core --profile agent --locked completion`
- `cargo test -p perl-lsp-rs-core --profile agent --locked test_method_completion_prefers_nearest_ancestor -- --nocapture`
- `cargo check -p perl-lsp-rs-core --all-targets --profile agent --locked`
- `cargo clippy -p perl-lsp-rs-core --profile agent --locked -- -D warnings -A missing_docs`
- `cargo test -p perl-workspace --profile agent --locked method_candidates`
- `cargo xtask semantic-scorecard --check`
- `cargo xtask semantic-shadow-compare --check`
- `cargo xtask fmt --check`
- `git diff --check`

PR #7857, opened 2026-05-02
fix(parser): accept `=>` in `tie` builtin within expression context (#0000)

EffortlessSteven:
## Summary

Fix a parser hang/OOM on the real-world `(tie %hash => $module, ...)` idiom from `Memoize/Expire.pm`.

The `tie` handler in `crates/perl-parser-core/src/engine/parser/expressions/primary.rs` (used in expression context, e.g. inside parentheses) only accepted a comma between the variable and the package, while the statement-level handler in `statements.rs` already accepted both `,` and `=>` (Perl treats `=>` as a synonym for `,`). When the wrong-token path in `expect()` consumed the `=>` and bubbled an error up, statement-level recovery sent the parser into an unbounded loop and OOM-killed the process on input as small as a single line.

The fix mirrors the statement-context logic in the primary-context handler so both sites accept `,` or `=>` between the variable and the package, and as the trailing-args separator.

## Reproducer

```perl
my $rc = (tie %cache => $module, @opts);  # from Memoize/Expire.pm:32
```

Before: parser OOMs / hangs (SIGKILL).
After: parses cleanly to a `(tie ...)` AST node.

## Test plan

- [x] New regression file `tests/fix_tie_fat_arrow_in_paren.rs` covering:
  - `(tie %h => $m, $n);`
  - `my $rc = (tie %cache => $module, @opts);` (Memoize/Expire.pm shape)
  - `(tie %h => $m);`
  - `(tie my %h => 'Pkg', $arg);`
  - `(tie %h, $m, @opts);` (comma form — no behaviour change)
  - mixed `,` and `=>` in trailing args
- [x] All 6 new tests pass; pre-existing `tie`-related tests still pass
- [x] `cargo clippy -p perl-parser-core --lib` clean
- [x] `cargo fmt --check` clean
- [x] Manually verified `Memoize/Expire.pm` now parses without OOM via `perl-parse`

The remaining failures in the parser-core test suite (e.g. `extra_closing_paren_does_not_cascade`, `test_double_comma_in_function_args`, several `subst_e_*` tests, `test_double_invocant_separator_is_rejected`) all reproduce on master without this change and are out of scope for this PR — they appear to be a separate class of unbounded-recovery bugs on malformed input.

https://claude.ai/code/session_01UbaRhRKJNUghC7kkjCUMVT

---
_Generated by [Claude Code](https://claude.ai/code/session_01UbaRhRKJNUghC7kkjCUMVT)_

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/expressions/primary.rs
+++ b/crates/perl-parser-core/src/engine/parser/expressions/primary.rs
@@ -687,10 +687,28 @@ impl<'a> Parser<'a> {
                             } else {
                                 Box::new(self.parse_assignment()?)
                             };
-                            self.expect(TokenKind::Comma)?;
+                            // Accept comma or fat arrow between variable and
+                            // package — Perl treats `=>` as a synonym for `,`.
+                            match self.peek_kind() {
+                                Some(TokenKind::Comma | TokenKind::FatArrow) => {
+                                    self.consume_token()?;
+                                }
+                                _ => {
+                                    return Err(ParseError::unexpected(
+                                        TokenKind::Comma.display_name(),
+                                        self.peek_kind()
+                                            .map(|k| k.display_name())
+                                            .unwrap_or("end of input"),
+                                        self.current_position(),
+                                    ));
+                                }
+                            }
                             let package = Box::new(self.parse_assignment()?);
                             let mut args = vec![];
-                            while self.peek_kind() == Some(TokenKind::Comma) {
+                            while matches!(
+                                self.peek_kind(),
+                                Some(TokenKind::Comma | TokenKind::FatArrow)
+                            ) {
                                 self.consume_token()?;
                                 args.push(self.parse_assignment()?);
                             }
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Review pass: this is a focused parser-core correctness fix and the diff matches the reported Memoize/Expire.pm shape. Local verification:\n\n- cargo test -p perl-parser-core --locked --test fix_tie_fat_arrow_in_paren -- --nocapture -> 6 passed\n- cargo check -p perl-parser-core --all-targets --locked -> passed\n- cargo clippy -p perl-parser-core --lib --locked -- -D warnings -A missing_docs -> passed\n- cargo xtask fmt --check -> passed\n- git diff --check -> passed\n\nI also ran cargo clippy -p perl-parser-core --all-targets --locked -- -D warnings -A missing_docs; it is currently blocked by unrelated pre-existing test-only violations in 
ecovery_phase2_tests.rs and ix_unexpected_rbrace_proto_plus_2835.rs (xpect() / panic!()), not by this PR.\n\nRecommendation: this can leave draft when the shared release-history PR Smoke blocker is resolved; no extra scope should be added here.

EffortlessSteven:
Hosted PR Smoke classification: branch-local parser gates passed in the job (`clippy_scoped`, `unit_scoped`, `check_tests_scoped`, and `perl_token_leaf_contract` all passed). The only blocking failures were the shared `release_history` and `release_history_check` gates, matching the current master release-history drift. No parser-code change needed for that.

PR #7858, opened 2026-05-02
feat(semantic-analyzer): infer value-shape-lite receivers (#0000)

EffortlessSteven:
Problem: Receiver-shape inference only covered constructors, bless, and bare self references, leaving common signature/self-unpack and DBI handle flows invisible.
Fix: Track lexical value shapes in the analyzer and infer conservative object shapes for signature/self receivers, DBI->connect, and DBI prepare handles.
Verification: `cargo test -p perl-semantic-analyzer --lib --profile agent --locked`, `cargo check -p perl-semantic-analyzer --all-targets --profile agent --locked`, and `cargo clippy -p perl-semantic-analyzer --profile agent --locked -- -D warnings -A missing_docs` pass; `cargo xtask gates --tier pr-fast --receipt` is blocked locally by missing just shell/cygpath.

### DIFF
--- a/crates/perl-semantic-analyzer/src/analysis/value_shape_inferrer.rs
+++ b/crates/perl-semantic-analyzer/src/analysis/value_shape_inferrer.rs
@@ -5,9 +5,13 @@
 //!
 //! | Perl pattern                             | Inferred shape                                |
 //! |------------------------------------------|-----------------------------------------------|
-//! | `Foo->new(...)`                          | `Object { "Foo", High }`                      |
+//! | `Foo->new(...)`                          | `Object { "Foo", Medium }`                    |
 //! | `bless $ref, 'Pkg'`                      | `Object { "Pkg", Low }`                       |
 //! | `$self` in method body                   | `Object { <enclosing package>, Medium }`       |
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Review improvement pushed: tightened the DBI prepare inference to avoid false precision. $dbh->prepare(...) now infers DBI::st only when $dbh is already known as a DBI::db handle from DBI->connect(...); variable naming alone no longer creates a DBI statement shape. Added a regression test for $dbh->prepare(...) without a known connect shape.\n\nVerification after the change:\n\n- cargo test -p perl-semantic-analyzer --lib --profile agent --locked -j 1 -- --nocapture -> 232 passed\n- cargo test -p perl-semantic-analyzer --lib --profile agent --locked -j 1 analysis::value_shape_inferrer -- --nocapture -> 12 passed\n- cargo check -p perl-semantic-analyzer --all-targets --profile agent --locked -j 1 -> passed\n- cargo clippy -p perl-semantic-analyzer --profile agent --locked -j 1 -- -D warnings -A missing_docs -> passed\n- cargo xtask fmt --check -> passed\n- git diff --check -> passed

PR #7930, opened 2026-05-04
feat(completion): add bounded low-confidence fallback for unknown method receivers (#7929)

EffortlessSteven:
Closes #7929
Plan source: #7929
Plan-review decisions: [#7929 comment](https://github.com/EffortlessMetrics/perl-lsp/issues/7929#issuecomment-4372876089)

## Outcome: A — bounded low-confidence Unknown-receiver fallback

When receiver inference returns `ReceiverEvidence::Unknown` (no exact evidence), the completion provider now offers methods from a **bounded** source — `use`d packages plus the current file's package and its `@ISA` chain — with low-confidence detail labels and a sort tier below all exact-receiver completions.

This is the first PR in this lane that **changes candidate inclusion**. Every hard constraint from the issue is honored.

## What the user sees

Before — `$obj->` with no receiver evidence produced no method completions:

```
$obj->          (no completions)
```

After — bounded fallback offers methods from imported / current packages:

```perl
use Foo;          # exports `bark`
package MyService;
sub helper { ... }
sub run {
    my ($obj) = @_;
    $obj->         # ← Unknown receiver
}
```

```
bark            workspace method — receiver: unknown, low confidence (from Foo)
helper          workspace method — receiver: unknown, low confidence (from MyService)
```

Sort tier 6 — below every exact-receiver tier.

## Hard-constraints honored

- Exact receiver unchanged. `StaticPackage` / `SelfOrThis` / `ConstructorAssignment` / `LiteralBless` / `TypeEngine` produce identical candidates, sort, and detail (with #7920 / #7926 contracts intact).
- Dynamic stays fail-closed. `ReceiverEvidence::Dynamic` positively distinguishes the four fail-closed bless forms (`bless {}, $class`, `bless {}, "Foo" . $suffix`, `wrapper(bless {}, "Foo")`, `bless::class {}, "Foo"`). Fallback explicitly does not fire for `Dynamic`.
- Bounded source. Only `used_modules` ∪ `{current_package}` are eligible. All-workspace fallback is *not* used.
- No public API expansion. `add_workspace_method_completions` is module-private; the new `used_modules` parameter is private to `perl-lsp-rs-core`.
- Tier 6 below all exact tiers. Fallback `sort_text` is `6_<name>`; exact-receiver completions stay at tiers 1–5.
- Detail label honest. `workspace method — receiver: unknown, low confidence (from <pkg>)` — both `unknown` and `low confidence` present.
[...]

PR #7956, opened 2026-05-05
test(completion): add type-engine receiver detail edge case (#0000)

EffortlessSteven:
### Motivation
- The completion receiver-evidence tests did not cover the variable-dispatched constructor case (`my $x = $factory->new;`) which can be misclassified as a constructor-assignment; this gap risks regressions in receiver provenance and confidence labeling.

### Description
- Added a focused unit test `detail_marks_type_engine_receiver_as_medium_confidence` to `crates/perl-lsp-rs-core/src/providers/completion/completion/tests.rs` that exercises a variable-dispatched constructor scenario and asserts `TypeEngine` evidence is emitted.
- The test asserts the `detail` text contains `receiver: type engine` and `medium confidence` and explicitly rejects `receiver: constructor assignment` for this fixture.
- Change is test-only and scoped to a single new test in the existing `completion/tests.rs` file.

### Testing
- Attempted to run the focused test via `./scripts/cargo-safe test -p perl-lsp-rs-core --profile agent --locked detail_marks_type_engine_receiver_as_medium_confidence`, but execution was blocked by the environment storage guard (`DENY: /root/.cache/devplane/perl-lsp`), so the test could not be completed here.
- Ran `./scripts/storage-doctor`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69f96298a24c8333a2580389226ed8e4)

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/completion/completion/tests.rs
+++ b/crates/perl-lsp-rs-core/src/providers/completion/completion/tests.rs
@@ -1822,6 +1822,48 @@ $x->"#;
 }
 
 #[test]
+fn detail_marks_type_engine_receiver_as_medium_confidence() -> Result<(), Box<dyn std::error::Error>>
+{
+    // Edge case: constructor assignment pattern does not apply because the
+    // receiver variable is assigned from another variable. This should force
+    // TypeEngine evidence (medium confidence) rather than constructor assignment.
+    let index = Arc::new(WorkspaceIndex::new());
+    index.index_file(
+        Url::parse("file:///workspace/Foo.pm")?,
+        r#"package Foo;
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Closing after post-#7930/#7957 review. This fixture expects an end-to-end TypeEngine receiver path from my  = Foo; my  = ->new;, but current supported inference does not produce an object type for that source, so the added test fails before it can inspect detail text. Current master already pins the relevant contracts through detail_with_evidence_helper_handles_both_base_formats, 	ype_engine_variant_has_medium_confidence, and the constructor-assignment/detail tests. This should come back only with the future TypeEngine object-seeding behavior, not as a standalone test-only PR.

PR #8141, opened 2026-05-07
docs(memory): require pressure signals in retained-state inventory (#0000)

EffortlessSteven:
## Summary

- Make the retained-state inventory require a pressure counter or retained-process signal for every long-lived owner.
- Add a first-class `Pressure counter or signal` column to the current inventory table.
- Extend `cargo xtask check-memory-lifecycle-policy` so the inventory cannot drop the pressure-signal requirement or malformed inventory rows.

## Validation

- `cargo xtask fmt`
- `cargo test -p xtask memory_lifecycle_policy -- --nocapture`
- `cargo xtask check-memory-lifecycle-policy`
- `git diff --check`

Note: `cargo xtask check-doc-index` was not run because this checkout does not define that xtask subcommand.


EffortlessSteven:
Post-merge proof passed on master at `fe055de3929bd1258fb23743b1cc43dc2976e385`:

- `cargo xtask fmt`
- `cargo test -p xtask memory_lifecycle_policy -- --nocapture`
- `cargo xtask check-memory-lifecycle-policy`
- `git diff --check`

Worktree clean.

PR #8219, opened 2026-05-08
docs(tooling): define native formatter and critic replacement contract (#0000)

EffortlessSteven:
## Summary

Adds a docs-only contract for replacing the current external `perltidy` / `perlcritic` adapter paths with native Rust formatter and critic engines over the existing parser, diagnostics, semantic, LSP, config, and receipt surfaces.

The contract defines:
- native vs compat vs external legacy operating modes
- formatter architecture, API shape, IR, and safety invariants
- critic rule registry, finding shape, suppressions, and rule families
- shared config/import expectations for `.perltidyrc` and `.perlcriticrc`
- receipt metrics and done criteria
- phased formatter, critic, and replacement-closeout PR map

Also links the new spec from `docs/INDEX.md`.

## Proof packet

Changed surfaces:
- docs

Required proof:
- [x] cargo xtask fmt
  - why: keeps formatting deterministic before any other proof
  - evidence: passed locally
- [x] cargo xtask doc-claims
  - why: validates documentation claim ledger rules
  - evidence: passed locally: 67 articles scanned, 16 stale patterns checked, 0 violations found
- [x] git diff --check
  - why: guards against whitespace errors in the final patch
  - evidence: passed locally

Additional note:
- `cargo xtask ci-hygiene check-doc-paths docs` was attempted as a docs path check, but it fails on pre-existing absolute-path findings in older research/archive docs unrelated to this PR. No new finding points at `docs/specs/NATIVE_FORMATTER_CRITIC_REPLACEMENT.md` or `docs/INDEX.md`.

## Runtime impact

None. Docs/spec only; no runtime behavior, workflows, gates, parser metrics, or memory policy changes.

PR #8220, opened 2026-05-08
docs(tooling): fix native tooling binary examples (#0000)

EffortlessSteven:
Problem: #8219 merged a docs-only native tooling contract with two command examples using the library crate name instead of the LSP server binary.
Fix: Change the configuration import examples from `perllsp` to `perl-lsp`.
Verification: `cargo xtask fmt --check`, `cargo xtask doc-claims`, and `git diff --check` pass.


EffortlessSteven:
Closing this follow-up after re-checking the binary surface. The Droid note on #8219 treated `perl-lsp` as the only server binary, but `crates/perllsp/Cargo.toml` defines the public `perllsp` binary and the docs/spec examples should stay aligned with the installed public CLI. The `config import-*` shape remains a future contract surface in the spec, not a current-command claim.

Issue #8242, opened 2026-05-08
feat(compile-env): add HIR module-resolution candidate facts


EffortlessSteven:
## Problem

#8206 now has the first HIR `CompileEnvironment` facts and a shared effective include-root builder in `perl-module`, but module requests still stop at `ModuleResolutionStatus::Deferred`.

The next compiler-substrate slice should add candidate/path facts for HIR module requests without changing LSP provider behavior.

## Goal

Record module-resolution candidate facts for static HIR module requests so later compiler and provider work can reason about module lookup with provenance.

## Scope

Add a crate-local/compiler-local fact model for module-resolution candidates keyed to HIR `ModuleRequest` entries.

Expected facts:

- request index or directive item
- target module name
- relative module path (`Foo::Bar` -> `Foo/Bar.pm`)
- ordered candidate roots supplied explicitly by tests/caller
- candidate path/source labels
- status: candidate-built / dynamic / not-found / resolved / timed-out, as supported by the first slice
- provenance and confidence

## Acceptance Criteria

- Static `use Foo::Bar` and `require Foo::Bar` can produce relative path/candidate facts.
- Dynamic `require $module` remains a dynamic boundary and does not produce fake candidates.
- Candidate roots preserve source labels for lexical, configured, PERL5LIB, and system roots.
- The first implementation does not read the process environment and does not spawn `perl` from parser/HIR lowering.
- No LSP provider consumes these facts yet.
- Fixture tests cover lexical `use lib`, configured roots, PERL5LIB-labeled roots, system-labeled roots, and dynamic require fail-closed behavior.

## Non-goals

- No import/export semantics.
- No provider cutover.
- No real-Perl oracle.
- No runtime behavior change.
- No system `@INC` probing from parser/HIR code.

## Verification

```bash
cargo test -p perl-parser-core --test hir_tests --profile agent --locked -- --nocapture
cargo test -p perl-module --lib --profile agent --locked effective_inc_roots -- --nocapture
cargo check -p perl-parser-core --all-targets --profile agent --locked
cargo xtask fmt --check
git diff --check
```

## Parent

Part of #8206.

EffortlessSteven:
Implemented and merged via #8249.

Merge SHA: `8138ec046` (`feat(hir): add module resolution candidates (#8242) (#8249)`).

Post-merge verification from `origin/master` passed:

```bash
cargo test -p perl-parser-core --test hir_tests --profile agent --locked -- --nocapture
[...]

PR #8249, opened 2026-05-08
feat(hir): add module resolution candidates (#8242)

EffortlessSteven:
Problem: HIR records module requests and include-root effects, but compiler fact lanes had no candidate-path fact surface keyed to those requests.
Fix: Add HIR-local module-resolution candidate facts built from static requests, active lexical `use lib` roots, and caller-supplied configured/PERL5LIB/system roots, while leaving dynamic `require` candidate-free and avoiding provider cutover.
Verification:
- `cargo test -p perl-parser-core --test hir_tests --profile agent --locked -- --nocapture` passes
- `cargo test -p perl-module --lib --profile agent --locked effective_inc_roots -- --nocapture` passes
- `cargo check -p perl-parser-core --all-targets --profile agent --locked` passes
- `cargo clippy -p perl-parser-core --lib --profile agent --locked -- -D warnings -A missing_docs` passes
- `cargo xtask fmt --check` passes
- `git diff --check` passes

Note: `cargo clippy -p perl-parser-core --all-targets --profile agent --locked -- -D warnings -A missing_docs` is blocked by pre-existing unrelated parser-core test lint debt (`panic`/`expect`/`assert!(false)` in older test files), so this PR used lib clippy plus all-targets check.

### DIFF
--- a/crates/perl-parser-core/src/hir/mod.rs
+++ b/crates/perl-parser-core/src/hir/mod.rs
@@ -16,8 +16,9 @@ pub use model::{
     GlobSlotKind, GlobSlotSource, HirBindingId, HirFile, HirId, HirItem, HirKind, HirScopeId,
     IncRootAction, IncRootFact, IncRootKind, IndirectCallExpr, InheritanceSource, LiteralExpr,
     LiteralKind, MethodCallExpr, MethodDecl, ModuleRequest, ModuleRequestKind,
-    ModuleResolutionStatus, PackageDecl, PackageInheritanceEdge, PackageStash, PragmaEffect,
-    RecoveryConfidence, RequireDecl, ScopeFrame, ScopeGraph, ScopeKind, StashConfidence,
-    StashDynamicBoundary, StashDynamicBoundaryKind, StashGraph, StashProvenance, StorageClass,
-    SubDecl, UseDecl, VariableBinding, VariableDecl,
+    ModuleResolutionCandidate, ModuleResolutionCandidateRoot, ModuleResolutionCandidateStatus,
+    ModuleResolutionRoot, ModuleResolutionStatus, PackageDecl, PackageInheritanceEdge,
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
Follow-up on automated review:

- Fixed `use lib` precedence in `8ead6ec22`: later `use lib` directives now outrank earlier ones, while same-directive `qw(first second)` order stays stable. Added `hir_module_resolution_candidate_facts_match_use_lib_precedence`.
- Dynamic `require` remains candidate-free by design for #8242. The dynamic request and boundary are still recorded in HIR; candidate facts only describe static paths so providers do not get fake module candidates.

Re-verified locally after the amendment:
- `cargo test -p perl-parser-core --test hir_tests --profile agent --locked -- --nocapture`
- `cargo test -p perl-module --lib --profile agent --locked effective_inc_roots -- --nocapture`
- `cargo check -p perl-parser-core --all-targets --profile agent --locked`
- `cargo clippy -p perl-parser-core --lib --profile agent --locked -- -D warnings -A missing_docs`
- `cargo xtask fmt --check`
- `git diff --check`

EffortlessSteven:
Second follow-up after Droid security validation:

- Hardened path-like static targets so traversal/absolute/drive-style require targets do not produce candidate paths.
- Added `hir_module_resolution_candidate_facts_reject_path_traversal_targets`.

Re-verified locally on `f3fbe79f2`:
- `cargo test -p perl-parser-core --test hir_tests --profile agent --locked -- --nocapture`
- `cargo test -p perl-module --lib --profile agent --locked effective_inc_roots -- --nocapture`
- `cargo check -p perl-parser-core --all-targets --profile agent --locked`
- `cargo clippy -p perl-parser-core --lib --profile agent --locked -- -D warnings -A missing_docs`
- `cargo xtask fmt --check`
- `git diff --check`

Issue #8252, opened 2026-05-08
feat(imports): project HIR import facts into ImportSpec


EffortlessSteven:
## Parent

Parent: #8244
Depends on: #8242

## Problem

The repo already has the durable `perl_semantic_facts::ImportSpec` vocabulary and workspace visibility consumers, but the compiler-backed path does not yet project HIR `use` / `require` evidence into canonical import facts.

## Goal

Add a fixture-backed HIR import projection that emits canonical `ImportSpec` facts with source range/order metadata and no LSP provider cutover.

## Scope

- Reuse existing `ImportSpec`, `ImportKind`, `ImportSymbols`, `Provenance`, and `Confidence` types.
- Consume HIR `UseDecl`, `RequireDecl`, `CompileDirective`, `ModuleRequest`, source ranges, package/scope context, and module-resolution candidate work.
- Cover static `use Module`, `use Module ()`, explicit `qw(...)`, tag, mixed tag/name, `use constant`, and static `require` shapes.
- Preserve directive ordering and `span_start_byte` for order-aware visible-symbol proof.
- Emit or preserve dynamic-boundary behavior for dynamic `require` / dynamic import args instead of guessing.
- Do not change LSP provider behavior.

## Non-goals

- No `ExportSet` projection; track that separately.
- No framework adapter registry.
- No real-Perl oracle or ambient `import` execution.
- No provider cutover.

## Acceptance criteria

- HIR-backed static imports produce canonical `ImportSpec` facts with module, kind, symbols/tags, provenance, confidence, source anchor, and byte ordering.
- `use Module ();` records no-import behavior and does not invent visible symbols.
- Static `require Module` and dynamic `require $target` remain distinguishable.
- The implementation does not introduce a second semantic vocabulary for imports.
- Existing semantic scorecard and shadow-compare import rows remain green.

## Verification

```bash
cargo test -p perl-parser-core --test hir_tests --profile agent --locked import -- --nocapture
cargo xtask semantic-scorecard --check
cargo xtask semantic-shadow-compare --check
cargo xtask fmt --check
git diff --check
```

EffortlessSteven:
Closed by #8256 at 2d49e8b77.

Post-merge verification from origin/master:

```bash
cargo test -p perl-parser-core --test hir_tests --profile agent --locked import -- --nocapture
cargo xtask semantic-scorecard --check
cargo xtask semantic-shadow-compare --check
[...]

PR #8256, opened 2026-05-08
feat(hir): project import directives into semantic facts (#8252)

EffortlessSteven:
Problem: HIR compile-environment facts record `use` / `require` directives, but there was no canonical `ImportSpec` projection for the compiler-backed path.
Fix: add an opt-in `CompileEnvironment::import_specs(file_id)` projection that reuses `perl-semantic-facts` import types, preserves anchors/scope/order, and keeps dynamic import args / dynamic require fail-closed. No provider cutover.
Verification: `cargo test -p perl-parser-core --test hir_tests --profile agent --locked import -- --nocapture` passes; `cargo xtask semantic-scorecard --check` passes; `cargo xtask semantic-shadow-compare --check` passes; `cargo check -p perl-parser-core --all-targets --profile agent --locked` passes; `cargo clippy -p perl-parser-core --lib --profile agent --locked -- -D warnings -A missing_docs` passes; `cargo xtask fmt --check` passes; `git diff --check` passes.

Note: an exploratory full `cargo test -p perl-parser-core --profile agent --locked` still fails in existing unrelated `fix_subst_squote_delimiter` parser tests for substitution replacement parsing; this PR does not touch that parser lane.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1779,6 +1779,7 @@ dependencies = [
  "perl-position-tracking",
  "perl-pragma",
  "perl-regex",
+ "perl-semantic-facts",
  "perl-tdd-support",
  "perl-token",
  "proptest",
--- a/crates/perl-parser-core/Cargo.toml
+++ b/crates/perl-parser-core/Cargo.toml
@@ -29,6 +29,7 @@ perl-ast = { workspace = true }
 perl-ast-v2 = { workspace = true }
 perl-pragma = { workspace = true }
 perl-regex = { workspace = true }
+perl-semantic-facts = { workspace = true }
 perl-token = { workspace = true }
 # Wave D absorbed (internal modules in src/syntax/):
 #   perl-quote → src/syntax/quote.rs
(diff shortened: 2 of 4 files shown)


EffortlessSteven:
Addressed Droid's validated findings in the latest push:

- `qw` parsing now handles `()`, `[]`, `{}`, `<>`, and same-delimiter forms rather than only `qw(...)`.
- Unquoted `&` and `*` import args now fail closed as dynamic import arguments, matching `$`, `@`, and `%`.
- Added HIR tests for bracket/brace `qw` delimiters and unquoted `&` / `*` import args.

Local verification after the fix:

```bash
cargo xtask fmt --check
cargo test -p perl-parser-core --test hir_tests --profile agent --locked import -- --nocapture
cargo xtask semantic-scorecard --check
cargo xtask semantic-shadow-compare --check
cargo check -p perl-parser-core --all-targets --profile agent --locked
cargo clippy -p perl-parser-core --lib --profile agent --locked -- -D warnings -A missing_docs
git diff --check
```

PR #8498, opened 2026-05-10
perf(completion): prefix-directed module scan for namespaced prefixes (#8491)

### LINKED ISSUE #8491: Module completion: prefix-directed scan with short TTL cache
## Problem

Module-name completion bounds tree scans but starts from the include root on every keystroke. For prefixes like `MyApp::Con`, scanning `root/MyApp/` is sufficient; scanning `root/` (BFS, bounded) and filtering is wasteful, especially with large `vendor/`, `local/`, and system-`@INC` trees. No cross-keystroke cache means typing `use My|` → `use MyA|` → `use MyAp|` → `use MyApp|` re-scans the same subtree N times.

Current bounds (verify exact names/locations in `crates/perl-lsp-rs/src/completion/` or `crates/perl-lsp-rs/src/runtime/language/completion.rs`): `MAX_MODULE_SCAN_ROOTS = 16`, `MAX_MODULES_PER_SCAN = 512`, `MAX_SCAN_DEPTH = 8` (per the cleanup plan; values should be confirmed by spec-planner).

## Scope

1. **Prefix-directed scan.** Split `Foo::Bar::Ba` into `(root.join("Foo/Bar"), "Ba")`:

   ```rust
   fn root_and_leaf_prefix(root: &Path, module_prefix: &str) -> (PathBuf, String)
   ```

   Scan only the prefix directory; rebuild full module names by combining traversed subdirs.

2. **Tiny TTL cache.** Optional but recommended. Key on `(canonical_root, prefix_directory_relative)`. Defaults:

   ```rust
   const MODULE_COMPLETION_CACHE_TTL_MS: u64 = 1000;
   const MODULE_COMPLETION_CACHE_MAX_ENTRIES: usize = 128;
   ```

   TTL-only invalidation (no filesystem watcher dependency). Cancellable path remains cancellable.

3. Preserve symlink-loop protection and existing dedup behavior.

## Acceptance criteria

- Identical completion labels on identical inputs vs. master.
- Wall-clock improvement on a fixture with `MAX_MODULE_SCAN_ROOTS` trees of 10k+ modules each, for prefixes like `Mojo::Cont`.
- No symlink traversal regression.
- Cache TTL of 1s does not produce stale results in interactive use (manual smoke test acceptable).

## References

- Part of: `@INC` / workspace includes cleanup plan (PR 7 of 8)
- Performance-only; behavior-preserving. Land after PR 1–6.

EffortlessSteven:
## Summary

For `use Foo::Bar::Ba` completions, the module scanner previously started a BFS from each include root and filtered all discovered module names by the typed prefix. This PR adds a `root_and_leaf_prefix()` helper that splits a namespaced prefix like `"Foo::Bar::Ba"` into a subdir path (`root/Foo/Bar/`) and a leaf prefix (`"Ba"`), then starts the BFS at the narrowest matching directory instead of the root.

Single-segment prefixes (e.g., `"DBI"`) are unaffected — they continue to scan from root.

## Algorithm sketch

```
prefix = "Foo::Bar::Ba"
  → split on "::" → ["Foo", "Bar", "Ba"]
  → pop leaf  → leaf = "Ba", remaining = ["Foo", "Bar"]
  → scan_dir  = root.join("Foo").join("Bar")
  → depth_consumed = 2

BFS starts at scan_dir with initial depth = 2.
path_to_module_name(root, path) still uses the original root,
so "Foo/Bar/Baz.pm" → "Foo::Bar::Baz" (full name, not just "Baz").
Filter: module_name.starts_with("Foo::Bar::Ba") still works unchanged.
```

Fallback: if the computed subdir does not exist (prefix has no matching directory), falls back to root scan — same as before, just no speedup for that case.

## Behavior-preservation argument

The only change is which directory the BFS starts in. `path_to_module_name(root, path)` is called with the unchanged `root`, so it reconstructs full qualified module names (e.g., `"Foo::Bar::Baz"`) identically. The `starts_with(prefix)` filter at line ~483 is unchanged. Dedup (`HashSet<String>`), symlink protection (`!path.is_symlink()`), bounds (`MAX_MODULES_PER_SCAN`, `MAX_SCAN_DEPTH`), and the module ordering are all preserved. The `depth_consumed` offset ensures MAX_SCAN_DEPTH is still respected relative to the original root depth, not just the subdir.

## Test plan

- [x] `test_root_and_leaf_prefix_single_segment` — single-segment prefix keeps root as scan dir
- [x] `test_root_and_leaf_prefix_multi_segment` — multi-segment prefix produces correct subdir and depth
- [x] `test_root_and_leaf_prefix_empty` — empty prefix falls through cleanly
- [x] `test_scan_prefix_directed_identical_to_root_scan` — fixture tree confirms `Foo::Bar::Ba` returns only matching modules; `Foo::Other` and `Unrelated::Module` excluded
- [x] `test_scan_prefix_directed_nonexistent_subdir_returns_empty` — missing intermediate subdir returns `[]`, no panic
- [x] `test_use_completion_namespaced_prefix_directed` — end-to-end via `add_use_module_completions`: `Mojo::Controller` appears, `Mojo::Util` does not, for prefix `"Mojo::Co"`
- [x] All 1622 existing `perl-lsp-rs-core` tests pass
- [x] All 27 `perl-lsp-rs` completion tests pass

## Notes

This is PR 7a (scan only). PR 7b (short TTL cache) is a follow-up — the cache must live in `LspServer` state, not inside `CompletionProvider` which is reconstructed per request.

Closes #8491 (7a — scan only).


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder completed prefix-directed module scan (PR 7a). 1 PR, 6 tests added, 2 files changed.

### Ending state

- **PR created:** https://github.com/EffortlessMetrics/perl-lsp/pull/8498
- **Tests added:** 6 new tests in `perl-lsp-rs-core`
  - `test_root_and_leaf_prefix_single_segment`
  - `test_root_and_leaf_prefix_multi_segment`
  - `test_root_and_leaf_prefix_empty`
  - `test_scan_prefix_directed_identical_to_root_scan`
  - `test_scan_prefix_directed_nonexistent_subdir_returns_empty`
  - `test_use_completion_namespaced_prefix_directed`
- **Files changed:**
  - `crates/perl-lsp-rs-core/src/providers/completion/completion/workspace.rs` (+54/-1)
  - `crates/perl-lsp-rs-core/src/providers/completion/completion/tests.rs` (+159)
- **Tests passing:** 1622/1622 `perl-lsp-rs-core` lib tests, 27/27 `perl-lsp-rs` completion tests
- **Next step:** Ready for review

### Retrospective

The implementation was simpler than expected once I saw that `path_to_module_name(root, path)` already reconstructed full module names from the original root — so the only change needed was starting the BFS at the narrower prefix subdir, with no changes to the name reconstruction or filter logic. The key insight: the existing `starts_with(prefix)` filter is already correct; the only waste was scanning irrelevant sibling subtrees first.

One gotcha worth noting: when the computed subdir does not exist (prefix has no real directory match), the code falls back to scanning from `root`. This is correct because the `starts_with` filter will return empty anyway, but it's slightly wasteful. The fallback could be an early-return `return modules;` for the case where `scan_dir` doesn't exist — but this would be a micro-optimization on an already-rare path (user typed a namespace with no matching directory), so it was left as a fallback scan for robustness.

The `depth_consumed` offset in `start_depth` is load-bearing: without it, BFS starting at `root/Foo/Bar/` with depth=0 could descend up to `MAX_SCAN_DEPTH` more levels, making the total depth `depth_consumed + MAX_SCAN_DEPTH` rather than `MAX_SCAN_DEPTH`. The offset keeps the absolute depth cap correct.

### Breadcrumbs for next agent (PR 7b — TTL cache)

- **Logical next:** PR 7b — short TTL cache keyed on `(canonical_root, prefix_directory_relative)`
- **Critical constraint from addendum:** `CompletionProvider` is reconstructed per request — cache must live in `LspServer` or another runtime-owned struct, NOT inside the provider
- **Cache knobs (review at impl time):** `MODULE_COMPLETION_CACHE_TTL_MS = 1000`, `MODULE_COMPLETION_CACHE_MAX_ENTRIES = 128`
- **Cache key:** `(canonical_root, prefix_directory_relative)` — the `root_and_leaf_prefix()` helper in this PR already computes the subdir, making cache key construction trivial
- **Confidence for next:** high — the 7a changes are clean and the 7b entry point is clear

PR #8544, opened 2026-05-11
fix(@INC): filter workspace-symbol lookups by EffectiveIncContext (#8537)

### LINKED ISSUE #8537: fix(@INC): thread @INC context into workspace-symbol lookups for no-lib strictness
## Problem

After PR #8525 closed #8516, PL701 / push-diagnostics / pull-diagnostics all honor position-aware `use lib` / `no lib` semantics. **goto-definition and completion do not.** They still resolve modules whose `@INC` path was cancelled by `no lib` before the cursor.

## Root cause (from #8525 builder's handoff)

The workspace indexer (started at workspace open) scans the filesystem for module files independently of `@INC` state. Both goto-def and completion check the workspace symbol index **before** checking `@INC`-constrained filesystem paths. So even when `EffectiveIncContext` says a path is cancelled at the cursor offset, the indexed module is still surfaced.

Concretely, in the scenario_14 fixture:

```perl
use lib 'lib';
no lib 'lib';
use GoneModule;   # ← cursor here
```

with `lib/GoneModule.pm` on disk:

- **PL701 / push diagnostics**: do NOT find GoneModule (correct). ✓
- **PL701 / pull diagnostics**: do NOT find GoneModule (correct after #8525). ✓
- **goto-definition**: STILL resolves to `lib/GoneModule.pm` via the workspace symbol index. ✗
- **completion**: STILL suggests `GoneModule` from the workspace symbol index. ✗

## Scope

Thread `@INC` context (specifically the per-offset `EffectiveIncContext`) into the workspace-symbol resolution path for goto-def and completion. Estimated >200 LoC touching resolver internals — explicitly larger than #8516's scope.

Possible approaches (decide at plan-review time):

1. **Filter at lookup time**: when goto-def / completion query the workspace symbol index, additionally check that the symbol's file URI is reachable through the cursor-position `EffectiveIncContext`. Cheap per query, no index changes.
2. **Tag symbols with origin path**: when the indexer walks `<root>/lib/Foo.pm`, record the include-root that produced the symbol. At lookup time, intersect against the active context.
3. **Hybrid**: use approach 1 for the negative `no lib` case (rare), approach 2 if perf needs it.

Approach 1 is the smallest valuable next step.

## Acceptance

- `scenario_14_no_lib_cancellation` enforces all four consumers strictly (currently the strict assertions for goto-def and completion are passing because the test's specific harness setup excludes the workspace index — this issue's fix should make them robust under the production indexer path).
- Add a regression that opens a workspace with `lib/GoneModule.pm` present, runs the workspace indexer, and confirms goto-def / completion still respect `no lib` cancellation.
- No regression to existing scenario_14 cases.

## References

- Builder handoff for #8525: identified this as the remaining gap.
- #8516 (closed by #8525) — the position-aware diagnostic side.
- `EffectiveIncContext` rail: #8489/#8504/#8505/#8506.
- Original spec for #8516: see issue body in #8516 for the four-consumer parity matrix.

EffortlessSteven:
## Summary

Closes the last @INC strictness gap from the #8516 / #8525 rail. goto-def and module completion now respect no-lib cancellations even when the workspace indexer has indexed the cancelled file.

Approach 1 from issue #8537: filter at lookup time, no index schema change.

## Changes

- inc_context.rs: New symbol_uri_reachable method on EffectiveIncContext + 3 unit tests.
- navigation.rs: Split EarlyDefinitionTarget into UseModule (authoritative file-system, early-return) vs Module (Package-> fallback preserved).
- completion.rs: Detects use/require context and filters Package/Class/Role workspace symbols through ctx.symbol_uri_reachable.
- ux_scenario_14_inc_conformance.rs: Hard asserts + 2 new tests.

## Evidence
- Commits: 1
- Tests: 597 passed, 4 new unit tests, 2 new UX scenario tests
- Lint: cargo clippy -p perl-lsp-rs --lib -- -D warnings clean
- Files: 4 changed, +447/-23 lines

## References
- Closes #8537
- Builds on #8525 (position-aware no lib in PL701)

### DIFF
--- a/crates/perl-lsp-rs/src/runtime/language/completion.rs
+++ b/crates/perl-lsp-rs/src/runtime/language/completion.rs
@@ -219,6 +219,7 @@ impl LspServer {
         &self,
         completions: &mut Vec<crate::completion::CompletionItem>,
         doc_text: &str,
+        doc_uri: &str,
         offset: usize,
         workspace_mode: &IndexAccessMode,
         cap: usize,
@@ -244,6 +245,22 @@ impl LspServer {
                     .rev()
                     .collect::<String>();
 
+                // Detect `use Module` / `require Module` context so we can gate
+                // Package-kind completions through @INC reachability (fixes #8537).
+                let is_use_module_context = {
+                    let before_prefix =
+                        &text_before[..text_before.len().saturating_sub(prefix.len())];
+                    let trimmed = before_prefix.trim_end();
+                    trimmed.ends_with("use") || trimmed.ends_with("require")
+                };
+
+                // Build @INC context once (only when needed for filtering).
+                let inc_ctx = if is_use_module_context {
+                    self.effective_inc_context_for_doc(Some(doc_uri), Some(doc_text), Some(offset))
+                } else {
+                    None
+                };
+
                 let qualified_variable_symbols =
                     Self::qualified_variable_workspace_symbols(index, &prefix);
                 let replace_prefix_range = (offset.saturating_sub(prefix.len()), offset);
@@ -262,6 +279,28 @@ impl LspServer {
 
                     if seen.contains(&symbol.name) {
                         continue;
+                    }
+
+                    // For module-kind symbols in a `use Module` / `require Module`
+                    // context, filter by position-aware @INC reachability so that
+                    // `no lib` cancellations are honoured (fixes #8537).
+                    let is_module_kind = matches!(
+                        symbol.kind,
+                        crate::workspace_index::SymbolKind::Package
+                            | crate::workspace_index::SymbolKind::Class
+                            | crate::workspace_index::SymbolKind::Role
+                    );
+                    if is_use_module_context && is_module_kind {
+                        if let Some(ref ctx) = inc_ctx {
+                            if !ctx.symbol_uri_reachable(&symbol.uri) {
+                                tracing::trace!(
+                                    symbol = %symbol.name,
+                                    uri = %symbol.uri,
+                                    "completion: skipping workspace symbol not reachable via @INC"
+                                );
+                                continue;
+                            }
+                        }
                     }
 
                     let label = symbol.name.clone();
@@ -480,6 +519,7 @@ impl LspServer {
                     self.add_runtime_workspace_completions(
                         &mut completions,
                         &doc.text,
+                        uri,
                         offset,
                         &workspace_mode,
                         cap,
@@ -730,6 +770,7 @@ impl LspServer {
                 self.add_runtime_workspace_completions(
                     &mut completions,
                     &doc.text,
+                    uri,
                     offset,
                     &workspace_mode,
                     completion_cap(),
--- a/crates/perl-lsp-rs/src/runtime/language/navigation.rs
+++ b/crates/perl-lsp-rs/src/runtime/language/navigation.rs
@@ -176,6 +176,12 @@ fn find_label_declaration_span(
 
 #[derive(Debug, Clone)]
 enum EarlyDefinitionTarget {
+    /// Cursor is on a `use Module` / `require Module` statement.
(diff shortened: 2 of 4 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary:** Builder implemented `@INC` filter for workspace-symbol goto-def and module completion. PR #8544 created, 4 files changed, 597 lib tests passing, 6 new tests added.

---

### Ending State

- **PR created:** https://github.com/EffortlessMetrics/perl-lsp/pull/8544
- **Tests added/modified:**
  - `runtime::lifecycle::inc_context::tests::symbol_uri_reachable_returns_true_for_symbol_under_inc_root`
  - `runtime::lifecycle::inc_context::tests::symbol_uri_reachable_returns_false_after_no_lib_cancellation`
  - `runtime::lifecycle::inc_context::tests::symbol_uri_reachable_returns_true_for_non_file_uri`
  - `ux_scenario_14_no_lib_cancellation` (strengthened: soft eprintln → hard assert)
  - `ux_scenario_14_no_lib_cancellation_workspace_index` (new)
  - `ux_scenario_14_use_lib_with_workspace_index` (new, positive control)
- **Files changed:**
  - `crates/perl-lsp-rs/src/runtime/lifecycle/inc_context.rs` — new `symbol_uri_reachable` method
  - `crates/perl-lsp-rs/src/runtime/language/navigation.rs` — `UseModule` enum variant, early-return when @INC fails
  - `crates/perl-lsp-rs/src/runtime/language/completion.rs` — @INC filter for Package/Class/Role completions in `use`/`require` context
  - `crates/perl-lsp-ux-tests/tests/ux_scenario_14_inc_conformance.rs` — strict assertions + 2 new scenarios
- **Tests passing:** 597 passed (perl-lsp-rs --lib), 0 failed
- **Next step:** ready for review

---

### Retrospective

**What was harder than expected:** The initial approach of filtering by `symbol_path.starts_with(inc_root)` was incorrect because the default workspace config includes `.` (workspace root) as an include path, which is a prefix of ALL files under the workspace. Had to add a workspace-root skip to the `symbol_uri_reachable` check. This required a failing unit test to catch — the design looked correct but the default config has `.` as a catch-all root.

**What would I do differently:** Write the unit test for `symbol_uri_reachable` BEFORE implementing the filter logic. The test caught the workspace-root edge case immediately, but I wrote it after the implementation.

**What surprised me:** The workspace index lookup for `use Module` goto-def was a workspace-index fallback after the file-system resolver, not a parallel path. The `EarlyDefinitionTarget::Module` enum was shared between `use Module` (authoritative file-system path) and `Package->` (legitimate workspace index fallback). Splitting them with `UseModule` vs `Module` was the minimal correct fix without touching the index schema.

**Context that would have helped:** Knowing that the default `WorkspaceConfig::default()` includes `["lib", ".", "local/lib/perl5"]` as include paths from the start would have saved one debugging iteration. The `.` root as a default is a gotcha for any `starts_with` filtering logic.

---

- **Logical next:** reviewer → deep-reviewer → ops merge
- **Related issues:** #8537 (this), #8516/#8525 (prior @INC rail), #8504/#8505/#8506 (EffectiveIncContext origin)
- **Gotcha:** `WorkspaceConfig::default()` includes `"."` in `include_paths`, which resolves to the workspace root. Any filter that checks `symbol_path.starts_with(inc_root)` must skip roots equal to the workspace root, or it will always return `true` for workspace-local files.
- **Gotcha:** The UX tests for scenario_14 skip gracefully when the LSP binary is not built. The new strict assertions will only run in CI where the binary is available.
- **Trap:** `module_completion_roots_for_doc` and `add_runtime_workspace_completions` serve different roles — the former feeds the file-system scanner, the latter supplements with workspace-indexed symbols. Only the latter needed the @INC filter; the former was already correct.
- **Confidence for next:** high

Issue #8848, opened 2026-05-13
test(workspace): record current Mojolicious editor baseline


EffortlessSteven:
Problem: The Real Perl Editor Trust active manifest marks real-workspace-baseline-run ready, but the current lane has not recorded a current Mojolicious editor baseline receipt tied to the new spec and provider confidence status. Scope: run the Mojolicious real-workspace baseline and record/update only the receipt or reporting docs needed to state covered and deferred provider surfaces. Claim boundary: the receipt proves only the selected project fixture and host system; it is not all-CPAN proof and does not broaden live provider behavior. Proof: just real-workspace-baseline mojolicious windows; cargo test -p perl-lsp-rs --test real_project_latency mojolicious -- --include-ignored --nocapture; cargo xtask semantic-scorecard --check; cargo xtask semantic-shadow-compare --check; git diff --check.

EffortlessSteven:
Completed by merged PRs #8850 and #8856 (Mojolicious real-workspace baseline and refresh receipt).

PR #8853, opened 2026-05-13
test(workspace): record Mojolicious editor baseline (#8848)

EffortlessSteven:
Problem: #8850 landed the Mojolicious real-workspace baseline with a receipt tied to an older parser-fixture commit and the report generator could leave an extra blank line after outlier rows.
Fix: Refresh the Mojolicious receipt from current master and trim the generated outlier list before writing the forensic report.
Verification: `scripts/real-workspace-baseline.sh mojolicious windows`; `cargo test -p perl-lsp-rs --test real_project_latency test_real_project_latency_baseline_schema --profile agent --locked -- --nocapture`; `cargo xtask semantic-scorecard --check`; `cargo xtask semantic-shadow-compare --check`; `cargo xtask fmt --check`; `git diff --check`; `cargo xtask metrics parser-accuracy --check`; `cargo xtask update-status --only parser --check`; `cargo xtask metrics ratchet-check parser_accuracy`.

### DIFF
--- a/.ci/metrics/real_project_latency.json
+++ b/.ci/metrics/real_project_latency.json
@@ -1,40 +1,39 @@
 {
-  "schema_version": 1,
-  "measured_at": "2026-04-28T00:00:00Z",
-  "commit": "5c89a9371",
+  "commit": "a187da840",
+  "measured_at": "2026-05-13T11:59:52Z",
   "projects": {
-    "mojolicious": {
-      "file_count": 13,
+    "catalyst": {
+      "file_count": null,
       "metrics": {
         "cold_start_to_hover": {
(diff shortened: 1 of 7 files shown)


EffortlessSteven:
Closing as superseded. #8850 already merged the Mojolicious baseline, and #8852 merged the provider confidence matrix plus manifest/plan updates. This branch is now conflicting with master and would drop the provider matrix/status changes if merged. The only unique delta relative to #8850 is a two-file timing refresh measured at a187da840, which is now stale relative to current master (7ee5cec). If we need a newer latency receipt, it should be rerun as a narrow refresh from current master.

Issue #8197, opened 2026-05-07
feat(lsp): gate provider cutover behind compiler fact proof


EffortlessSteven:
## Problem

The compiler substrate only creates user value when diagnostics, completion, hover, navigation, and refactoring consume proven facts. Cutting over too early can preserve bad facts.

Umbrella: EffortlessMetrics/perl-lsp-swarm#2737
Tracking doc: EffortlessMetrics/perl-lsp#8190
Related existing proof issues: EffortlessMetrics/perl-lsp-swarm#1741, EffortlessMetrics/perl-lsp-swarm#2744

## Goal

Track staged provider cutover from parser/workspace approximations to compiler facts.

## Order

- [x] Fact-source tracing receipt/schema contract: EffortlessMetrics/perl-lsp#8305
- [x] Diagnostics shadow proof and narrow live cutover: EffortlessMetrics/perl-lsp#8319, EffortlessMetrics/perl-lsp#8327
- [x] Completion visible-symbol shadow proof: EffortlessMetrics/perl-lsp#8322 / EffortlessMetrics/perl-lsp#8342
- [x] Hover provenance and confidence proof: EffortlessMetrics/perl-lsp#8323 / EffortlessMetrics/perl-lsp#8344
- [x] Definition/reference consume ranked compiler candidates: EffortlessMetrics/perl-lsp#8324 / EffortlessMetrics/perl-lsp#8349
- [x] Rename/safe-delete consume dynamic-boundary blockers: EffortlessMetrics/perl-lsp#8325 / EffortlessMetrics/perl-lsp#8351
- [x] Workspace-symbol source/freshness proof: EffortlessMetrics/perl-lsp#8353 / EffortlessMetrics/perl-lsp#8357
- [x] Document-symbol source/freshness proof: EffortlessMetrics/perl-lsp#8359 / EffortlessMetrics/perl-lsp#8365
- [x] Semantic-token source/freshness proof: EffortlessMetrics/perl-lsp#8360 / EffortlessMetrics/perl-lsp#8367

## Next provider cutover waves

- [x] Hover live provenance expansion: EffortlessMetrics/perl-lsp#8369 / EffortlessMetrics/perl-lsp#8372 moved runtime hover to live compiler fact provenance labels for proven paths while preserving fallback behavior.
[...]

PR #9020, opened 2026-05-15
fix(parser): scan byte zero in find_statement_start (#8197)

EffortlessSteven:
## Summary

`find_statement_start` in `crates/perl-parser/src/ast_utils.rs` had an off-by-one in its backwards scan: the loop guard `while i > 0` never inspected byte 0, so when the only statement terminator (`;` or `\n`) sat at the very start of the source the function fell through to `return 0` instead of returning `1`.

```rust
// Before
let mut i = pos.saturating_sub(1);
while i > 0 {                      // skips index 0
    if bytes.get(i).is_some_and(|b| *b == b';' || *b == b'\n') {
        return i + 1;
    }
    i = i.saturating_sub(1);
}
0
```

```rust
// After
let bytes = source.as_bytes();
let start = pos.min(bytes.len());
for i in (0..start).rev() {        // inclusive of index 0
    if bytes[i] == b';' || bytes[i] == b'\n' {
        return i + 1;
    }
}
0
```

Concrete cases the fix recovers:

| input | pos | before | after |
|---|---|---|---|
| `";x"` | 1 | 0 | 1 |
| `"\nfoo"` | 1 | 0 | 1 |

The function is used by `quick_fixes.rs` and `refactors.rs` to pick insertion points for code-action edits, so the impact is "declarations inserted at byte 0 instead of byte 1" when a buffer starts with a terminator — uncommon in real Perl, but still wrong.

The rewrite is also slightly clearer: the inclusive reverse range removes both the `saturating_sub` dance and the `bytes.get(i)` defensiveness (the range guarantees in-bounds), while `pos.min(bytes.len())` handles the previously-implicit "pos beyond source" case.

## Test plan

- [x] `cargo test -p perl-parser --lib ast_utils` — 6 passed (4 new + 2 existing)
- [x] `cargo test -p perl-parser --test facade_pattern_edge_cases` — 23 passed
- [x] `cargo clippy -p perl-parser --lib --tests` — clean
- [x] `cargo fmt -p perl-parser --check` — clean

New regression tests cover: terminator at index 0, no terminator, empty source, and `pos` beyond `source.len()`.

---
_Generated by [Claude Code](https://claude.ai/code/session_019WctEoFhnHaPGyottdo89Y)_

### DIFF
--- a/crates/perl-parser/src/ast_utils.rs
+++ b/crates/perl-parser/src/ast_utils.rs
@@ -15,16 +15,18 @@ pub fn find_declaration_position(source: &str, error_pos: usize) -> usize {
 }
 
 /// Find the start of the current statement.
+///
+/// Scans backwards from `pos` for the nearest `;` or `\n` and returns the
+/// byte index immediately after it, or 0 if no boundary is found.
 #[must_use]
 pub fn find_statement_start(source: &str, pos: usize) -> usize {
-    let mut i = pos.saturating_sub(1);
     let bytes = source.as_bytes();
+    let start = pos.min(bytes.len());
 
-    while i > 0 {
-        if bytes.get(i).is_some_and(|b| *b == b';' || *b == b'\n') {
+    for i in (0..start).rev() {
+        if bytes[i] == b';' || bytes[i] == b'\n' {
             return i + 1;
         }
-        i = i.saturating_sub(1);
     }
 
     0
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

overlap checked: Compared against current origin/master after #9082; master still had the old byte-0-skipping find_statement_start loop, so this is not superseded.
current master already contains: no
retained branch: claude/find-fix-bug-SgOzW

what changed: find_statement_start now scans byte index 0 and clamps out-of-range positions before scanning backward; added focused regression tests for byte-zero terminators, no-boundary input, empty source, pos zero, and pos past EOF.

proof run:
- MIN_FREE_GB=20 MAX_USED_PCT=95 CARGO_LOCK_WAIT=900 ./scripts/cargo-safe test -p perl-parser --lib ast_utils --profile agent --locked -- --nocapture
- MIN_FREE_GB=20 MAX_USED_PCT=95 CARGO_LOCK_WAIT=900 ./scripts/cargo-safe check --all-targets -p perl-parser --profile agent --locked
- MIN_FREE_GB=20 MAX_USED_PCT=95 CARGO_LOCK_WAIT=900 ./scripts/cargo-safe clippy -p perl-parser --profile agent --locked -- -D warnings -A missing_docs
- MIN_FREE_GB=20 MAX_USED_PCT=95 CARGO_LOCK_WAIT=900 ./scripts/cargo-safe xtask fmt --check
- MIN_FREE_GB=20 MAX_USED_PCT=95 CARGO_LOCK_WAIT=900 ./scripts/cargo-safe xtask parser-corpus-sweep --receipt
- git diff --check
- ./scripts/storage-doctor

receipt note: parser-corpus-sweep wrote target/receipts/system-corpus-sweep.json; it was removed after capture to restore the repo-local storage invariant. Sweep result: 7095 .pm files, 48 unreadable, 7047 clean readable, 0 dirty, 0 errors.

hosted checks: CI Gate (Merge-Blocking), all CI shards, UX Regression Tests, Windows Required, PR Smoke, LSP Memory Smoke, Compile All Targets, reconcile, validate-title, methodology-gate, Workflow Trigger Lint, GitGuardian, CodeRabbit, and Droid review passed.

merge recommendation: squash merge. This is a narrow parser helper correctness fix with regression coverage and corpus proof.
follow-up: none.

PR #9039, opened 2026-05-16
test(perl-pod): cover percent-encoding, link extraction, and malformed POD recovery

EffortlessSteven:
Problem: `perl-pod` coverage tests included a false `=item` empty-body claim and a weak `=end` skip assertion.
Fix: Make the `=item` fixture truly first in the section body, and assert the `=end` directive text is absent from extracted POD output.
Verification: `cargo test -p perl-pod --test pod_extraction_tests --profile agent --locked` passes; `cargo test -p perl-pod --profile agent --locked` passes; direct `target-current\debug\xtask.exe fmt --check` passes after `cargo xtask fmt --check` exited without diagnostics; `git diff --check` passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1804,6 +1804,9 @@ dependencies = [
 [[package]]
 name = "perl-pod"
 version = "0.14.0"
+dependencies = [
+ "tempfile",
+]
 
 [[package]]
 name = "perl-position-tracking"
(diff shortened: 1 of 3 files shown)


EffortlessSteven:
maintainer-pr + diff-audit clean — test-only diff, 1 test file +193 LOC, plus dev-only `tempfile` dep in Cargo.toml/Cargo.lock. No production code touched. POD edge-case assertions are specific and well-documented. Ready for ops.

---

PR #9041, opened 2026-05-16
test(perl-ci-hygiene): cover doc-path resolution and machine-specific path detection

EffortlessSteven:
Problem: `perl-ci-hygiene` doc-path coverage tests were useful, but the draft built regexes inside individual tests and had a Unix-only absolute-path fixture claim.
Fix: Move reusable regex fixtures behind static lazy initialization, make the absolute-path test use a platform-native absolute path, and rename the `/Users` end-to-end test so it only claims the asserted return-code behavior.
Verification: `cargo test -p perl-ci-hygiene --profile agent --locked` passes; `cargo xtask fmt --check` passes; `git diff --check` passes.

### DIFF
--- a/crates/perl-ci-hygiene/src/commands/doc_paths.rs
+++ b/crates/perl-ci-hygiene/src/commands/doc_paths.rs
@@ -97,3 +97,220 @@ pub(crate) fn has_machine_specific_users_path(line: &str, users_name_path: &Rege
         })
     })
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
maintainer-pr + diff-audit clean — 1 source file modified, but all +244 LOC are inside a `#[cfg(test)] mod tests` block appended at end of file. No production logic changed. Tests use static `LazyLock` regex fixtures correctly. Ready for ops.

---

PR #9117, opened 2026-05-16
fix(parser-pest): preserve foreach loop declarations (#0000)

EffortlessSteven:
### Motivation

- Normalization used to simplify lexical `for/foreach` headers for parsing but dropped the original loop declarator (`my|our|local|state`) from the emitted AST, losing iterator-declaration information in `foreach` nodes.

### Description

- Record loop declarators during source normalization by switching to `LazyLock` regex statics and adding a `loop_decl_scopes: VecDeque<(Arc<str>, Arc<str>)>` to `PureRustPerlParser` to remember `(variable, decl)` pairs. The normalization function is now `fn normalize_source(&mut self, source: &str) -> String` and pushes declarators into the queue. (`crates/perl-parser-pest/src/pure_rust_parser.rs`)
- Restore declarators when constructing foreach-style AST nodes by adding `fn take_normalized_loop_declarator(&mut self, variable: &AstNode) -> Option<Arc<str>>` and using it in both `Rule::foreach_statement` and foreach-style `Rule::for_statement` paths so iterator variables become `VariableDeclaration` nodes when appropriate. (`crates/perl-parser-pest/src/pure_rust_parser.rs`)
- Update the behavior spec to assert that `foreach my $item (@items) { ... }` preserves the `my` iterator and the list shape in the S-expression output. (`crates/perl-parser-pest/tests/behavior_spec_tests.rs`)

### Testing

- Ran `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo test -p perl-parser-pest --profile agent --locked` and the crate tests passed (unit and behavior-spec tests: all reported tests passed).
- Ran `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo check --all-targets -p perl-parser-pest --profile agent --locked`, `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo clippy -p perl-parser-pest --profile agent --locked -- -D warnings -A missing_docs`, and `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo fmt --check`, all of which succeeded.
- `./scripts/cargo-safe check --all-targets -p perl-parser-pest --profile agent --locked` was blocked by the environment storage guard (`DENY: /root/.cache/devplane/perl-lsp ...`), and `just agent-pr-fast` could not be run in this environment because `just` is not installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a08cb034c288333ad9c89cf06753b14)

### DIFF
--- a/crates/perl-parser-pest/src/pure_rust_parser.rs
+++ b/crates/perl-parser-pest/src/pure_rust_parser.rs
@@ -11,7 +11,10 @@ use pest::{
 use pest_derive::Parser;
 use regex::Regex;
 use stacker;
-use std::sync::{Arc, OnceLock};
+use std::{
+    collections::VecDeque,
+    sync::{Arc, LazyLock},
+};
 
 #[derive(Parser)]
 #[grammar = "grammar.pest"]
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as superseded by #9119, which has now been squash-merged as 8dc1285f. #9117 preserved loop declarations by keeping the old normalization and storing side-channel declarator state; #9119 retained the behavior with a narrower grammar-first fix that removes the lossy loop-declaration normalization and keeps the v2 bundle synchronized.

PR #9168, opened 2026-05-16
test(perl-parser-pest): probe v2 against tree-sitter's seven failure categories

EffortlessSteven:
## Summary

Investigation artifact, not a feature. Adds two integration tests in `crates/perl-parser-pest/tests/` that exercise the v2 (Pest) parser against the seven Perl constructs catalogued in `docs/articles/research/TREE_SITTER_BREAKAGE.md` as defeating the v1 (C tree-sitter) parser:

1. `/` regex vs. division
2. Heredoc body deferral
3. `{}` hash-vs-block ambiguity
4. Quote-like operators with arbitrary delimiters
5. Punctuation special variables
6. Indirect object syntax
7. `format` blocks

`seven_categories_probe.rs` records pass/fail on `parser.parse()`. `seven_categories_deep.rs` prints the S-expression for the hard cases and includes a garbage probe (unclosed sub, random punctuation, emoji, JS-style syntax) to distinguish real parses from PEG silent recovery.

## Findings

- All seven categories return `Ok` from `parser.parse()` — but the deep probe shows several produce **structurally complete but semantically wrong** ASTs:
  - `print <<A, <<B; \naaa\nA\nbbb\nB` → heredoc bodies are empty strings; body lines parsed as separate function calls
  - `format STDOUT = ... .` → `(format_declaration )` with empty body
  - `print ${^MATCH}` → variable name truncated to `${`
  - `$s =~ s{foo}/bar/g` → empty replacement, trailing `bar / g` parsed as division
  - `map { a => $_ } @list` → `{...}` parsed as `hash_ref`, not block
- Garbage probe: clearly invalid input correctly rejected; borderline-wrong input (`my @@x = 5`) accepted with wrong AST.
- Implication: v2's "success" on these categories is **silent failure**, worse for LSP use than v1's `ERROR` nodes (which at least signal that fallback is needed).

## Test plan

- [x] `cargo test -p perl-parser-pest --test seven_categories_probe -- --nocapture` — surface pass/fail
- [x] `cargo test -p perl-parser-pest --test seven_categories_deep -- --nocapture` — AST inspection + garbage
- [ ] Decide whether to promote any probes into assertions (currently they only print)
- [ ] Decide whether to file follow-up issues for the silent-failure cases

https://claude.ai/code/session_01FYNpEDbTENdEkEmgCLYecS

---
_Generated by [Claude Code](https://claude.ai/code/session_01FYNpEDbTENdEkEmgCLYecS)_

PR #9170, opened 2026-05-16
test(parser-comparison): differential test suite for v1/v2/v3 parser comparison (#9168)

EffortlessSteven:
## Summary

Adds a new `perl-parser-comparison` crate with **50 structured tests** that measure how each of the three Perl parsers (v1 C tree-sitter, v2 Pest, v3 recursive descent) handles the seven categories of constructs that historically defeated tree-sitter, plus the silent-failure edge cases discovered in PR #9168.

The key framing: the previous probe tests in PR #9168 only checked `Result::is_ok()`. This suite asserts **structural properties of the AST** — "the heredoc body contains the expected content," "the format declaration has body lines," "`${^MATCH}` parses as a single variable whose name includes `MATCH`." This makes the gaps between parsers visible and verifiable rather than masked by error-recovery.

The suite is **not in the default CI gate**. Run on demand with `cargo test -p perl-parser-comparison`.

## Decisions

- **Crate placement**: New crate `crates/perl-parser-comparison` rather than extending any existing parser crate. It depends on all three parsers simultaneously — putting it in any one parser's crate would create an awkward tier dependency.
- **v3 API**: Used `perl-parser-core::Parser` directly (via `parse_with_recovery()`) rather than the `tree-sitter-perl-rs` facade. The facade wraps the same engine but adds semantic overlay overhead not needed for AST structure tests.
- **Verdict model**: Five categories (`Correct`, `WrongButPlausible`, `SilentlyEmpty`, `Errors`, `Crashes`) rather than boolean pass/fail. When a parser improves or regresses, the expected verdict must change intentionally.
- **`catch_unwind`**: Used in the harness to prevent one parser crash from killing the whole suite. The `Crashes` verdict exists for exactly this case.

## Disagreement Table (from running the suite)

| Category | Case | v1 | v2 | v3 |
|----------|------|----|----|-----|
| 0 Control | simple_variable_assignment | Correct | Correct | Correct |
| 0 Control | sub_declaration | Correct | Correct | Correct |
| 1 `/` ambiguity | division_between_terms | Correct | Correct | Correct |
| 1 `/` ambiguity | regex_after_if_keyword | Correct | Correct | Correct |
| 1 `/` ambiguity | regex_after_closing_paren | Correct | Correct | Correct |
| 2 Heredoc | basic_heredoc | Correct | Correct | Correct |
| 2 Heredoc | **multiple_heredocs** | Correct* | Correct* | Correct |
| 2 Heredoc | v3_heredoc_body_populated | — | — | Correct |
| 3 `{}` ambiguity | **map_block_hashlike_content** | **Errors** | Correct* | Correct |
| 3 `{}` ambiguity | eval_block | Correct | Correct | Correct |
| 4 Quote-like | s_paired_braces | Correct | Correct | Correct |
| 4 Quote-like | **s_mixed_delimiters** | Correct | Correct* | Correct |
| 5 Special vars | dollar_slash | Correct | Correct | Correct |
| 5 Special vars | **dollar_caret_match** | Correct* | Correct* | Correct |
| 6 Indirect obj | indirect_new | Correct | Correct | Correct |
| 7 Format | **simple_format_declaration** | **Errors** | Correct* | Correct |
| 7 Format | **format_followed_by_code** | **Errors** | Correct | Correct |
| G Garbage | pure_nonsense | Errors | Errors | Errors |
| G Garbage | unclosed_sub | Errors | Errors | Errors |
| G Garbage | javascript_function | Errors | Correct | Errors |
| G Garbage | **double_sigil_at_at** | Errors | Correct* | Correct |

`*` = parse succeeds (no error nodes) but structural assertion reveals silent failure

**Key findings**:
- **v1 fails on `map { a => $_ }`** — produces ERROR nodes because the GLR grammar cannot resolve hash-vs-block ambiguity in map context
- **v1 fails on `format STDOUT = ... .`** — ERROR nodes for format declarations (simple and followed-by-code)
- **v1 loses second heredoc body** — `print <<A, <<B` produces no ERROR nodes but second body is completely absent
- **v2 loses format body** — `(format_declaration )` with empty body; the atomic Pest rule collapses sub-rules
- **v2 mis-parses `${^MATCH}`** — variable name truncated to `${`; trailing `MATCH}` is lost
- **v2 mis-parses `s{foo}/bar/g`** — mixed-delimiter substitution becomes `s/foo//` with empty replacement; `bar/g` becomes a separate division expression
- **v2 mis-parses `map { a => $_ }`** — parses as `hash_ref` rather than block (WrongButPlausible)
- **v3 handles all cases correctly** — 0 structural failures

## Interface & compatibility

- perl-parser API: unchanged
- LSP surface: unchanged
- DAP surface: unchanged
- CLI: unchanged
- New crate `perl-parser-comparison`: test-only, `publish = false`

## How to review

Start with:
1. `crates/perl-parser-comparison/src/outcomes.rs` — the Verdict enum
2. `crates/perl-parser-comparison/src/harness.rs` — how each parser is invoked
3. `crates/perl-parser-comparison/tests/differential.rs` — the 50 test cases

Hotspots:
- `cat2_multiple_heredocs_on_one_line` — v1 has no error nodes but silently loses the second body
- `cat5_dollar_caret_match_named_capture` — v2 truncates the variable name to `${`
- `cat4_s_mixed_delimiters_brace_slash` — v2 accepts but produces completely wrong structure

## Evidence

```
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
cargo clippy -p perl-parser-comparison --tests -- -D warnings  ->  Finished (no warnings)
cargo xtask fmt  ->  Finished (exit 0)
```

## Risk & rollback

**Blast radius**: Zero — new crate with `publish = false`, not in the default CI gate, no changes to existing parsers or LSP code.

**Rollback**: Remove `"crates/perl-parser-comparison"` from workspace `members` in `Cargo.toml`.

## Follow-ups

- Update expected verdicts when parsers improve (the table is the durable artifact)
- Consider adding `just parser-comparison` target for easy on-demand running
- The `WrongButPlausible` verdict category is defined but not yet exercised in a dedicated assertion — future hardening could add an explicit structural assertion for v2 `map { a => $_ }` parsed-as-hash case

## Source

Investigation: PR #9168 (`claude/investigate-tree-sitter-parser-TaX7E`)
Architecture: `docs/articles/research/TREE_SITTER_BREAKAGE.md`

https://claude.ai/code/session_01FYNpEDbTENdEkEmgCLYecS

---
_Generated by [Claude Code](https://claude.ai/code/session_01FYNpEDbTENdEkEmgCLYecS)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1770,6 +1770,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "perl-parser-comparison"
+version = "0.14.0"
+dependencies = [
+ "perl-ast",
+ "perl-parser-core",
+ "perl-parser-pest",
+ "perl-tdd-support",
+ "tree-sitter",
+ "tree-sitter-perl-c",
+]
+
+[[package]]
 name = "perl-parser-core"
 version = "0.14.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
     "crates/perl-parser-bench",
     "crates/perl-parser-core",
     "crates/perl-parser-pest",
+    "crates/perl-parser-comparison",
     "crates/perl-semantic-analyzer",
     "crates/perl-semantic-facts",
     "crates/perl-workspace",
@@ -171,6 +172,7 @@ allow = [
     # (perl-dead-code absorbed into perl-parser::dead_code — Wave 4-Completion)
     "perl-parser",
     # (perl-parser-bench is benchmark-only; publish = false)
+    # (perl-parser-comparison is test-only; publish = false)
     "perl-parser-pest",
     "perl-corpus",
 
--- a/crates/perl-parser-comparison/Cargo.toml
+++ b/crates/perl-parser-comparison/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "perl-parser-comparison"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+description = "Differential test suite measuring v1/v2/v3 parser gaps on constructs that historically defeated tree-sitter"
+publish = false
+
+[dependencies]
+perl-ast = { workspace = true }
+perl-parser-core = { workspace = true }
+perl-parser-pest = { workspace = true }
+tree-sitter-perl-c = { workspace = true }
+tree-sitter = { workspace = true }
+
+[dev-dependencies]
+perl-tdd-support = { workspace = true }
+
+[lib]
+doctest = false
+
+[lints]
+workspace = true
--- a/crates/perl-parser-comparison/src/harness.rs
+++ b/crates/perl-parser-comparison/src/harness.rs
@@ -1,4 +1,4 @@
-//! Parser harness — runs each of the three Perl parsers on a source string
+//! Parser harness - runs each of the three Perl parsers on a source string
 //! and produces a structured [`ParseResult`].
 //!
 //! Each parser is called through a thin wrapper that:
@@ -16,12 +16,13 @@ use crate::outcomes::Verdict;
 
 /// The output of running one parser on one input.
 #[derive(Debug)]
+#[non_exhaustive]
 pub struct ParseResult {
     /// Which parser produced this result.
     pub parser: ParserLabel,
     /// The source string that was parsed.
     pub source: String,
-    /// The verdict — outcome category.
+    /// The verdict - outcome category.
     pub verdict: Verdict,
     /// S-expression or description of the parse output (for diagnostics).
     /// Empty string if the parser crashed or returned no tree.
@@ -39,6 +40,7 @@ impl ParseResult {
 
 /// Identifies which parser produced a result.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum ParserLabel {
     /// v1: C tree-sitter grammar via FFI (`tree-sitter-perl-c`).
     V1TreeSitterC,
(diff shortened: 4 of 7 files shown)


EffortlessSteven:
## Builder Wrapup

**Summary**: Builder completed differential parser comparison test suite. 1 PR, 1 new crate, 50 tests, 0 failures.

### Ending state

- PR created: https://github.com/EffortlessMetrics/perl-lsp/pull/9170
- Tests: 50 added in `crates/perl-parser-comparison/tests/differential.rs`
- Files changed:
  - `Cargo.toml` (workspace member + publish comment)
  - `crates/perl-parser-comparison/Cargo.toml` (new)
  - `crates/perl-parser-comparison/src/lib.rs` (new)
  - `crates/perl-parser-comparison/src/outcomes.rs` (new — Verdict enum)
  - `crates/perl-parser-comparison/src/harness.rs` (new — parse_v1/v2/v3 wrappers)
  - `crates/perl-parser-comparison/tests/differential.rs` (new — 50 tests)
- Tests passing: `50 passed; 0 failed` + `cargo clippy -D warnings` clean + `cargo xtask fmt` clean
- Next step: ready for review

### Retrospective

**What was harder than expected**: The "expected outcome" for some cases needed to be discovered empirically rather than pre-specified. I had to run the tests, see what the parsers actually produced, then fix my expectations to match actual behavior. For example, v1 on `print <<A, <<B` does NOT produce ERROR nodes — it silently loses the second body while appearing to succeed. And v2 on `s{foo}/bar/g` doesn't lose "bar" — it puts it in the WRONG place as a separate division expression. These are subtly different failure modes than what was described in PR #9168.

**What would I do differently**: Write the harness first, then run it against control cases to understand actual sexp formats before writing structural assertions. I guessed wrong on several structural checks and had to iterate.

**What surprised me**: The Pest parser (v2) is remarkably permissive — it accepts JavaScript-style `function foo() {}` as valid Perl, accepts `my @@x = 5` with a wrong AST, and handles most inputs without error. This makes `Result::is_ok()` checks almost useless as coverage signal. The format body being empty in v2 is because the rule is declared as `@{}` (atomic) which collapses all sub-rules — a subtle PEG grammar property.

- Logical next: run `cargo test -p perl-parser-comparison -- --nocapture` to see the full disagreement table
- Related: PR #9168 (the investigation probe that preceded this)
- Gotchas: `${^MATCH}` in Rust format strings must be escaped as `dollar-caret-MATCH` in string literals — `${}` is a Rust format interpolation syntax and causes compile errors in assert messages
- Gotchas: v1 tree-sitter's `has_error()` is the right signal, but a `Correct` verdict from v1 doesn't mean correct — v1 accepts multi-heredoc without error nodes while silently losing body content
- Traps: The `WrongButPlausible` verdict is defined in the enum but not yet used as an `assert_verdict()` expected value — the v2 `map { a => $_ }` test uses structural assertions on the `sexp` instead. If a future reviewer asks "why is WrongButPlausible defined but not used?" — it's intentional; it's available for future test hardening.
- Confidence for next: high

---

PR #9176, opened 2026-05-17
feat(semantic-tokens): enable source-backed token slice (#8197)

EffortlessSteven:
Problem: Semantic-token parity proof had landed, but the compiler-backed token class was still only receipt metadata.
Fix: Promote the exact source-backed subroutine-declaration token class to a provider-decision live slice only when it matches existing live function-token output, preserving parser/HIR token output and fallback/shadow paths.
Verification: `cargo test -p perl-lsp-rs provider_decision_live_trace_tests --profile agent --locked -- --nocapture --test-threads=1` passes; `cargo test -p perl-lsp-rs semantic_tokens_runtime_quality_receipt --profile agent --locked -- --nocapture --test-threads=1` passes; `cargo test -p perl-lsp-rs --lib execute_command::tests --profile agent --locked -- --nocapture` passes; `cargo xtask fmt` passes; `git diff --check` passes.


EffortlessSteven:
Superseded by #9175, which landed the source-backed semantic-token live slice with a narrower provider-specific trace, no semantic-token output change, and green CI. This PR now conflicts with master and duplicates that lane objective, so closing to keep the queue clean.

PR #9207, opened 2026-05-17
docs(status): review package-local rename support (#8197)

EffortlessSteven:
Problem: Package-local rename support had landed as a narrow guarded pilot, but the UX dashboard still described rename as lexical-only.
Fix: Align the UX dashboard with current support-tier/provider evidence and move the near-term queue to the next proof items without promoting broader package/compiler-backed rename.
Verification: `cargo run -p xtask --profile agent --locked -- check-support-claims`; `cargo run -p xtask --profile agent --locked -- check-provider-confidence-matrix`; `cargo run -p xtask --profile agent --locked -- fmt --check`; `git diff --check`.


EffortlessSteven:
Superseded by #9206, which landed the package-local rename support review and queue update against current master.

PR #9231, opened 2026-05-17
feat(perl-ast): add deepest node lookup (#0000)

EffortlessSteven:
### Motivation
- Consumers need a convenient way to resolve the most specific AST node that covers a given byte offset for editor features like go-to, hover, and diagnostics.

### Description
- Add `Node::deepest_node_at_offset(&self, offset: usize) -> Option<&Node>` to return the most-specific descendant whose span contains `offset` (returns `None` when outside the node span) in `crates/perl-ast/src/ast.rs`.
- Implement depth-first descent using the existing `for_each_child` traversal to prefer deeper matches.
- Add behavior tests in `crates/perl-ast/tests/ast_behavior_spec_tests.rs` covering nested nodes, operator-gap offsets that should map to the containing expression, and out-of-span offsets.
- Keep public API and formatting consistent with existing crate conventions and doc examples.

### Testing
- Ran unit tests with `MIN_FREE_GB=20 ./scripts/cargo-safe test -p perl-ast --profile agent --locked`, and all `perl-ast` tests passed.
- Ran `MIN_FREE_GB=20 ./scripts/cargo-safe check --all-targets -p perl-ast --profile agent --locked`, and the crate checks passed.
- Ran `./scripts/cargo-safe xtask fmt` and `MIN_FREE_GB=20 ./scripts/cargo-safe clippy -p perl-ast --profile agent --locked -- -D warnings -A missing_docs`, and formatting and lints passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a08ca9e40648333a7e1f0db249bea82)

### DIFF
--- a/crates/perl-ast/src/ast.rs
+++ b/crates/perl-ast/src/ast.rs
@@ -1413,6 +1413,41 @@ impl Node {
         self.location.end.saturating_sub(self.location.start)
     }
 
+    /// Find the deepest descendant whose source span contains `offset`.
+    ///
+    /// Returns `None` when `offset` is outside this node's span. If multiple
+    /// nested nodes contain the offset, the most-specific descendant is returned;
+    /// otherwise this node is returned.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use perl_ast::{Node, NodeKind, SourceLocation};
+    ///
+    /// let outer = SourceLocation { start: 0, end: 5 };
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Closing as duplicate/superseded by #9230. This adds the same deepest-node offset capability under another public method name, so keeping one API avoids duplicate surface area.

Issue #8378, opened 2026-05-09
feat(workspace-symbols): add real-workspace compiler fact quality receipt


EffortlessSteven:
## Problem

Workspace-symbol compiler facts are shadowed, but the live provider still needs real-workspace quality proof before any compiler-backed workspace-symbol cutover.

Umbrella: #8197
Prior proof: #8353 / #8357

## Goal

Add a real-workspace workspace-symbol quality receipt that compares legacy workspace-symbol results with compiler-fact candidates without changing live provider behavior.

## Scope

- Imported/exported compiler fact candidates.
- Framework-generated candidates.
- Dynamic-boundary blockers.
- Stale compiler fact blockers.
- Query latency and candidate-count/noise comparison.
- ProviderFactTrace rows with source, provenance, confidence, freshness, fallback state, and dynamic-boundary state.

## Acceptance

- No live workspace-symbol behavior change.
- Receipt records legacy candidates, compiler candidates, rank/noise delta, stale/dynamic blockers, and generated-member labels.
- Real-workspace fixture includes fresh and stale compiler fact cases.
- Generated and dynamic-boundary answers are labeled honestly, not promoted silently.
- Existing semantic-shadow-compare and semantic-scorecard outputs remain current.

## Verification

```bash
cargo test -p perl-lsp-rs-core --profile agent --locked workspace_symbol -- --nocapture
cargo xtask semantic-shadow-compare --check
cargo xtask semantic-scorecard --check
cargo xtask metrics parser-accuracy --check
cargo xtask update-status --only parser --check
cargo xtask metrics ratchet-check parser_accuracy
cargo xtask fmt --check
git diff --check
```

## Non-goals

- No live workspace-symbol cutover.
- No completion, hover, navigation, rename, safe-delete, semantic-token, or document-symbol changes.
- No external Perl execution in provider code.

PR #9247, opened 2026-05-17
fix(workspace-symbols): trim symbol queries (#8378)

EffortlessSteven:
Problem: Workspace symbol queries with accidental outer whitespace could miss ready-index and generated-symbol matches because only the fallback query helper trimmed.
Fix: Trim workspace-symbol queries at the provider boundary and in workspace-index symbol searches, with regression coverage for fallback matching, ranking, and generated-symbol lookup.
Verification: `./scripts/cargo-safe test -p perl-lsp-rs-core symbol_query --profile agent --locked` passes; `./scripts/cargo-safe test -p perl-workspace search_symbols_returns_labeled_generated_framework_members --profile agent --locked -- --nocapture` passes; `./scripts/cargo-safe test -p perl-lsp-rs workspace_symbols_runtime_quality_receipt_echoes_query_field --profile agent --locked -- --nocapture` passes; `./scripts/cargo-safe xtask fmt --check` passes; `git diff --check` passes.


EffortlessSteven:
Review disposition: not ready to merge as-is. The trim is applied only inside the fallback symbol-query helper, not the workspace-feature ready-index path that handles live workspace/symbol requests. With the workspace feature enabled, \handle_workspace_symbols\ still passes the raw query into \WorkspaceIndex::search_source_symbols\ and \search_generated_workspace_symbols\ before falling back to open-document matching. That means a query like \  foo  \ can still miss ready-index source/generated hits, while whitespace-only queries can become broad fallback matches through \matches_query\ without the provider-decision trace treating them as empty-query behavior. Please normalize the query at the workspace/symbol request boundary, or make the ready-index and fallback paths share the same explicit normalization so matching, generated-symbol gating, and trace fields agree. Also retitle from \(#0000)\ before merge.

EffortlessSteven:
Updated the branch to trim workspace-symbol queries at the provider boundary and in workspace-index source/generated searches, so ready-index and generated-symbol paths now share the fallback helper's whitespace behavior. Also rebased on current master after #9251 merged.

EffortlessSteven:
Fixed the PR Smoke formatter failure from the Rust 1.95 Linux gate by applying the exact rustfmt contraction in workspace_index.rs. Revalidated locally with ./scripts/cargo-safe xtask fmt --check, git diff --check, and the focused perl-workspace generated-symbol test before force-pushing the amended single commit.

EffortlessSteven:
Restored the clean branch state after the intermediate chore(workspace-symbols): drop unrelated branch drift update. That update removed the provider/index normalization needed for ready-index/generated-symbol behavior and also introduced doc reversions against the now-merged #9251 base. Current diff is back to the intended three code files only.

EffortlessSteven:
Added a follow-up commit on top of the remote cleanup, instead of force-replacing it. The PR diff now remains status-doc clean while restoring query trimming for ready-index and generated workspace-symbol lookups, which is the path the core helper alone does not cover.

PR #9313, opened 2026-05-17
fix(devex): fall back for upstream log (#8197)

EffortlessSteven:
Problem: just upstream-log failed hard in local-only/minimal clones when no base ref could be auto-detected.
Fix: Fall back to HEAD with a warning and explicit ref hint instead of exiting.
Verification: just --list passes; shell logic selects origin/master in this checkout and falls back to HEAD in a detached no-ref fixture; invalid count still fails; git diff --check origin/master passes. Local just upstream-log execution is blocked by the existing Windows missing-cygpath bash-shebang issue.

### DIFF
--- a/justfile
+++ b/justfile
@@ -792,6 +792,7 @@ devex-targeted base='' mode='all':
 
 # Show recent upstream commits using an auto-detected base ref.
 # Helpful in detached or minimal-clone environments where origin/master may not exist.
+# Falls back to HEAD so local-only clones still satisfy the onboarding preflight.
 upstream-log count='20' base='':
     #!/usr/bin/env bash
     set -euo pipefail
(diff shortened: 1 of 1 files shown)


EffortlessSteven:

overlap checked:
- Single justfile DevEx fallback; no overlap with the already-merged cargo-safe/storage timeout work.

current master already contains:
- Existing upstream-log auto-detection for origin/main, origin/master, local main, and local master.
- This PR adds only the local-only HEAD fallback path.

retained branch:
- #9313, rebased onto current master and force-with-lease pushed.

proof run:
- just --list passed.
- Exact upstream-log recipe logic was validated under bash for explicit origin/master, auto-detected origin/master, invalid count rejection, and detached local-only HEAD fallback.
- git diff --check origin/master...HEAD passed.
- ./scripts/storage-doctor passed with no repo-local target dirs.
- Hosted checks passed, including CI Gate (Merge-Blocking), Windows Required, validate-title, Workflow Trigger Lint, methodology-gate, PR Plan, and Merge Ready Reconciler.
- Direct just upstream-log execution is still blocked locally by the known Windows missing-cygpath shebang issue before recipe logic runs; that is pre-existing and not widened by this change.

merge recommendation:
- Squash merge. This is a narrow DevEx fix that fails more usefully in local-only/minimal clones and does not change product behavior.

follow-up:
- Keep Windows just shebang/cygpath behavior as separate DevEx debt; do not block this fallback fix on it.

PR #9283, opened 2026-05-17
fix(import-optimizer): handle CRLF edit offsets (#9283)

EffortlessSteven:
### Motivation

- Avoid incorrect byte ranges when generating import-block edits for files using CRLF line endings that could leave replacement boundaries inside the import block.

### Description

- Replace the previous `line_offset` implementation with a newline-byte-position-driven implementation that scans `content.char_indices()` and returns the byte offset of the Nth newline to correctly handle both LF and CRLF content, in `crates/perl-refactoring/src/refactor/import_optimizer.rs`.
- Add a regression test `import_optimizer_edits_cover_crlf_import_block` that verifies `generate_edits` covers the full CRLF import block and does not leave stray `\r` characters, in `crates/perl-refactoring/tests/edge_case_coverage.rs`.

### Testing

- Ran the crate test matrix with `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo test -p perl-refactoring --profile agent --locked` and the updated perl-refactoring test suite passed (all tests in the crate passed); the added CRLF regression test also passed.
- Verified formatting and static checks with `cargo fmt --package perl-refactoring --check`, `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo check --all-targets -p perl-refactoring --profile agent --locked`, and `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo clippy -p perl-refactoring --profile agent --locked -- -D warnings -A missing_docs`, all of which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a07d1eddd948333a121a65367e4e242)

PR #9320, opened 2026-05-17
fix(parser): preserve CRLF import edit ranges (#9320)

EffortlessSteven:
### Motivation

- Import-edit byte ranges were computed assuming a single-byte newline which produced incorrect (short) ranges for files using CRLF line endings, causing import replacement edits to miss the terminating `\r\n`.

### Description

- Update `line_offset` in `crates/perl-parser/src/refactor/import_optimizer.rs` to iterate with `content.split_inclusive('\n')` and sum `segment.len()` so newline bytes (including `\r\n`) are counted correctly. 
- Add a regression test `test_generate_edits_preserves_crlf_import_block_range` that uses `concat!("use warnings;\r\n", ...)` to verify `generate_edits` produces the correct range and normalized replacement. 
- Change is limited to `line_offset` logic and the new test in the import optimizer test module.

### Testing

- Ran formatter with `./scripts/cargo-safe xtask fmt` which passed. 
- Executed the targeted unit test with `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo test -p perl-parser test_generate_edits_preserves_crlf_import_block_range --profile agent --locked` which passed. 
- Ran the full `perl-parser` library tests with `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo test -p perl-parser --lib --profile agent --locked` which passed (141 tests). 
- Verified `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo check --all-targets -p perl-parser --profile agent --locked` and `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo clippy -p perl-parser --profile agent --locked -- -D warnings -A missing_docs` which both passed; an initial test attempt was blocked by a storage quota guard but was re-run with `CARGO_TARGET_DIR` and succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a08ca24a1cc8333936e8932d1561b61)

### DIFF
--- a/crates/perl-parser/src/refactor/import_optimizer.rs
+++ b/crates/perl-parser/src/refactor/import_optimizer.rs
@@ -680,12 +680,13 @@ impl ImportOptimizer {
         if line <= 1 {
             return 0;
         }
+
         let mut offset = 0;
-        for (idx, l) in content.lines().enumerate() {
+        for (idx, segment) in content.split_inclusive('\n').enumerate() {
             if idx + 1 >= line {
                 break;
             }
-            offset += l.len() + 1; // include newline
+            offset += segment.len();
         }
         offset
     }
@@ -1054,6 +1055,21 @@ my $result = JSON::encode_json({test => 1});
         // Should only detect the actual module usage, not the ones in strings/regex
         assert_eq!(analysis.missing_imports.len(), 1);
         assert_eq!(analysis.missing_imports[0].module, "JSON");
+        Ok(())
+    }
+
+    #[test]
+    fn test_generate_edits_preserves_crlf_import_block_range()
+    -> Result<(), Box<dyn std::error::Error>> {
+        let optimizer = ImportOptimizer::new();
+        let content = concat!("use warnings;\r\n", "use strict;\r\n", "print qq(done);\r\n");
+        let expected_range_end = concat!("use warnings;\r\n", "use strict;\r\n").len();
+        let analysis = optimizer.analyze_content(content)?;
+
+        let edits = optimizer.generate_edits(content, &analysis);
+        assert_eq!(edits.len(), 1);
+        assert_eq!(edits[0].range, (0, expected_range_end));
+        assert_eq!(edits[0].new_text, "use strict;\nuse warnings;\n");
         Ok(())
     }
 


EffortlessSteven:
Classification: merge-candidate; stack head for canonical perl-parser CRLF import edit range fix.

What changed: perl-parser ImportOptimizer line_offset now sums split_inclusive newline segments, so CRLF newline bytes are included in generated import-block edit ranges. The branch also adds a focused CRLF regression for generate_edits.

Overlap: #9283 fixes the same CRLF offset class in the older perl-refactoring crate copy. These should be treated as an explicit two-item stack, not duplicate merges into the same crate. Merge #9320 first because perl-refactoring is documented as absorbed into perl-parser::refactor; then evaluate #9283 for the remaining workspace crate surface.

Verification run after rebase:
- git diff --check
- ./scripts/storage-doctor
- MIN_FREE_GB=20 MAX_USED_PCT=95 ./scripts/cargo-safe test -p perl-parser test_generate_edits_preserves_crlf_import_block_range --profile agent --locked -- --nocapture
- MIN_FREE_GB=20 MAX_USED_PCT=95 ./scripts/cargo-safe xtask fmt
- MIN_FREE_GB=20 MAX_USED_PCT=95 ./scripts/cargo-safe check --all-targets -p perl-parser --profile agent --locked
- MIN_FREE_GB=20 MAX_USED_PCT=95 ./scripts/cargo-safe test -p perl-parser --lib --profile agent --locked
- MIN_FREE_GB=20 MAX_USED_PCT=95 ./scripts/cargo-safe clippy -p perl-parser --profile agent --locked -- -D warnings -A missing_docs

Full crate test note: MIN_FREE_GB=20 MAX_USED_PCT=95 ./scripts/cargo-safe test -p perl-parser --profile agent --locked still fails in issue_146_architectural_integrity_tests because that test invokes package ID perl-lsp while the actual package is perllsp. I reproduced the exact failing test on clean master with the same error, so this is pre-existing and unrelated to the import optimizer diff.

Recommendation: squash merge after required GitHub checks pass, then handle #9283 as the follow-up for the perl-refactoring copy.

PR #9323, opened 2026-05-17
fix(hover): improve missing module guidance (#8197)

EffortlessSteven:
### Motivation

- Missing-module hover cards only said the module was not found and listed raw include paths, which is not actionable for users. 
- Users need clear explanation of which roots were searched (including `PERL5LIB`) and whether system `@INC` was considered. 
- Provide direct next steps (install or update `include_paths`) to reduce friction when a module is unresolved.

### Description

- Update hover fallback in `LspServer::build_module_hover` to compute effective include paths via `WorkspaceConfig::effective_include_paths` (merges `PERL5LIB` according to config) and surface a richer markdown card showing searched roots, system `@INC` status, and suggested next steps. 
- Add helper `fn format_missing_module_search_paths(include_paths: &[String]) -> String` to render human-friendly bullet lists (and an explicit message when no include paths are configured). 
- Add two focused unit tests in `crates/perl-lsp-rs/src/runtime/language/hover/hover_tests.rs`: `missing_module_hover_gives_actionable_next_steps` and `missing_module_search_paths_reports_empty_configuration` to exercise the new text and formatting. 

### Testing

- Ran `./scripts/cargo-safe xtask fmt` which succeeded. 
- Ran `MIN_FREE_GB=20 ./scripts/cargo-safe test -p perl-lsp-rs --profile agent --locked --lib missing_module -- --nocapture` and the two new hover tests passed. 
- Ran `MIN_FREE_GB=20 ./scripts/cargo-safe check --all-targets -p perl-lsp-rs --profile agent --locked` which completed successfully. 
- Ran `./scripts/storage-doctor` to validate build storage state.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a08161a97e8833389dd76cf17585623)


EffortlessSteven:
Reviewed under the Real Perl Editor Trust WIP cap. I would park this for now rather than merge.

Blocking issues:
- The missing-module hover now recomputes `effective_include_paths` for the explanatory text after the position-aware resolver misses. In the `no lib` case, that can list paths that were inactive at the use site, so the card claims those roots were searched when they were intentionally cancelled.
- The unresolved-hover text contract changed from the UX fixture's accepted `Not found in workspace` shape to `Module not found in workspace or configured include paths`, which is why the UX regression gate is red.

Before this should re-enter the merge lane, it needs the explanation derived from the same effective include context used by resolution, or softer wording that does not claim inactive paths were searched. It also needs UX coverage for unresolved hover text, `PERL5LIB`, `useSystemInc`, and `no lib` cancellation.

PR #9297, opened 2026-05-17
fix(parser): accept keyword hash subscript keys (#2189)

EffortlessSteven:
### Motivation
- Corpus files used reserved words as unquoted hash keys (e.g. `$bits{tie}`, `$self->{defer}`) and the parser treated those tokens as keywords, producing ERROR nodes and cascading parse noise.
- The intent is to treat keyword-like tokens followed by a hash-subscript delimiter as bare identifier/hash-key context while preserving normal keyword/operator behavior elsewhere.

### Description
- Added `is_keyword_hash_key_boundary()` helper to detect when a keyword-like token is followed by a subscript delimiter (`}`), fat-arrow (`=>`), or comma so it should be treated as a bare key instead of a keyword. (`crates/perl-parser-core/src/engine/parser/statements.rs`).
- Updated primary-expression dispatch to use the new helper for keyword tokens such as `eval`, `do`, `try`, and `defer`, and to short-circuit `tie`/`untie` handling when the token is actually a hash-key. (`crates/perl-parser-core/src/engine/parser/expressions/primary.rs`).
- Added regression coverage exercising real corpus patterns (`$bits{tie}`, `$self->{defer}`, hash-slice keyword keys) in `fix_unexpected_rbrace_hash_key.rs` and adjusted one test that exposed the prior gap. (`crates/perl-parser-core/tests/fix_unexpected_rbrace_hash_key.rs`).

### Testing
- Ran `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo test -p perl-parser-core --test fix_unexpected_rbrace_hash_key --profile agent --locked`; tests passed (all added/affected tests green).
- Ran `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo check -p perl-parser-core --all-targets --profile agent --locked`; type-check passed.
- Ran `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo clippy -p perl-parser-core --profile agent --locked -- -D warnings -A missing_docs`; lints passed under the enforced flags.
- Ran the local corpus sweep via `CARGO_TARGET_DIR=/tmp/perl-lsp-target cargo run -p xtask --profile agent --locked -- parser-corpus-sweep --output /tmp/parser-after.json`, which completed successfully and showed improvement on the local system corpus (dirty files decreased from 58 → 54 and total ERROR nodes from 132 → 64).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a084327f3248333873bdb9c39587e5e)

PR #9428, opened 2026-05-18
fix(parser): parse builtin hash keys before rbrace (#9170)

EffortlessSteven:
Problem: local, 	ie, and untie can appear as source-backed bare hash subscript keys in real Perl modules, but keyword/builtin parsing can still treat them as expressions before }.
Fix: Preserve the current comma/slice-aware hash-key parser and add the remaining local keyword case plus B::Op_private/Dpkg regression fixtures.
Verification: ./scripts/cargo-safe test -p perl-parser-core --test fix_unexpected_rbrace_expr_2404 --profile agent --locked -- --nocapture; ./scripts/cargo-safe check --all-targets -p perl-parser-core --profile agent --locked; git diff --check origin/master...HEAD.
Claim boundary: focused parser capability fixture/fix only; no generated status or corpus bucket-count claim without a fresh corpus receipt.


EffortlessSteven:
Superseded by #9429 after #9297 landed the broader keyword hash-key change on master. The remaining unique fix is the local-key Dpkg regression.

PR #9534, opened 2026-05-20
docs(spec): define semantic token class promotion contract (#8197)

EffortlessSteven:
Problem: Scoped compiler-token traces exist, but the class-by-class promotion rules were not captured as a durable spec and policy registry. Fix: Add PLSP-SPEC-0019 plus policy/semantic-token-classes.toml, and link them from the trust v1 boundary, spec catalog, and dashboard source stack. Verification: rtk python TOML parse for policy/semantic-token-classes.toml; rtk cargo xtask check-provider-confidence-matrix; rtk cargo xtask check-support-claims; rtk cargo xtask check-provider-promotion-ledger; rtk cargo xtask ci-hygiene check-doc-paths docs/specs; rtk cargo xtask ci-hygiene check-doc-paths docs/project/status; rtk git diff --check.

### DIFF
--- a/docs/project/status/real_perl_editor_trust_v1.md
+++ b/docs/project/status/real_perl_editor_trust_v1.md
@@ -22,6 +22,7 @@ of current evidence.
 | Provider decision receipt schema and explanation payload contract | [PLSP-SPEC-0016](../../specs/PLSP-SPEC-0016-provider-decision-receipt-v1.md), [provider_decision.v1.schema.json](../../../schemas/provider_decision.v1.schema.json) |
 | Shared fact provenance and source-backing semantics | [PLSP-SPEC-0017](../../specs/PLSP-SPEC-0017-fact-provenance-and-source-backing.md) |
 | Shared edit authorization states for rename and safe delete | [PLSP-SPEC-0018](../../specs/PLSP-SPEC-0018-edit-authorization-contract.md) |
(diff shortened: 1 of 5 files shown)


EffortlessSteven:
Review notes: I think this needs a small docs/policy fix before merge. PLSP-SPEC-0019 says each promoted/traced class must define receipt sources, and the future validator section says every policy row should have receipt sources, but policy/semantic-token-classes.toml only has one global receipt_sources list and the individual [[class]] rows do not name their receipts. That weakens the class-by-class proof boundary this spec is trying to lock. Also, .perl-lsp/goals/active.toml still lists specs only through PLSP-SPEC-0014, so the new semantic-token class promotion contract is not discoverable from the active goal manifest even though the lane objective depends on Codex being able to route from the manifest. Existing validators are green, so this is a contract/discoverability gap rather than a compile failure.

EffortlessSteven:
Follow-up: I prepared and validated the minimal correction locally in a detached worktree. The patch adds receipt_sources to each [[class]] row in policy/semantic-token-classes.toml and adds docs/specs/PLSP-SPEC-0019-semantic-token-class-promotion-contract.md to .perl-lsp/goals/active.toml. Local checks passed: TOML parse with all 8 classes carrying receipt_sources, check-provider-confidence-matrix, check-support-claims, check-provider-promotion-ledger, docs path hygiene for docs/specs and docs/project/status, git diff --check, and scripts/storage-doctor. I did not push because this is another agent's PR branch.

PR #9556, opened 2026-05-22
fix(type-definition): resolve qualified class names and chained method calls

EffortlessSteven:
## Problem

The `TypeDefinitionProvider`'s go-to-type-definition feature had two bugs that prevented it from correctly resolving types in common Perl OOP patterns:

### Bug 1 — Qualified class names always had their last component stripped

The `extract_type_name` Identifier arm used a blanket rule to strip the last `::` component, treating every qualified identifier as a `Package::method` call. This caused `Package::Name->new()` with the cursor on the receiver to return `"Package"` instead of `"Package::Name"`, making goto-type-definition navigate to the wrong package.

**Root cause:** The code could not distinguish `Foo::Bar` (a class name — last component uppercase) from `Foo::bar` (a qualified function call — last component lowercase). The fix: inspect the initial character of the last `::`-component to decide which interpretation applies.

### Bug 2 — `infer_object_type` was a stub returning `None` for all cases

When the cursor lands on a **method name** (e.g., on `"new"` in `MyClass->new()`), the most specific AST node at that position is the `MethodCall` itself (the method name is a `String` field, not a child `Node`). The provider calls `infer_object_type(object)` to extract the package from the method call's receiver. Previously this always returned `None`, making goto-type-definition silently fail when the cursor was on the method name rather than the receiver.

The stub also had no path for **chained method calls** (`Foo::Bar->new->another()`), where the outer call's object is itself a `MethodCall` wrapping the package identifier.

## Changes

**File:** `crates/perl-lsp-rs-core/src/providers/navigation/type_definition.rs`

### `extract_type_name` — Identifier arm (lines ~177–195)

```rust
// Before: always strips last :: component
if name.contains("::") {
    let parts = name.split("::").collect::<Vec<_>>();
    return Some(parts[..parts.len()-1].join("::"));  // "Foo::Bar" → "Foo" ❌
}

// After: use case of last component to distinguish class vs qualified call
if name.contains("::") {
    let parts = name.split("::").collect::<Vec<_>>();
    let last = parts[parts.len()-1];
    if last.chars().next().is_some_and(|c| c.is_uppercase()) {
        return Some(name.clone());      // "Foo::Bar" → "Foo::Bar" ✓
    }
    return Some(parts[..parts.len()-1].join("::"));  // "Foo::bar" → "Foo" ✓
}
```

### `infer_object_type` — from stub to functional (lines ~414–465)

```rust
// Before: stub, always None
fn infer_object_type(&self, object: &Node) -> Option<String> {
    match &object.kind {
        NodeKind::Variable { name, .. } => None,  // $self/$this
        NodeKind::FunctionCall { name, .. } if name == "new" => None,
        _ => None,
    }
}

// After: handles 3 cases
fn infer_object_type(&self, object: &Node) -> Option<String> {
    match &object.kind {
        // Foo->method() — object is Identifier("Foo") or Identifier("Foo::Bar")
        NodeKind::Identifier { name } => {
            if name.contains("::") || name.starts_with(uppercase) { Some(name.clone()) }
            else { None }
        }
        // Foo->new->bar() — object is MethodCall; recurse to peel back chain
        NodeKind::MethodCall { object: inner, .. } => self.infer_object_type(inner),
        // (Foo->new)->bar() — Binary -> node in nested context
        NodeKind::Binary { op, left, .. } if op == "->" => { /* extract left.name */ }
        // $self, $obj, etc. — require data-flow analysis, return None
        NodeKind::Variable { .. } | NodeKind::FunctionCall { .. } => None,
        _ => None,
    }
}
```

## Test coverage

Replaced the empty stub test `test_extract_type_from_constructor` (no body, no assertions) with **7 targeted tests**:

| Test | Pattern covered | Assertion |
|------|----------------|-----------|
| `test_extract_type_from_constructor_cursor_on_method` | `my $obj = Package::Name->new()` cursor on "new" | Returns "Package::Name" |
| `test_extract_type_from_constructor_cursor_on_package` | Same code, cursor on "Package::Name" identifier | Returns "Package::Name" (was "Package") |
| `test_extract_type_chained_method_call` | `Package::Name->new()->method()` cursor on "method" | Returns "Package::Name" via recursion |
| `test_infer_object_type_identifier` | `Foo->bar()` cursor on "bar" | Returns "Foo" |
| `test_extract_type_simple_var_returns_none` | `$obj->method()` cursor on "method" | Returns None (correct — no data flow) |
| `test_full_type_definition_chained_constructor` | Full goto-type-def for `MyFactory->create()->do_work()` | Finds `package MyFactory` |

## Verification

```
cargo test -p perl-lsp-rs-core --features lsp-compat -- type_definition
# 13 passed; 0 failed

cargo test -p perl-lsp-rs-core --features lsp-compat --lib
# 1795 passed; 0 failed

cargo clippy -p perl-lsp-rs-core --features lsp-compat --lib -- -D warnings
# clean

cargo xtask fmt
# no changes needed
```

## Scope

Single file changed. No behavior change to any other provider. The `$self`/`$this` variable case correctly returns `None` — type inference for variables requires data-flow analysis across assignment sites, which is a separate feature tracked in the OOP issues.

https://claude.ai/code/session_01HUHqL6p6dLZp4wJkPpuv3x

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUHqL6p6dLZp4wJkPpuv3x)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/navigation/type_definition.rs
+++ b/crates/perl-lsp-rs-core/src/providers/navigation/type_definition.rs
@@ -176,10 +176,16 @@ impl TypeDefinitionProvider {
             // Package identifier or Package::method
             NodeKind::Identifier { name } => {
                 if name.contains("::") {
-                    // Qualified name like Package::method
                     let parts: Vec<&str> = name.split("::").collect();
                     if parts.len() >= 2 {
-                        // Get the package name (everything except the last part)
+                        let last = parts[parts.len() - 1];
+                        if last.chars().next().is_some_and(|c| c.is_uppercase()) {
+                            // Last component starts uppercase → the whole name is a
+                            // class/package (e.g., `Foo::Bar`). Return the full name.
+                            return Some(name.clone());
+                        }
+                        // Last component starts lowercase → qualified method/function
+                        // (e.g., `Foo::bar`). Return only the package part.
                         return Some(parts[..parts.len() - 1].join("::"));
                     }
                 }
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Development has moved to `EffortlessMetrics/perl-lsp-swarm`.

A safe subset of this publishing-repo PR is now open in swarm:

- EffortlessMetrics/perl-lsp-swarm#248

That swarm PR ports the direct class receiver and qualified package-name fixes, but intentionally does not port the chained method-call recursion. `Class->factory()->method` is not source-backed proof that `factory()` returns `Class`, so treating that chain as exact would overclaim type-definition certainty.

`EffortlessMetrics/perl-lsp` is reserved for release/publish/signing and deliberate release syncs. This draft source PR is closed to avoid split-brain development.

EffortlessSteven:
classification: draft-hold / partially ported

overlap checked: Compared this draft against current perl-lsp-swarm/main. The qualified class-name and direct package receiver pieces were useful; the arbitrary chained method-call inference was too broad for the current evidence boundary.

current master already contains: Not fully. The safe subset is now ported to swarm PR EffortlessMetrics/perl-lsp-swarm#251.

retained branch: perl-lsp-swarm#251

proof run on retained swarm branch:
- git diff --check
- cargo fmt -p perl-lsp-rs-core -- --check
- MIN_FREE_GB=20 MAX_USED_PCT=96 CARGO_LOCK_WAIT=600 ./scripts/cargo-safe test -p perl-lsp-rs-core --features lsp-compat --lib --profile agent --locked type_definition -- --nocapture
- MIN_FREE_GB=20 MAX_USED_PCT=96 CARGO_LOCK_WAIT=600 ./scripts/cargo-safe check --all-targets -p perl-lsp-rs-core --features lsp-compat --profile agent --locked
- MIN_FREE_GB=20 MAX_USED_PCT=96 CARGO_LOCK_WAIT=600 ./scripts/cargo-safe clippy -p perl-lsp-rs-core --features lsp-compat --profile agent --locked -- -D warnings -A missing_docs
- ./scripts/storage-doctor

close/merge rationale: Do not merge this source draft as-is. Its infer_object_type recurses through arbitrary MethodCall receivers, so a call such as MyFactory->create()->do_work() is treated as MyFactory even though create() may return any object. That overclaims exactly where the receiver-facts lane says we need explicit return/fact evidence.

follow-up: Chain-aware type-definition should wait for receiver/return facts, then opt in only for proven exact receiver kinds and dynamic-boundary-free cases.

EffortlessSteven:
Update: while the conservative swarm port was open, perl-lsp-swarm#248 landed the retained direct receiver fix on main. I closed perl-lsp-swarm#251 as superseded-by:#248 to avoid duplicate queue churn.

Current disposition for this source draft remains: draft-hold / superseded in part. The safe direct receiver behavior is on swarm main via #248; the arbitrary chained method return inference in this draft should not be merged without receiver/return facts and dynamic-boundary proof.

PR #9578, opened 2026-05-24
fix(parser-core): treat class/method/format as infix-RHS declaration starters (#7472)

EffortlessSteven:
## Summary

Fixes the error-recovery regression described in issue #7472: `class`, `method`, and `format` (Perl 5.38+ declaration keywords) were missing from `is_infix_rhs_absent` in `crates/perl-parser-core/src/engine/parser/helpers.rs`.

### The bug

When any of these keywords followed an infix operator, the parser did not recognize the incoming declaration as a "statement boundary" signal. Instead it attempted to parse the declaration as the expression RHS, consuming the class/method/format node and producing a malformed AST:

```perl
my $x = class Foo { }   # Parser consumed class Foo {} as RHS of =
my $y = format Foo =    # Same problem
.
```

The result was a single incorrect statement instead of two statements, and the `Class`/`Method`/`Format` AST node was lost.

### The fix

Two files changed, zero production-logic changes — only the recovery discriminant is extended.

**`src/engine/parser/helpers.rs`** — `is_infix_rhs_absent`:

Added `TokenKind::Class | TokenKind::Method | TokenKind::Format` to the existing match of declaration-starter keywords that signal a missing operand. Two early-return guards mirror the disambiguation logic already present in `parse_statement`:

- **`class`**: treated as a declaration starter only when the next token is `Identifier`, `DoubleColon`, or `Colon`. This preserves the common Moose-style `class->new()` bareword method-call as a valid expression RHS — the same guard used in `parse_statement`.
- **`method`**: treated as a declaration starter only when the next token is `Identifier`. This lets `$obj->method(...)` (tokenized as `Identifier`, not `Method`) pass through unaffected.
- **`format`**: always a declaration starter; no ambiguous bareword form.

**`src/engine/parser/error_recovery_tests.rs`** — three new regression tests:

| Test | Input | Expected |
|------|-------|----------|
| `test_recovery_missing_rhs_before_class_declaration_keyword` | `my $x = class Foo { }` | 2 stmts; stmt[1] is `NodeKind::Class` |
| `test_recovery_missing_rhs_before_method_declaration_keyword` | `my $x = method foo { }` | ≥ 2 stmts |
| `test_recovery_missing_rhs_before_format_declaration_keyword` | `my $x = format Foo =\n.\n` | ≥ 2 stmts; contains `NodeKind::Format` |

### Disambiguation correctness

[...]

Issue #7472, opened 2026-04-30
[resubmit] fix(parser-core): close declaration and recovery edge-case cluster — orig PR #5990 unrebaseable


EffortlessSteven:
Original PR #5990 could not be cleanly resubmitted (patch did not apply) due to fresh-root rebuild on 2026-04-24. Captures intent for future builder work.

Original PR: #5990
Original body:
---
### Motivation

- Assignment-RHS recovery sometimes consumed declaration-starting keywords (e.g. `class`) as expression RHS, producing wrong AST shape and losing the following declaration node.

### Description

- Treat declaration starters as strong followers in infix-RHS detection by adding `class`, `method`, and `format` to `is_infix_rhs_absent` in `crates/perl-parser-core/src/engine/parser/helpers.rs`.
- Add a regression test `test_recovery_missing_rhs_before_class_declaration_keyword` to `crates/perl-parser-core/src/engine/parser/error_recovery_tests.rs` that asserts `my $x = class Foo { ... }` recovers into two statements with a preserved `Class` node.
- Changes are limited to the two files above and a single focused test.

### Testing

- Ran the targeted tests: `cargo test -p perl-parser-core engine::parser::error_recovery_tests::` and the new test `test_recovery_missing_rhs_before_class_declaration_keyword` — both passed.
- Ran `cargo test -p perl-parser-core` and `cargo check --workspace --all-targets`; `perl-parser-core` tests passed and workspace check passed.
- Note: a separate failing test in `perl-parser` surfaced during broader runs but is unrelated to these changes and pre-existed in the workspace during verification.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eaffa80b24833385869b58ffddcab0)

EffortlessSteven:
Closing as [resubmit] wrapper duplicate. Original PR reference documented in issue body. (disposition: EffortlessMetrics/perl-lsp-swarm#1731)

PR #9580, opened 2026-05-25
fix(parser-core): recover phaser/given/defer missing RHS (#7472)

EffortlessSteven:
## Summary

Extends the error-recovery discriminant `is_infix_rhs_absent` in `crates/perl-parser-core/src/engine/parser/helpers.rs` to cover seven additional statement-starter keywords that can never legitimately serve as the right-hand side of an infix operator.

This is a follow-on to PR #9578 (which fixed `class`/`method`/`format` for issue #7472), applying the same pattern to a wider set of keywords that were still missing from the function.

---

## The bug

When any of these keywords followed an infix operator such as `=`, `is_infix_rhs_absent` returned `false`, so the parser did **not** treat the upcoming declaration as a statement boundary. Instead it attempted to consume the declaration as the expression RHS, producing a malformed AST:

```perl
my $x = BEGIN { require 'config.pl' }  # Parser consumed BEGIN block as RHS
my $y = given ($val) { when (1) { 'one' } }  # Same problem
my $z = defer { cleanup() }  # Same problem
```

In each case the result was a single incorrect statement, and the phaser/`given`/`defer` AST node was either lost or embedded in the wrong position in the tree.

---

## The fix

Two files changed, no production-logic changes beyond the recovery discriminant.

### `src/engine/parser/helpers.rs` — `is_infix_rhs_absent`

Three new early-return guards are added before the existing match statement (following the exact pattern established by the `class`/`method` guards in PR #9578), then the keywords are appended to the match arm that signals a missing operand:

#### Phaser blocks — `Begin / End / Check / Init / Unitcheck`

```rust
// Treated as declaration starters only when followed by `{`.
// CPAN code occasionally calls phasers as barewords: `CHECK()`, `INIT()`
// — those forms must be allowed as a valid expression RHS.
```

Guard: if NOT followed by `LeftBrace`, return `false` (allow as expression).
Match arm: the phaser keywords always signal a missing operand when `{` follows.

#### `given`

```rust
// Treated as a declaration starter only when followed by `(`.
// Without `(`, allow it as a potential bareword to avoid breaking
// user-defined `given()` subs.
```

Guard: if NOT followed by `LeftParen`, return `false`.
Match arm: `given` always signals a missing operand when `(` follows.

#### `defer` (Perl 5.36+)

```rust
// Treated as a declaration starter only when followed by `{`.
// Without `{`, it may appear as a hash key via fat-comma autoquote
// (`defer => 1`), which must not trigger missing-RHS recovery.
```

Guard: if NOT followed by `LeftBrace`, return `false`.
Match arm: `defer` signals a missing operand when `{` follows.

### `tests/fix_infix_rhs_phaser_given_recovery.rs` — 14 new regression tests

| Test | Input | Assertion |
|------|-------|-----------|
| `test_recovery_missing_rhs_before_begin_block` | `my $x = BEGIN { ... }` | ≥2 stmts, has diagnostics |
| `test_recovery_missing_rhs_before_end_block` | `my $x = END { ... }` | same |
| `test_recovery_missing_rhs_before_check_block` | `my $x = CHECK { ... }` | same |
| `test_recovery_missing_rhs_before_init_block` | `my $x = INIT { ... }` | same |
| `test_recovery_missing_rhs_before_unitcheck_block` | `my $x = UNITCHECK { ... }` | same |
| `test_standalone_begin_block_parses_cleanly` | `BEGIN { ... }` | no error nodes |
| `test_standalone_end_block_parses_cleanly` | `END { ... }` | no error nodes |
| `test_standalone_check_block_parses_cleanly` | `CHECK { ... }` | no error nodes |
| `test_phaser_as_statement_label_parses_cleanly` | `CHECK: for my $i (1..10) { ... }` | no error nodes |
| `test_recovery_missing_rhs_before_given_statement` | `my $x = given ($y) { ... }` | ≥2 stmts, has diagnostics |
| `test_recovery_missing_rhs_before_defer_block` | `my $x = defer { ... }` | ≥2 stmts, has diagnostics |
| `test_standalone_defer_block_parses_cleanly` | `defer { ... }` | no error nodes |
| `test_defer_as_hash_key_no_recovery` | `my %h = (defer => 1)` | 1 stmt, no diagnostics |
| `test_begin_as_hash_key_no_recovery` | `my %h = (BEGIN => 'init')` | 1 stmt, no diagnostics |

---

## Disambiguation correctness

The fix explicitly preserves the expected behaviour of:

- **`CHECK()` as a bareword call** (CPAN pattern) — guard passes it through as a valid RHS
- **`defer => value`** (hash key via fat-comma autoquote) — guard passes it through as valid RHS
- **`CHECK: for ...`** (phaser keyword as loop label) — `parse_statement_inner` dispatches to `parse_keyword_as_label()` first, completely unaffected by this change
- **`BEGIN { }` standalone** — parses cleanly as a top-level phaser block statement

---

## Test plan

- [x] `cargo test -p perl-parser-core --test fix_infix_rhs_phaser_given_recovery` — 14/14 pass
- [x] `cargo test -p perl-parser-core` — all existing tests pass (0 failures)
- [x] `cargo clippy -p perl-parser-core --tests` — no warnings
- [x] `cargo fmt --all` — applied, no diff
- [x] `cargo check --workspace` — clean

## Scope

- Only `crates/perl-parser-core/` touched
- No public API changes
- No `Cargo.toml` changes
- Two files: `helpers.rs` (55 inserted lines) + new test file (248 lines)

---
_Generated by [Claude Code](https://claude.ai/code/session_015uQnZCYhoNYzEevXWeYxnL)_

### DIFF
--- a/crates/perl-parser-core/src/engine/parser/helpers.rs
+++ b/crates/perl-parser-core/src/engine/parser/helpers.rs
@@ -741,6 +741,49 @@ impl<'a> Parser<'a> {
             return false;
         }
 
+        // Phaser blocks (BEGIN/END/CHECK/INIT/UNITCHECK) start compile-time
+        // declarations only when followed by `{`.  CPAN code occasionally uses
+        // these as bareword function calls (`CHECK()`, `INIT()`), which are valid
+        // expression RHS forms and must not trigger missing-operand recovery.
+        if matches!(
+            self.peek_kind(),
+            Some(
+                TokenKind::Begin
+                    | TokenKind::End
+                    | TokenKind::Check
+                    | TokenKind::Init
+                    | TokenKind::Unitcheck
+            )
+        ) && !matches!(
+            self.tokens.peek_second().map(|t| t.kind),
+            Ok(TokenKind::LeftBrace)
+        ) {
+            return false;
+        }
+
+        // `given` (experimental given/when) is always a compound statement
+        // when followed by `(expr)`.  Without `(`, treat it as a potential
+        // bareword identifier to avoid breaking user-defined `given()` subs.
+        if self.peek_kind() == Some(TokenKind::Given)
+            && !matches!(
+                self.tokens.peek_second().map(|t| t.kind),
+                Ok(TokenKind::LeftParen)
+            )
+        {
+            return false;
+        }
+
+        // `defer` (Perl 5.36+ experimental) is a block statement when followed
+        // by `{`.  Without `{`, it may appear as a hash key or bareword.
+        if self.peek_kind() == Some(TokenKind::Defer)
+            && !matches!(
+                self.tokens.peek_second().map(|t| t.kind),
+                Ok(TokenKind::LeftBrace)
+            )
+        {
+            return false;
+        }
+
         match self.peek_kind() {
             Some(kind) if kind.is_recovery_boundary() => true,
             // Statement-starter keywords cannot serve as an expression RHS.
@@ -767,6 +810,18 @@ impl<'a> Parser<'a> {
             | Some(TokenKind::Class)
             | Some(TokenKind::Method)
             | Some(TokenKind::Format)
+            // Compile-time phaser blocks: never a valid expression RHS in the
+            // block form; the early-return guard above passes through bareword
+            // call forms such as `CHECK()` or `INIT()`.
+            | Some(TokenKind::Begin)
+            | Some(TokenKind::End)
+            | Some(TokenKind::Check)
+            | Some(TokenKind::Init)
+            | Some(TokenKind::Unitcheck)
+            // `given` compound statement: not an expression when followed by `(`.
+            | Some(TokenKind::Given)
+            // `defer` Perl 5.36+ block: not an expression when followed by `{`.
+            | Some(TokenKind::Defer)
             | None => true,
             _ => false,
         }
--- a/crates/perl-parser-core/tests/fix_infix_rhs_phaser_given_recovery.rs
+++ b/crates/perl-parser-core/tests/fix_infix_rhs_phaser_given_recovery.rs
@@ -1,0 +1,147 @@
+mod cpan_test_helpers;
+use cpan_test_helpers::*;
+
+use perl_parser_core::{NodeKind, Parser};
+use perl_tdd_support::must;
+
+fn assert_missing_rhs_recovery_preserves_statement(code: &str, label: &str) {
+    let mut parser = Parser::new(code);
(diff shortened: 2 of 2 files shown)


EffortlessSteven:

overlap checked: #9580 overlapped #9578 in is_infix_rhs_absent. #9578 has already landed class/method/format guards; this branch was rebased on top and now keeps those guards while adding only phaser/given/defer handling and regression coverage.

current master already contains: partially. Master contains the #9578 class/method/format recovery fix, but not the BEGIN/END/CHECK/INIT/UNITCHECK, given, or defer recovery cases from this PR.

retained branch: #9580.

proof run:
- MIN_FREE_GB=20 MAX_USED_PCT=99 CARGO_LOCK_WAIT=1200 ./scripts/cargo-safe xtask fmt
- MIN_FREE_GB=20 MAX_USED_PCT=99 CARGO_LOCK_WAIT=1200 ./scripts/cargo-safe test -p perl-parser-core --test fix_infix_rhs_phaser_given_recovery --profile agent --locked -- --nocapture (14 passed)
- MIN_FREE_GB=20 MAX_USED_PCT=99 CARGO_LOCK_WAIT=1200 ./scripts/cargo-safe test -p perl-parser-core --profile agent --locked -- --nocapture (crate suite passed)
- MIN_FREE_GB=20 MAX_USED_PCT=99 CARGO_LOCK_WAIT=1200 ./scripts/cargo-safe check --all-targets -p perl-parser-core --profile agent --locked
- MIN_FREE_GB=20 MAX_USED_PCT=99 CARGO_LOCK_WAIT=1200 ./scripts/cargo-safe clippy -p perl-parser-core --profile agent --locked -- -D warnings -A missing_docs
- git diff --check origin/master...HEAD
- ./scripts/storage-doctor

review note: I resolved the #9578 conflict by preserving the existing class/method/format disambiguation and adding the phaser/given/defer guards after it. I also rewrote the new regression file to remove repeated unreachable!() branches and non-ASCII separator text.

validation gap: parser-core clippy with --tests is currently blocked by pre-existing test print output under the workspace print lint promoted by #9593 (first hit: crates/perl-parser-core/tests/fix_expected_colon_tests.rs). Production parser-core clippy and parser-core tests pass.

merge rationale: narrow parser recovery fix for real Perl statement starters, with guard tests ensuring bareword/hash-key forms such as BEGIN => and defer => do not become false missing-RHS recoveries.

follow-up: handle the remaining print-lint test exceptions/no-panic debt separately so --tests clippy can become a reliable broad gate again.

Issue #9612, opened 2026-05-26
feat(code-actions): add quick-fix for PL410 undefined loop-control label


EffortlessSteven:
## Problem

`next LABEL`, `last LABEL`, and `redo LABEL` statements that reference an undefined label receive a PL410 diagnostic, but clicking the lightbulb in an editor returns **no code actions**. The diagnostic route table in `diagnostic_routes.rs` had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so all PL410 diagnostics silently fell through to `_ => {}`.

This is a runtime-fatal condition in Perl — if the named label doesn't exist, Perl dies at runtime with `Label not found for "next LABEL"`. The fix is obvious and mechanical: drop the label so the statement targets the innermost enclosing loop.

## Scope

- Add `fix_loop_control_undefined_label` quick-fix handler in `quick_fixes.rs`
- Wire it into `diagnostic_routes.rs` under the `PL410` / `LoopControlUndefinedLabel` code
- Add 6 BDD integration tests in `tests/quick_fix_pl410_bdd.rs` covering `next`, `last`, `redo`, title naming, invalid-range guard, and dispatch smoke test

## Constraints

- Fix must only apply when the diagnostic code is exactly `PL410` / `LoopControlUndefinedLabel`
- Must validate byte range before accessing source text (`valid_diagnostic_range`)
- The extracted operator must be one of `next`, `last`, `redo` — anything else returns no actions
- `is_preferred: true` since there is exactly one sensible fix

## Files / surfaces

- `crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs`
- `crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs`
- `crates/perl-lsp-rs-core/tests/quick_fix_pl410_bdd.rs` (new)

## Validation

All 6 new BDD tests pass; 17 pre-existing `quick_fix_new_codes_bdd` tests remain green; `cargo clippy` and `cargo xtask fmt` clean.

## Non-goals

- Does not add a "suggest defining a label" alternative fix (would require AST rewrite guidance — out of scope)
- Does not affect the diagnostic emission logic in `loop_control_label.rs`

PR #9635, opened 2026-05-28

EffortlessSteven:
## Summary

Closes #9612.

When a `next LABEL`, `last LABEL`, or `redo LABEL` statement references a label that doesn't exist in the file, Perl will die at runtime with `Label not found for "next LABEL"`. The diagnostic system already flags this as **PL410**, but clicking the lightbulb returned no code actions — `diagnostic_routes.rs` had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so every PL410 silently fell through to `_ => {}`.

This PR wires up the missing quick-fix so editors can offer a one-click repair.

Added `fix_loop_control_undefined_label(source, diagnostic)`:

- Calls `valid_diagnostic_range` (shared guard) to reject invalid / non-char-boundary ranges — same defensive pattern used by `fix_unused_import`, `fix_deprecated_array_base`, etc.
- Extracts the operator keyword (`next` / `last` / `redo`) and label name from the source text at the diagnostic range using `split_ascii_whitespace`.
- Returns empty `Vec` when: range is invalid, the first token isn't a recognised loop-control keyword, or the second token (label) is absent.
- Produces a single `CodeAction` that replaces `{op} LABEL` with just `{op}`, making the statement target the innermost enclosing loop.
- Sets `is_preferred: true` — there is exactly one sensible mechanical fix.
- Title format: `"Drop undefined label '{LABEL}' from '{op}'"` (includes both the label name and the operator for unambiguous identification in multi-diagnostic scenarios).

Added a match arm immediately after PL602:

```rust
// PL410: Loop-control statement targets an undefined label
}
```

### `crates/perl-lsp-rs-core/tests/quick_fix_pl410_bdd.rs` *(new)*

6 BDD integration tests following the Gherkin GIVEN/WHEN/THEN pattern established by `quick_fix_new_codes_bdd.rs`:

| # | Test | What it verifies |
|---|------|-----------------|
| 1 | `next_undefined_label_produces_drop_action` | `next OUTER` → action with correct title, kind, `is_preferred` |
| 2 | `last_undefined_label_produces_drop_action` | `last MISSING` → action + edit produces `last` without the label |
| 3 | `redo_undefined_label_produces_drop_action` | `redo NOWHERE` → action with correct title, kind, `is_preferred` |
| 4 | `title_names_operator_and_label` | Title includes both operator name and label name |
| 5 | `non_char_boundary_range_produces_no_action` | Range straddling a UTF-8 boundary → no action produced |
| 6 | `pl410_dispatch_smoke_test` | All three operators (`next ALPHA`, `last BETA`, `redo GAMMA`) reach the handler via the dispatch table |

## Test results

```
running 6 tests
test last_undefined_label_produces_drop_action ... ok
test next_undefined_label_produces_drop_action ... ok
test non_char_boundary_range_produces_no_action ... ok
test pl410_dispatch_smoke_test ... ok
test redo_undefined_label_produces_drop_action ... ok
test title_names_operator_and_label ... ok

```

All 17 pre-existing `quick_fix_new_codes_bdd` tests remain green. `cargo clippy -p perl-lsp-rs-core --lib` reports zero errors. `cargo xtask fmt` clean.

## Design decisions

- **Source-text extraction over message parsing**: The operator and label are parsed from `source[range.0..range.1]` rather than the diagnostic message string. This is more robust — the message format is a display concern, while the source text at the reported range is a structural fact.
- **No "suggest adding a label" alternative**: The issue spec explicitly excludes this. Adding a label is an AST-rewrite operation requiring context the quick-fix layer doesn't have.
- **`is_preferred: true`**: There is exactly one sensible safe fix (drop the label). A second competing action would only add confusion.

## Non-goals

- Does not touch the diagnostic emission logic in `loop_control_label.rs`.
- Does not add a Perl::Critic policy alias arm (no such alias exists for PL410).

---
_Generated by [Claude Code](https://claude.ai/code/session_01FscfksMhbmQ8qAfPzLBNKA)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -196,6 +196,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         c if c == DiagnosticCode::SecuritySignalHandler.as_str() => {
             actions.extend(quick_fixes::fix_security_signal_handler(source, &qf_diag));
         }
+        // PL410: Loop-control statement targets an undefined label
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         // PL405: printf/sprintf format specifier count mismatch
         c if c == DiagnosticCode::PrintfFormatMismatch.as_str()
             || c == "native.common.printf_format_arity" =>
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1972,6 +1972,46 @@ fn diagnostic_line_range(source: &str, range: (usize, usize)) -> Option<(usize, 
     Some((line_start, line_end))
 }
 
+/// Drop the undefined label from a `next LABEL` / `last LABEL` / `redo LABEL`
+/// statement so it targets the innermost enclosing loop instead (PL410).
+///
+/// Perl dies at runtime with `Label not found for "next LABEL"` when a
+/// loop-control statement names a label that does not exist in the file.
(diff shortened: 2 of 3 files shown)


EffortlessSteven:

PR #9638, opened 2026-05-28

EffortlessSteven:
## Summary

Closes #9612.

`next LABEL`, `last LABEL`, and `redo LABEL` statements that reference an undefined label produce a PL410 diagnostic, but clicking the lightbulb returned **no code actions** — the dispatch table in `diagnostic_routes.rs` had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so all PL410 diagnostics silently fell through to `_ => {}`.

At runtime this is fatal: Perl dies with `Label not found for "next LABEL"`. The only mechanical fix is to drop the label so the statement targets the innermost enclosing loop, which is always valid.

- Added `fix_loop_control_undefined_label(source, diagnostic)`:
  - Validates the byte range with the existing `valid_diagnostic_range` guard (returns empty if range is out-of-bounds or not on a char boundary)
  - Extracts the loop-control operator from the start of the diagnostic span
  - Returns no actions unless the operator is exactly `next`, `last`, or `redo`
  - Returns one preferred `QuickFix` action replacing `op LABEL` with bare `op`

- Added routing arm matching `DiagnosticCode::LoopControlUndefinedLabel.as_str()` ("PL410") to the new handler

7 BDD integration tests following the established GIVEN/WHEN/THEN pattern:

| Test | Covers |
|------|--------|
| `next_undefined_label_produces_drop_label_action` | `next` operator fires, action is preferred, new_text is `"next"` |
| `next_drop_label_edit_removes_label_from_source` | Applying the edit produces source with bare `next` and no label text |
| `last_undefined_label_produces_drop_label_action` | `last` operator fires, new_text is `"last"` |
| `redo_undefined_label_produces_drop_label_action` | `redo` operator fires, new_text is `"redo"` |
| `action_title_names_the_operator` | Title contains the operator for all three cases |
| `invalid_range_returns_no_action` | Out-of-bounds range → no action produced |
| `all_three_operators_reach_pl410_handler` | Smoke test: all three ops in one source file each produce an action |

## Validation

- **7 new tests**: all pass
- **17 pre-existing `quick_fix_new_codes_bdd` tests**: all still pass
- `cargo clippy -p perl-lsp-rs-core --lib`: clean (no new warnings)
- `cargo fmt -p perl-lsp-rs-core -- --check`: clean

## Design notes

The fix is one-action-only (`is_preferred: true`) because there is exactly one safe mechanical transformation: drop the label and target the innermost loop. The alternative (defining the missing label on an enclosing loop) requires structural AST rewrite guidance and is deliberately out of scope, consistent with the issue spec and the existing `suggestion` field on the diagnostic that already tells the user both options.

The implementation reuses the existing `valid_diagnostic_range` helper (also used by `fix_security_signal_handler` and `fix_printf_format_arity`) for consistent byte-safety across all quick-fix handlers.

---
_Generated by [Claude Code](https://claude.ai/code/session_015KZWaVhF4hozSbce8dtU1p)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -202,6 +202,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         {
             actions.extend(quick_fixes::fix_printf_format_arity(source, &qf_diag));
         }
+        // PL410: loop-control statement targets undefined label
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         _ => {}
     }
 
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1983,3 +1983,38 @@ fn valid_diagnostic_range(source: &str, range: (usize, usize)) -> Option<(usize,
     }
     Some((start, end))
 }
+
+/// Drop an undefined label from a `next LABEL`, `last LABEL`, or `redo LABEL`
+/// statement (PL410).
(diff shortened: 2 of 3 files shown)


EffortlessSteven:

PR #9641, opened 2026-05-29
feat(code-actions): add quick-fix for PL410 undefined loop-control label (#9612)

EffortlessSteven:
## Summary

Closes #9612.

`next LABEL`, `last LABEL`, and `redo LABEL` statements whose label is not defined in the file receive a PL410 diagnostic, but clicking the lightbulb in an editor returned **no code actions** — the diagnostic route table had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so every PL410 diagnostic silently fell through to `_ => {}`.

This is a runtime-fatal condition in Perl: if the named label doesn't exist, Perl dies at runtime with `Label not found for "next LABEL"`. The only safe automated fix is to drop the label so the statement targets the innermost enclosing loop.

Added `fix_loop_control_undefined_label`:

- Validates the byte range with the existing `valid_diagnostic_range` guard before touching source text
- Extracts the operator keyword from the source at the diagnostic range (first whitespace-delimited token)
- Guards that the op is exactly one of `next`, `last`, `redo` — any other value returns no actions
- Returns a single `is_preferred: true` action that replaces the entire `op LABEL` span with the bare `op`
- Title clearly names both the op and the effect: `"Remove undefined label — write '{op}' to target innermost loop"`

Added the PL410 routing arm just before the catch-all `_ => {}`:

```rust
// PL410: next/last/redo LABEL references an undefined label
}
```

6 BDD integration tests following the same GIVEN/WHEN/THEN pattern as `quick_fix_new_codes_bdd.rs`:

| Test | What it verifies |
|---|---|
| `pl410_next_undefined_label_produces_action` | `next LABEL` → action produced, `is_preferred`, label removed from result |
| `pl410_last_undefined_label_produces_action` | `last LABEL` → action produced, `is_preferred`, label removed |
| `pl410_redo_undefined_label_produces_action` | `redo LABEL` → action produced, `is_preferred`, label removed |
| `pl410_action_title_names_the_op` | Title contains the op name and describes the fix intent |
| `pl410_invalid_range_returns_no_actions` | Out-of-bounds range → zero actions (range guard) |
| `pl410_dispatch_smoke_reaches_handler` | Dispatch table wire-up confirmed end-to-end |

## Test results

```
running 6 tests
test pl410_action_title_names_the_op ... ok
test pl410_dispatch_smoke_reaches_handler ... ok
test pl410_invalid_range_returns_no_actions ... ok
test pl410_last_undefined_label_produces_action ... ok
test pl410_next_undefined_label_produces_action ... ok
test pl410_redo_undefined_label_produces_action ... ok

running 17 tests (quick_fix_new_codes_bdd — pre-existing)
test result: ok. 17 passed; 0 failed; 0 ignored
```

`cargo clippy -p perl-lsp-rs-core --lib` — no new warnings. `cargo fmt --check` — clean.

## Non-goals (per spec)

- Does not add a "suggest defining a label" alternative fix — that would require AST-level rewrite guidance outside the quick-fix layer
- Does not affect the PL410 diagnostic emission logic in `loop_control_label.rs`
- Does not handle the `GotoUndefinedLabel` (PL409) case — separate diagnostic code, separate issue if needed

---
_Generated by [Claude Code](https://claude.ai/code/session_01FPjGncFfN9Rhnj6p83VuVX)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -202,6 +202,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         {
             actions.extend(quick_fixes::fix_printf_format_arity(source, &qf_diag));
         }
+        // PL410: next/last/redo LABEL references an undefined label
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         _ => {}
     }
 
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1983,3 +1983,38 @@ fn valid_diagnostic_range(source: &str, range: (usize, usize)) -> Option<(usize,
     }
     Some((start, end))
 }
+
+/// Drop undefined label from `next`/`last`/`redo LABEL` (PL410).
+///
+/// The only safe automated fix is to remove the label so the statement
+/// targets the innermost enclosing loop. For example, `next OUTER` becomes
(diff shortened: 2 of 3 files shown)


EffortlessSteven:

PR #9645, opened 2026-05-30

EffortlessSteven:
## Problem

`next LABEL`, `last LABEL`, and `redo LABEL` statements that reference an undefined label receive a PL410 diagnostic, but clicking the lightbulb in the editor returned **no code actions**. The diagnostic route table in `diagnostic_routes.rs` had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so all PL410 diagnostics silently fell through to `_ => {}`.

This is a runtime-fatal condition in Perl — if the named label doesn't exist, Perl dies with `Label not found for "next LABEL"`. The mechanical fix is obvious: drop the label so the statement targets the innermost enclosing loop.

Closes #9612.

Added `fix_loop_control_undefined_label(source, diagnostic)`:

- Validates the byte range with the existing `valid_diagnostic_range` guard (returns no actions on bad/non-char-boundary ranges rather than panicking)
- Extracts the operator (`next`, `last`, or `redo`) by prefix-matching the source text at the diagnostic start — anything else returns no actions
- Walks past the operator to find the label identifier (ASCII alphanumeric + underscore run)
- Emits a single `QuickFix` action with `is_preferred: true` that deletes ` LABEL` from the statement, leaving bare `next`/`last`/`redo`
- The action title names both the label being removed and the operator so the user understands exactly what changes: `"Remove label 'MISSING' — use bare 'next' to target innermost loop"`

The approach is intentionally conservative: introducing a matching `LABEL:` on an enclosing loop is out of scope because it requires AST rewrite guidance the code-action layer cannot supply (see spec non-goals).

Added one routing arm immediately before the PL602 block:

```rust
// PL410: Loop-control statement references an undefined label
}
```

6 BDD integration tests covering the full contract:

| Test | What it verifies |
|------|-----------------|
| `next_pl410_produces_drop_label_action` | `next MISSING` → preferred QuickFix action exists, title names the label |
| `last_pl410_edit_removes_label_from_source` | Applying the edit to `last NOPE` yields `last;` with no trace of `NOPE` |
| `redo_pl410_produces_drop_label_action` | `redo NOWHERE` → action exists, edit yields `redo;` |
| `pl410_action_title_names_the_operator` | Title contains both the operator name and the removed label |
| `pl410_non_char_boundary_range_produces_no_action` | Range splitting a multi-byte UTF-8 char is rejected gracefully |
| `pl410_dispatch_smoke_test` | End-to-end: route table is wired; at least one action produced for valid PL410 |

## Verification

```
cargo test -p perl-lsp-rs-core --test quick_fix_pl410_bdd
# running 6 tests … test result: ok. 6 passed

cargo test -p perl-lsp-rs-core --test quick_fix_new_codes_bdd
# running 17 tests … test result: ok. 17 passed

cargo test -p perl-lsp-rs-core --lib
# test result: ok. 1921 passed

cargo fmt -p perl-lsp-rs-core -- --check
# (clean, no diff)

cargo clippy -p perl-lsp-rs-core --lib
# 0 new warnings introduced (one pre-existing Clone/Copy warning in inline_completion)
```

## Design notes

- **Why only one action?** The spec says `is_preferred: true` because there is exactly one sensible mechanical fix. Offering "define the label instead" would require knowing which enclosing loop to attach it to — that's an AST rewrite the action layer can't safely synthesize.
- **Why extract from source text rather than the message?** Extracting from source at the validated range start is more robust: the message format could change, the source won't. The operator extraction uses a simple prefix match on `"next "`, `"last "`, `"redo "` which is unambiguous at a LoopControl node's start position.
- **Guard pattern** follows exactly the `fix_security_signal_handler` and `fix_printf_format_arity` precedents already established in the file.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BQ6G1cBUfh9goNHz5yTX7j)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -192,6 +192,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         c if c == DiagnosticCode::DeprecatedArrayBase.as_str() => {
             actions.extend(quick_fixes::fix_deprecated_array_base(source, &qf_diag));
         }
+        // PL410: Loop-control statement references an undefined label
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         // PL602: Global signal handler assignment
         c if c == DiagnosticCode::SecuritySignalHandler.as_str() => {
             actions.extend(quick_fixes::fix_security_signal_handler(source, &qf_diag));
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1972,6 +1972,67 @@ fn diagnostic_line_range(source: &str, range: (usize, usize)) -> Option<(usize, 
     Some((line_start, line_end))
 }
 
+/// Drop the undefined label from a `next LABEL`, `last LABEL`, or `redo LABEL`
+/// statement (PL410).
+///
+/// The only safe mechanical transformation when the named label does not exist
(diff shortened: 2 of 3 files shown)


EffortlessSteven:

PR #9662, opened 2026-06-02

EffortlessSteven:
## Summary

Closes #9612.

`next LABEL`, `last LABEL`, and `redo LABEL` statements that reference an undefined label receive a PL410 diagnostic. Before this PR, clicking the lightbulb in an editor returned **no code actions** — the dispatch table in `diagnostic_routes.rs` had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so all PL410 diagnostics fell through to `_ => {}`.

This PR wires up a quick-fix handler that drops the undefined label, so `next FOO` becomes `next`, targeting the innermost enclosing loop — the only safe mechanical fix (an undefined label is runtime-fatal in Perl).

Added `fix_loop_control_undefined_label(source, diagnostic) -> Vec<CodeAction>`:

- Calls `valid_diagnostic_range` to guard against out-of-bounds or non-char-boundary ranges (same pattern as all other handlers in this file)
- Parses the operator (`next` / `last` / `redo`) from the backtick-quoted prefix of the diagnostic message (`` `next FOO` references a label... ``)
- Whitelist-checks the operator — anything other than the three documented ops returns no actions, preventing misrouted diagnostics from producing unexpected edits
- Sanity-checks that the source text at the range actually starts with the operator
- Returns a single `is_preferred: true` action titled `"Drop undefined label: write \`{op}\` to target innermost loop"` that replaces the full `op LABEL` span with just `op`

Added private helper `parse_loop_control_op(message) -> Option<&str>` that extracts the first word from inside the backtick pair.

Added a new match arm for `DiagnosticCode::LoopControlUndefinedLabel` (PL410) immediately after the PL405 arm, routing to the new handler.

### `crates/perl-lsp-rs-core/tests/quick_fix_pl410_bdd.rs` *(new)*

Six BDD integration tests following the same GIVEN/WHEN/THEN pattern as `quick_fix_new_codes_bdd.rs`:

| Test | What it covers |
|------|---------------|
| `next_undefined_label_produces_drop_action` | `next FOO` → action title, kind, `is_preferred` |
| `next_undefined_label_edit_strips_label_from_source` | Edit correctness: `next FOO` → `next` in full source context |
| `last_undefined_label_produces_drop_action` | `last MISSING` → action + edit correctness |
| `redo_undefined_label_produces_drop_action` | `redo NOWHERE` → action + edit correctness |
| `invalid_range_produces_no_action` | Out-of-bounds range yields no action (guard path) |
| `all_three_loop_ops_reach_pl410_handler` | Dispatch smoke test: all three ops routed correctly |

## Test results

```
running 6 tests
test all_three_loop_ops_reach_pl410_handler ... ok
test next_undefined_label_produces_drop_action ... ok
test next_undefined_label_edit_strips_label_from_source ... ok
test last_undefined_label_produces_drop_action ... ok
test redo_undefined_label_produces_drop_action ... ok
test invalid_range_produces_no_action ... ok

```

The 17 pre-existing `quick_fix_new_codes_bdd` tests remain green. `cargo fmt --check` and `cargo clippy --lib` (no new warnings) are clean.

## Design decisions

- **Single action, `is_preferred: true`**: The spec calls for exactly one sensible fix (drop the label). There is no "suggest defining the label" alternative because that would require AST rewrite guidance and is out of scope per #9612.
- **Message-based op extraction**: The operator is parsed from the message rather than re-scanning source, consistent with how other handlers (PL700, PL301, PL300) extract names — the diagnostic already carried this information forward.
- **Range replaces full `op LABEL` span**: The diagnostic range from `loop_control_label.rs` covers the entire node (`node.location.start..node.location.end`), which is `next FOO` without the trailing semicolon. Replacing that span with `next` leaves the semicolon untouched.

## Non-goals (per spec)

- Does not add a "suggest defining a label" alternative (requires AST rewrite guidance)
- Does not affect the diagnostic emission logic in `loop_control_label.rs`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qeorxdg2LZnFEokudkLivN)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -202,6 +202,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         {
             actions.extend(quick_fixes::fix_printf_format_arity(source, &qf_diag));
         }
+        // PL410: next/last/redo LABEL where LABEL is not defined in this file
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         _ => {}
     }
 
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1964,6 +1964,64 @@ fn parse_printf_format_mismatch(message: &str) -> Option<(usize, String)> {
     specifiers.checked_sub(supplied).filter(|&n| n > 0).map(|n| (n, call_name))
 }
 
+/// Drop the undefined label from a loop-control statement (PL410).
+///
+/// `next LABEL`, `last LABEL`, and `redo LABEL` where `LABEL` is not defined
+/// anywhere in the current file are runtime-fatal in Perl. The only safe
+/// mechanical fix is to drop the label so the statement targets the innermost
+/// enclosing loop instead.
+///
+/// The diagnostic range covers the full `next LABEL` token span. We replace
+/// that span with just the operator keyword.
+pub fn fix_loop_control_undefined_label(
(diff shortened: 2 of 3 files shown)


EffortlessSteven:

Issue #9715, opened 2026-06-05
Lock empty folding-range edge cases in UX receipt


EffortlessSteven:
Problem: The folding-range UX contract says empty files return an empty array, but scenario 27 only checked that empty and single-line files did not error.\n\nFix: Tighten scenario 27 so empty and single-line files must return empty folding-range arrays.

EffortlessSteven:
Closed as part of UX receipts disposition (EffortlessMetrics/perl-lsp-swarm#1731). The UX receipt infrastructure has shipped in the main codebase via commits starting with 5cb6d18 (queue: merge framework). This tracking issue is now complete.

EffortlessSteven:
Resolved via receipt infrastructure deployment.

EffortlessSteven:
Closed as part of UX receipts disposition (EffortlessMetrics/perl-lsp-swarm#1731). The UX receipt infrastructure has shipped in the main codebase via commits starting with 5cb6d18 (queue: merge framework). This tracking issue is now complete.

PR #9716, opened 2026-06-05
test(ux): lock empty folding ranges (#9715)

EffortlessSteven:
Problem: Scenario 27 only checked that empty and single-line foldingRange requests did not error.\nFix: Require empty and single-line files to return empty folding-range arrays, matching the UX contract.\nVerification: `cargo test -p perl-lsp-ux-tests --test ux_scenario_27_folding_range --profile agent --locked` passes; `cargo test -p perl-lsp-ux-tests --test editor_ux_fixture_matrix --profile agent --locked` passes; `just agent-pr-fast` passes.


EffortlessSteven:
Landed via consolidated train-3 PR #9917 (queue convergence 2026-06-06) — scenario/fixture union with per-PR integration notes in its body; LSP 3.18 claim-guard markers preserved.

Issue #9717, opened 2026-06-05
UX: lock same-file rename edit receipt


EffortlessSteven:
Problem: The core rename UX scenario still accepted degraded null output for a simple same-file subroutine rename, leaving the first-session edit payload unprotected.\n\nExpected: Scenario 23 should require concrete same-file edits for the declaration and call sites, and the matrix should describe that user-visible contract.

EffortlessSteven:
Closed as part of UX receipts disposition (EffortlessMetrics/perl-lsp-swarm#1731). The UX receipt infrastructure has shipped in the main codebase via commits starting with 5cb6d18 (queue: merge framework). This tracking issue is now complete.

EffortlessSteven:
Resolved via receipt infrastructure deployment.

PR #9718, opened 2026-06-05
test(ux): lock same-file rename edits (#9717)

EffortlessSteven:
Problem: The core rename UX scenario accepted degraded null output for a simple same-file subroutine rename.
Fix: Require scenario 23 to return same-file edits for the declaration and call sites, and update the fixture matrix receipt.
Verification: `cargo test -p perl-lsp-ux-tests --test ux_scenario_23_rename_workflow --profile agent --locked -- --nocapture --test-threads=1` passes; `cargo test -p perl-lsp-ux-tests --test editor_ux_fixture_matrix --profile agent --locked` passes; `cargo xtask fmt --check --package perl-lsp-ux-tests` passes; `cargo clippy -p perl-lsp-ux-tests --test ux_scenario_23_rename_workflow --profile agent --locked -- -D warnings -A missing_docs` passes; `cargo check -p perl-lsp-ux-tests --all-targets --profile agent --locked` passes; `just agent-pr-fast` passes.

### DIFF
--- a/crates/perl-lsp-ux-tests/fixtures/editor_ux_fixture_matrix.json
+++ b/crates/perl-lsp-ux-tests/fixtures/editor_ux_fixture_matrix.json
@@ -581,7 +581,7 @@
       ],
       "expected_outcomes": [
         "prepareRename and rename do not error",
-        "rename returns clean null or valid workspace edit for the opened file"
+        "same-file subroutine rename returns edits for the declaration and call sites"
       ],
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Landed via consolidated train-3 PR #9917 (queue convergence 2026-06-06) — scenario/fixture union with per-PR integration notes in its body; LSP 3.18 claim-guard markers preserved.

PR #9905, opened 2026-06-05

EffortlessSteven:
## Summary

Closes #9612.

`next LABEL`, `last LABEL`, and `redo LABEL` statements that reference an undefined label produce a PL410 diagnostic but previously returned **no code actions** — the lightbulb was empty because `diagnostic_routes.rs` had no arm for `DiagnosticCode::LoopControlUndefinedLabel`.

At runtime this is fatal: Perl dies with `Label not found for "next LABEL"`. The only safe mechanical fix is to drop the label so the statement targets the innermost enclosing loop, which is exactly what bare `next;` / `last;` / `redo;` do.

### Changes

**`quick_fixes.rs`** — new `fix_loop_control_undefined_label`:
- Validates the byte range via `valid_diagnostic_range` before touching source text; invalid/non-char-boundary ranges return no actions.
- Extracts the operator from the leading whitespace-split token of the diagnostic range.
- Guards that the operator is one of `next | last | redo`; anything unexpected returns no actions.
- Replaces the full `OP LABEL` span with the bare operator string.
- `is_preferred: true` — there is exactly one sensible mechanical fix.

**`diagnostic_routes.rs`** — wire `PL410` / `LoopControlUndefinedLabel` to the new handler, consistent with how every other diagnostic code is routed in the match arm chain.

**`tests/quick_fix_pl410_bdd.rs`** (new) — 7 BDD integration tests:
| Test | What it covers |
|------|---------------|
| `next_label_produces_drop_label_action` | action produced for `next OUTER`, kind=QuickFix, is_preferred |
| `next_label_edit_leaves_bare_next` | applying the edit produces `next;`, removes the label name |
| `last_label_edit_leaves_bare_last` | same for `last PHANTOM` |
| `redo_label_edit_leaves_bare_redo` | same for `redo NOWHERE` |
| `action_title_names_the_operator` | title includes the operator so users know what is being fixed |
| `invalid_range_produces_no_action` | out-of-bounds range → no action (guard works) |
| `pl410_dispatch_reaches_handler` | smoke test: PL410 + PL700 diagnostics both produce actions |

## Test plan

- [x] 7 new `quick_fix_pl410_bdd` tests pass
- [x] 17 pre-existing `quick_fix_new_codes_bdd` tests remain green
- [x] `cargo clippy -p perl-lsp-rs-core` — no new warnings introduced
- [x] `cargo fmt -p perl-lsp-rs-core` — no formatting changes needed in production code; test file auto-formatted

## What was not done (per spec)

- No "suggest defining a label" alternative fix — that would require AST-rewrite guidance and is explicitly out of scope in the issue.
- No changes to the diagnostic emission logic in `loop_control_label.rs` — that is correct and untouched.

https://claude.ai/code/session_01BT6QBZpp941sLqa8KeqpbD

---
_Generated by [Claude Code](https://claude.ai/code/session_01BT6QBZpp941sLqa8KeqpbD)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -202,6 +202,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         {
             actions.extend(quick_fixes::fix_printf_format_arity(source, &qf_diag));
         }
+        // PL410: next/last/redo targets an undefined label
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         _ => {}
     }
 
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1983,3 +1983,37 @@ fn valid_diagnostic_range(source: &str, range: (usize, usize)) -> Option<(usize,
     }
     Some((start, end))
 }
+
+/// Drop an undefined label from a `next LABEL`, `last LABEL`, or `redo LABEL`
+/// statement (PL410).
+///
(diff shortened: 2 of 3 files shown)


EffortlessSteven:
Closing: PL410 undefined-loop-control-label quick-fix already landed in source/master (commit 1da4c25e6). Queue convergence 2026-06-06.

PR #9907, opened 2026-06-05
feat(code-actions): add quick-fix for PL410 undefined loop-control label (#9612)

EffortlessSteven:
## Summary

Closes #9612.

`next LABEL`, `last LABEL`, and `redo LABEL` statements that reference an undefined label receive a PL410 diagnostic, but clicking the lightbulb in an editor previously returned **no code actions**. The `diagnostic_routes.rs` dispatch table had no arm for `DiagnosticCode::LoopControlUndefinedLabel`, so all PL410 diagnostics fell through silently to `_ => {}`.

This PR wires the missing route and implements the one sensible automated repair: drop the label so the statement targets the innermost enclosing loop.

## Why this fix is safe

At runtime Perl dies with `Label not found for "next LABEL"` — there is no recoverable behavior. The only mechanical repair that preserves intent (continue iterating the nearest loop) is to drop the label. Because there is exactly one sensible fix, the action is `is_preferred: true`.

The fix does **not** suggest adding a matching label: that would require AST-rewrite guidance and is outside scope (the spec calls this out explicitly as a non-goal).

New function `fix_loop_control_undefined_label`:

- Validates the byte range with the shared `valid_diagnostic_range` guard (returns `Vec::new()` on out-of-bounds or non-char-boundary range).
- Extracts the operator (`next`, `last`, `redo`) from the start of the source text at the diagnostic range.
- Guards against unknown operators — future variants won't accidentally produce wrong fixes.
- Guards against empty label (bare `next`/`last`/`redo` shouldn't carry PL410, but the check is cheap).
- Returns a single `CodeAction` replacing `OP LABEL` with bare `OP`, titled `"Drop undefined label 'LABEL' (target innermost loop)"`, `is_preferred: true`.

Added match arm for `PL410 / LoopControlUndefinedLabel`:

```rust
// PL410: next/last/redo LABEL references an undefined label
}
```

6 BDD integration tests following the `GIVEN / WHEN / THEN` pattern from `quick_fix_new_codes_bdd.rs`:

| # | Test | What it verifies |
|---|------|-----------------|
| 1 | `pl410_next_undefined_label_produces_drop_action` | `next OUTER` → action drops label, `is_preferred`, edit removes label text |
| 2 | `pl410_last_undefined_label_produces_drop_action` | `last MISSING` → same for `last` |
| 3 | `pl410_redo_undefined_label_produces_drop_action` | `redo NOWHERE` → same for `redo` |
| 4 | `pl410_action_title_includes_label_name` | Title contains the label name so the user can identify it |
| 5 | `pl410_out_of_bounds_range_returns_no_actions` | Out-of-bounds range returns no actions (range guard works) |
| 6 | `pl410_dispatch_smoke_test_handler_is_wired` | End-to-end through `CodeActionsProvider` — dispatch arm is present |

## Test results

```
running 6 tests
test pl410_action_title_includes_label_name ... ok
test pl410_dispatch_smoke_test_handler_is_wired ... ok
test pl410_last_undefined_label_produces_drop_action ... ok
test pl410_next_undefined_label_produces_drop_action ... ok
test pl410_redo_undefined_label_produces_drop_action ... ok
test pl410_out_of_bounds_range_returns_no_actions ... ok

test result: ok. 6 passed; 0 failed
```

Pre-existing `quick_fix_new_codes_bdd` suite: **17 passed, 0 failed** (no regressions).

`cargo clippy -p perl-lsp-rs-core --lib`: clean.

## Non-goals (per spec)

- Does not add a "suggest defining a label" alternative fix.
- Does not affect the diagnostic emission logic in `loop_control_label.rs`.
- Does not change the `GotoUndefinedLabel` (PL409) handler.

https://claude.ai/code/session_01Q27CQ4KYxfgp89uxWvdxXv

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q27CQ4KYxfgp89uxWvdxXv)_

### DIFF
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/diagnostic_routes.rs
@@ -202,6 +202,10 @@ fn quick_fixes_for_diagnostic(source: &str, diagnostic: &Diagnostic) -> Vec<Code
         {
             actions.extend(quick_fixes::fix_printf_format_arity(source, &qf_diag));
         }
+        // PL410: next/last/redo LABEL references an undefined label
+        c if c == DiagnosticCode::LoopControlUndefinedLabel.as_str() => {
+            actions.extend(quick_fixes::fix_loop_control_undefined_label(source, &qf_diag));
+        }
         _ => {}
     }
 
--- a/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
+++ b/crates/perl-lsp-rs-core/src/providers/code_actions/quick_fixes.rs
@@ -1983,3 +1983,45 @@ fn valid_diagnostic_range(source: &str, range: (usize, usize)) -> Option<(usize,
     }
     Some((start, end))
 }
+
+/// Drop the undefined label from a `next`/`last`/`redo LABEL` statement (PL410).
+///
+/// At runtime Perl dies with `Label not found for "next LABEL"` — the only
+/// safe mechanical fix is to remove the label so the statement targets the
+/// innermost enclosing loop. This fix is `is_preferred: true` because there
+/// is exactly one sensible automated repair.
+pub fn fix_loop_control_undefined_label(
+    source: &str,
+    diagnostic: &QuickFixDiagnostic,
+) -> Vec<CodeAction> {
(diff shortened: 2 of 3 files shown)


EffortlessSteven:
Closing: PL410 undefined-loop-control-label quick-fix already landed in source/master (commit 1da4c25e6). Queue convergence 2026-06-06.

PR #10012, opened 2026-07-18
ci(badge-endpoints): tolerate Actions PR-creation permission denial (#0000)

EffortlessSteven:
Problem: The 'Badge Endpoints' workflow fails on master because the repo Actions settings disallow the default GITHUB_TOKEN from opening PRs — peter-evans/create-pull-request@v8 exits with 'GitHub Actions is not permitted to create or approve pull requests', turning the run red.

Fix: Give that step an id (`cpr`) and `continue-on-error: true` so the permission denial no longer fails the job, and add a follow-up step (`if: steps.cpr.outcome == 'failure'`) that emits ::notice:: annotations explaining the badge PR was skipped and how to re-enable it. Badge regeneration, drift cleanup, and branch-reset steps are untouched and still fail loudly on genuine errors. Workflow-only change — no PAT secrets, no repo-settings changes.

Verification: YAML parsed cleanly (python3 yaml); `actionlint v1.7.7` reports no issues on .github/workflows/badge-endpoints.yml; full diff re-read for correctness. `just pr-fast` could not run: the pre-push hook's `nix develop` fails environmentally (flake pins rust 1.95.0, missing from the pinned nixpkgs rust-bin) — unrelated to this YAML-only change, so pushed with `--no-verify` per the hook's documented bypass policy for environmental failures.

Residual manual step: a maintainer may still want to enable 'Allow GitHub Actions to create and approve pull requests' under Settings > Actions > General to restore automated badge PRs; until then badge diffs are regenerated but the PR is skipped with a ::notice::.

### DIFF
--- a/.github/workflows/badge-endpoints.yml
+++ b/.github/workflows/badge-endpoints.yml
@@ -89,15 +89,11 @@ jobs:
           git push origin --delete automation/badge-endpoints || echo "automation/badge-endpoints did not exist remotely (or delete failed); continuing -- create-pull-request will create it fresh if needed"
 
       - name: Open badge refresh PR
-        id: cpr
-        # The repository's Actions settings may disallow the default
-        # GITHUB_TOKEN from opening pull requests ("GitHub Actions is not
-        # permitted to create or approve pull requests"). Badge regeneration
-        # above is the payload that must keep failing loudly; opening the PR
-        # is best-effort delivery, so tolerate that permission denial here and
-        # surface it as a ::notice:: in the next step instead of turning the
-        # whole run red.
-        continue-on-error: true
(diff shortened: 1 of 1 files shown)


EffortlessSteven:
Addressed the review concern about masking arbitrary `create-pull-request` failures.

The workflow no longer uses `continue-on-error`. Badge PR creation is now gated by the explicit repository variable `BADGE_ENDPOINTS_AUTO_PR=true`; when unset/false, badge generation still succeeds and a notice records the intentional manual follow-up. When enabled, any genuine PR-creation failure fails the job normally. The notice also documents the required Actions setting and manual fallback.

The branch was rebased onto current master; the diff remains workflow-only. YAML parsing and the repository workflow-policy lint pass.

PR #10019, opened 2026-07-18
fix(lsp): match qualified subroutine implementations (#6751)

### LINKED ISSUE #6751: [PATTERN] Qualified Name Comparison Bug in 6+ Files
## Summary

A recurring bug pattern has been identified across 6+ files where qualified names are compared using direct string equality (name == want_name) instead of proper qualified name parsing.

## Bug Pattern

**Incorrect pattern (currently used):**
```rust
name == want_name
```

**Correct pattern:**
```rust
split_qualified_name(name).1 == want_name
```

## Affected Files

Based on deep-dive analysis from PR #6730:

| File | Line(s) | Issue |
|------|---------|-------|
| src/parser/module.rs | ~245 | Direct comparison of qualified names |
| src/parser/subroutine.rs | ~189 | Same pattern in subroutine parsing |
| src/resolver/qualified.rs | ~78, ~112 | Multiple occurrences |
| src/checker/imports.rs | ~156 | Import name validation |
| src/checker/exports.rs | ~203 | Export name validation |
| src/ast/identifiers.rs | ~45 | Identifier comparison |

## Root Cause

Qualified names (e.g., Foo::Bar::baz) contain module path components that must be stripped before comparing the base identifier. The direct comparison fails when:
- Names have different module prefixes but same base name
- Names are fully qualified vs partially qualified
- Module aliases are used

## Proposed Fix

1. **Immediate fix**: Audit all occurrences and replace with split_qualified_name() calls
2. **Long-term fix**: Create a shared helper or macro
3. **Prevention**: Add lint rule to catch direct qualified name comparisons

## Related PRs
- #6730 (deep-dive analysis)

## Acceptance Criteria
- [ ] All 6+ files audited and fixed
- [ ] Shared helper/macro created in src/utils/qualified_names.rs
- [ ] Regression tests added
- [ ] Lint rule considered for future prevention

---
*This issue was created as part of the architectural pattern documentation initiative.*

EffortlessSteven:
## What this does

Qualified Perl declarations such as `sub Foo::process` were missed by semantic declaration lookup and the LSP implementation provider when callers searched for the bare method name.

**Fix:** compare the bare component returned by the shared qualified-name parser while preserving package-scope and non-matching-name guards.

## Verification

| Check | Result |
| --- | --- |
| qualified declaration regressions and negative controls | pass on current `origin/master` transplant |
| `cargo check --all-targets`, `cargo clippy` for both owning crates | pass |
| `just pr-fast` | 15/18 gates pass; three unrelated timing-sensitive suites failed under load, and representative failures pass individually on untouched `origin/master` |

## Review map

- `DeclarationProvider::collect_subroutine_declarations`: matches qualified declarations by bare name.
- `ImplementationProvider::find_method_in_package`: applies the same comparison while retaining package scope.
- Unit tests: cover qualified matches, different-name rejection, and cross-package rejection.

Closes #6751

### DIFF
--- a/crates/perl-lsp-rs/src/features/implementation_provider.rs
+++ b/crates/perl-lsp-rs/src/features/implementation_provider.rs
@@ -399,13 +399,14 @@ impl ImplementationProvider {
                         // Compare bare names so a qualified declaration such as
                         // `sub Foo::process` matches a lookup for `process`
                         // (issue #6751), mirroring `find_method_in_ast`.
-                        let (_, sub_bare) =
+                        let (sub_package, sub_bare) =
                             perl_parser::qualified_name::split_qualified_name(sub_name);
                         let (_, method_bare) =
                             perl_parser::qualified_name::split_qualified_name(method_name);
-                        if current_package.as_deref() == Some(package_name)
-                            && sub_bare == method_bare
-                        {
+                        let package_matches = sub_package == Some(package_name)
+                            || (sub_package.is_none()
+                                && current_package.as_deref() == Some(package_name));
+                        if package_matches && sub_bare == method_bare {
                             let target_uri = parse_uri(uri);
                             results.push(LocationLink {
                                 origin_selection_range: None,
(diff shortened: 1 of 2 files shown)


EffortlessSteven:
Addressed the unresolved review threads in commit `947926a1b`.

- Qualified subroutines now resolve by their explicit qualifier (`sub Foo::bar`) even when declared inside `package Other`; bare declarations still use their enclosing package.
- Added positive/negative regressions for both declaration lookup and implementation lookup, including rejection of `Foo::bar` under `Other`.
- A qualified lookup no longer falls back to an unrelated first declaration when the requested package has no match.

Verification on the updated branch:
- semantic targeted test: 1 passed
- implementation targeted test: 1 passed
- all-target check for `perl-semantic-analyzer` and `perl-lsp-rs`: passed
- clippy with only three pre-existing baseline lints allowed: passed
- unfiltered strict clippy still reports the existing unused import, non-octal permission literal, and unnecessary cast outside this diff.

EffortlessSteven:
CI diagnosis/follow-up: the previous meta-shard failure was an ancestry mismatch, not a code failure. The PR branch predated the current `master` workflow’s `ignored_tests_check_refs` gate, so CI ran a newer workflow against an older xtask checkout. Rebasing onto current `origin/master` preserves the same two-file PR diff and is now pushed at head `cf056fa61`; hosted checks are restarting against the aligned branch.

Issue #4926, opened 2026-04-23
quality: error classification (every error carries a category tag)


EffortlessSteven:
Parent: EffortlessMetrics/perl-lsp-swarm#2901 (Quality Wave). Agent-routing-aware.

> **This is an epic, not a single change.** The parser slice is merged. Everything remaining — system adoption, receipt exposure, routing proof, enforcement — is tracked through the named child issues below.

## Contract (merged)

Merged in #10025. `ErrorCategory` and the `ErrorClass` trait live in `crates/perl-parser-core/src/syntax/error/mod.rs` and are re-exported through the `perl-parser` facade.

```rust
pub trait ErrorClass {
    fn error_class(&self) -> ErrorCategory;
}

#[non_exhaustive]
pub enum ErrorCategory {
    Advisory,      // informational; does not describe invalid input
    UserError,     // input is invalid or needs a user-facing correction
    ResourceLimit, // a configured safety limit was exceeded
    Bug,           // our code violated an internal invariant
    Infra,         // an external dependency or service is unavailable
    Transient,     // the operation may succeed if retried
    Protocol,      // the other side violated a protocol or format contract
}
```

**Seven categories, not the five originally specified in this issue.** `Advisory` and `ResourceLimit` were added during the parser slice because a style advisory about *valid* code and a depth-bound rejection are neither user errors nor bugs, and routing them as either produces the wrong action — a builder assigned to a non-defect, or an unbounded retry of an input that will never fit.

**Stable machine identity is `ErrorCategory::as_str()`:** `advisory`, `user_error`, `resource_limit`, `bug`, `infra`, `transient`, `protocol`. Receipt and log layers record that token. `Debug` output is not identity and must not be serialized as such.

Adoption so far: `ParseError` only. It is the sole `ErrorClass` implementation in the repository.

## Governed surface

"Every error type" is too broad. Not every internal `io::Error`, temporary helper enum, or private implementation detail should implement `ErrorClass`. The governed surface is:

> Every **owned** error that crosses a **durable operational boundary** — public API, protocol response, structured log, receipt, agent handoff, or retry/escalation decision — must expose an `ErrorCategory`.

For foreign errors:

- wrap or classify them at the boundary;
- do not classify by matching on error message text;
- do not add traits to unrelated external types;
- preserve the underlying error as a separate `source`.

## Dependency placement — resolved, no move needed

[...]

PR #10025, opened 2026-07-19
feat(parser): classify parse errors for agent routing (#4926)

EffortlessSteven:
## What this does

`ParseError` reaches agentic callers through `perl-parser`'s public error surface with no stable operational category, so a caller cannot tell a style advisory about valid code from malformed input, or a depth limit from a cancelled parse. Routing on `Debug` strings or error text is the only alternative, and it breaks on any wording change.

**Fix:** add `ErrorCategory` and the `ErrorClass` trait to the parser error facade, so every `ParseError` carries a routing category and a stable machine token. Advisories stay advisories, recursion and nesting limits are resource limits, cancellation is transient, and malformed or recovered source is a user error. Existing diagnostics, severities, and syntax-kind APIs are unchanged.

## Verification

`ErrorCategory` is a new public surface, so there is no red/green against `main`: the tests pin the routing contract and the token spelling.

| Check | Result |
| --- | --- |
| routing boundaries | `Advisory` advisory, `RecursionLimit` and `NestingTooDeep` resource_limit, `Cancelled` transient, `UnexpectedEof` and `Recovered` user_error |
| future variants | `match self` is exhaustive, so a new `ParseError` variant fails to compile until it is classified |
| stable tokens | all 7 categories assert their `as_str` spelling |
| facade import path | `perl-parser` re-exports `ErrorCategory` and `ErrorClass` |
| `cargo fmt`, `clippy -D warnings` (lib), `check --all-targets` | clean for both crates |
| `just pr-fast` | 15 passed, 3 blocking failures reproduced on the unpatched parent (`release_history`, `release_history_check`, DAP `test_e2e_scopes_have_no_cross_contamination`) |

## Review map

- `crates/perl-parser-core/src/syntax/error/mod.rs`: defines the category, the trait, `as_str`, and the exhaustive `impl ErrorClass for ParseError`. `ErrorCategory` derives `Hash` so receipt layers can aggregate counts by category.
- `crates/perl-parser-core/src/lib.rs`, `src/engine/error/mod.rs`, `crates/perl-parser/src/{lib,core}.rs`: re-export the contract through the canonical parser surfaces.
- `crates/perl-parser/tests/error_category_tests.rs`: pins the facade import path and the four routing boundaries.

### DIFF
--- a/crates/perl-parser-core/src/engine/error/mod.rs
+++ b/crates/perl-parser-core/src/engine/error/mod.rs
@@ -27,7 +27,7 @@ pub mod recovery {
 
 /// Error types and result aliases used by the parser engine.
 pub use crate::syntax::error::{
-    BudgetTracker, ErrorContext, ParseBudget, ParseDiagnosticSeverity, ParseError, ParseOutput,
-    ParseResult, RecoveryKind, RecoverySalvageClass, RecoverySalvageProfile, RecoverySite,
-    get_error_contexts,
+    BudgetTracker, ErrorCategory, ErrorClass, ErrorContext, ParseBudget, ParseDiagnosticSeverity,
+    ParseError, ParseOutput, ParseResult, RecoveryKind, RecoverySalvageClass,
+    RecoverySalvageProfile, RecoverySite, get_error_contexts,
 };
--- a/crates/perl-parser-core/src/lib.rs
+++ b/crates/perl-parser-core/src/lib.rs
@@ -159,8 +159,8 @@ pub use ast::{GotoTargetForm, Node, NodeKind, SourceLocation};
 pub use error::classifier::{RecoverySalvageMetrics, classify_recovery_salvage};
 /// Parse error, budget, and output types.
 pub use error::{
-    BudgetTracker, ParseBudget, ParseDiagnosticSeverity, ParseError, ParseOutput, ParseResult,
-    RecoverySalvageClass, RecoverySalvageProfile,
+    BudgetTracker, ErrorCategory, ErrorClass, ParseBudget, ParseDiagnosticSeverity, ParseError,
+    ParseOutput, ParseResult, RecoverySalvageClass, RecoverySalvageProfile,
 };
 
 /// Builtin function signature lookup tables.
--- a/crates/perl-parser-core/src/syntax/error/mod.rs
+++ b/crates/perl-parser-core/src/syntax/error/mod.rs
@@ -349,6 +349,50 @@ impl BudgetTracker {
 /// strategies across all pipeline stages.
 pub type ParseResult<T> = Result<T, ParseError>;
 
+/// Operational category used by agentic tooling to route parser failures.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum ErrorCategory {
+    /// The diagnostic is informational and does not describe invalid input.
+    Advisory,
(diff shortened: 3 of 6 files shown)


EffortlessSteven (commented):
Review findings before this draft is promoted: (1) `ParseError::Advisory` is documented as a valid/non-invalidating construct, so routing it as `UserError` may cause agents to reject valid code; (2) `RecursionLimit` and `NestingTooDeep` are resource/depth limits and may need a distinct transient/resource category rather than malformed-source handling; (3) the manually maintained 13-variant matrix will not force an update when a future variant is added. The changed-scope tests and exports are coherent, but please resolve the taxonomy/coverage contract before merge. Broad local perl-parser-core testing also hit two unrelated pre-existing failures in `fix_proto_dollar_dollar.rs`.

EffortlessSteven:
Thanks, all three taxonomy findings are addressed in `fdfc3f6`.

**1. `Advisory` routed as `UserError`.** Fixed. `ErrorCategory::Advisory` is now its own category, and `ParseError::Advisory` routes to it. An advisory describes valid code, so agents surface it instead of rejecting the source.

**2. `RecursionLimit` / `NestingTooDeep` treated as malformed source.** Fixed. Both now route to `ErrorCategory::ResourceLimit`. These are structural bounds, not bad input, and retrying unchanged input will not cure them, so they are deliberately not `Transient` either. `Cancelled` remains the only `Transient` case.

**3. The 13-variant matrix will not force an update on a new variant.** Fixed by construction rather than by inventory. `impl ErrorClass for ParseError` uses an exhaustive `match self` with no wildcard arm, so adding a `ParseError` variant fails to compile until it is given a routing category. The hand-maintained matrix test is replaced by focused boundary tests (advisory, resource limit, transient, user error), which no longer drift as variants are added.

This expands the category vocabulary from five to seven (`Advisory` and `ResourceLimit` are new). `ErrorCategory::as_str` exposes a stable machine token per category so receipt and log layers can record it without depending on `Debug`. No serde dependency was added to `perl-parser-core`, which currently has none.

**On the unrelated failures.** Confirmed pre-existing, reproduced on the unpatched parent commit (`646c34b`) with the same commands: `just pr-fast` fails there on `release_history` and `release_history_check`, and DAP `test_e2e_scopes_have_no_cross_contamination` fails with `channel closed/timeout waiting for stopped`. `fix_proto_dollar_dollar` was likewise reproduced on untouched master earlier. Focused checks on the changed crates are clean: `cargo fmt`, `clippy --lib -D warnings`, `check --all-targets`, and the parser-core and facade tests all pass.

Scope note: this is the parser slice of #4926. LSP, DAP, receipt serialization, and enforcement remain open there and are tracked as follow-ups.

EffortlessSteven:
Follow-up review pass in `bc7e139`, both behavior-preserving:

- `ErrorCategory` now derives `Hash`, so receipt and log layers can aggregate counts with `HashMap<ErrorCategory, _>` without a newtype. Deriving it now avoids a later addition once consumers exist.
- The three new tests dropped their `Result<(), Box<dyn Error>>` returns. No `?` was used in any of them, so the trailing `Ok(())` was dead weight, and every sibling test in `syntax::error::tests` uses a plain `fn`.

Re-verified after the change: `cargo fmt --check` clean, `clippy --lib -D warnings` clean, `check --all-targets` 0 errors, and all three tests pass (`2 passed` in parser-core, `1 passed` in the facade).

Issue #10061, opened 2026-07-28
Sync residual RTK guidance cleanup from perl-lsp-swarm


EffortlessSteven:
## Problem

The source/release-lineage repository still publishes RTK-prefixed commands in current CI, coverage, accepted-spec, checklist, plan, generated quality-action, and positive policy-test surfaces. The complete cleanup has been reviewed and merged in `perl-lsp-swarm`.

## Authority

- development issue: `EffortlessMetrics/perl-lsp-swarm#5153`
- clean reviewed PR: `EffortlessMetrics/perl-lsp-swarm#5155`
- merged development commit: `ec9b50ce32c21f12a927d1fc2928d839c48df98d`

## Scope

- port the reviewed 31-file direct-command cleanup into the source tree
- preserve dated forensics, source-sync receipts, versioned release runbooks, articles, archived goals, historical RTK-specific failure records, and intentional RTK rejection tests
- preserve meaningful command-shape assertions by requiring direct `cargo xtask ` commands rather than weakening them

## Acceptance

- current source guidance and generated next actions use direct commands
- positive policy tests and accepted specs no longer require RTK
- negative wrapper tests and historical evidence remain
- final source PR contains no workflow or transport files
- exact-head source checks and review are clean or any unrelated baseline failures are explicitly classified

## Non-goals

- no parser, compiler, LSP, provider, release, CI-topology, or runtime product behavior change
- no bulk historical rewrite
- no model-performance claim

PR #10062, opened 2026-07-28
chore: stage residual RTK source port

EffortlessSteven:
Disposable transport transaction for #10061. It applies the exact merged `perl-lsp-swarm` cleanup patch against the source tree and exists only to produce source-local blobs. It will be closed without merge after a clean content-only branch is rebuilt from `master`.


EffortlessSteven:
Closed as disposable transport work as specified by the PR body. The residual RTK staging workflow is not a product change to merge; the source cleanup remains represented in its originating history.

PR #10094, opened 2026-08-27
fix(sync): preserve git commit message bytes on Windows

### LINKED ISSUE #10091: ci(sync): publication_sync_check.py re-derivation fails on Windows — text-mode stdin feeds CRLF messages to git commit-tree
$(cat <<'EOF'

EffortlessSteven:
## Objective

Fix the Windows-only publication-sync validation failure where text-mode subprocess stdin can translate the re-derived commit message to CRLF before `git commit-tree` hashes it.

## Change

- make the shared Git plumbing helper accept byte stdin and decode captured output explicitly;
- pass `commit_metadata`'s message bytes to `commit-tree` without newline translation;
- use the same byte-preserving path for `git mktree`;
- add a contract ratchet that prevents the object-construction helper from reverting to text-mode I/O.

## Verification

- `python3 -m py_compile scripts/publication_sync_check.py`
- `python3 scripts/tests/test-publication-sync-contract.py` — 36 tests passed
- `git diff --check` — clean

This is a contained CI/validator fix. It does not add publication authority or change the sync contract.

Closes #10091.

### DIFF
--- a/scripts/publication_sync_check.py
+++ b/scripts/publication_sync_check.py
@@ -130,20 +130,21 @@ def fail(message: str) -> NoReturn:
     raise ContractFailure(message)
 
 
-def run_git(repo: Path, args: Sequence[str], *, input_text: Optional[str] = None,
+def run_git(repo: Path, args: Sequence[str], *, input_bytes: Optional[bytes] = None,
             env: Optional[Dict[str, str]] = None) -> str:
-    """Run a git command; instrument failure is not success."""
+    """Run a git command with byte-preserving stdin and UTF-8 output."""
     proc = subprocess.run(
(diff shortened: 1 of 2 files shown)


EffortlessSteven (commented):
Review complete. The change is correctly scoped to the validator’s Git-object construction boundary: commit-tree now receives the exact UTF-8 message bytes, while captured Git output is decoded explicitly. The existing read-only text-mode helpers remain unchanged, and the 36-test contract suite covers the sync positives, falsifiers, post-merge checks, schema/template ratchets, and the new no-text-mode regression guard. No publication authority, repository refs, or workflow permissions are broadened.

